PR #1, opened 2018-01-12
[osal] Minor fixes to the os abstraction porting layer.

turon:
- Support pointer alignment on 64-bit architectures.
- C++ extern "C" typo.


sjanc:
Hi,

Just wonder about those 64bits,   mynewt is 32bit and adding 64bit would be much more invasive than this small change. (also static assert about void poitner being 32bit is no longer valid with this change).

 What was the purpose of this PR?

PR #8, opened 2018-03-19
nimble/mesh: Make use of advertising extensions

rymanluk:
With this patch Mesh can make use of multi instances which came with
advertising extension. It creates one advertising instance for
PB-ADV and one for PB-GATT and PROXY.

That said, it is possible to create additional instances which are
used for non mesh purpose. Note that this brings limitation on Mesh
Friendship due to timings.

To make use of it application needs to set BLE_EXT_ADV and
BLE_MULTI_ADV_INSTANCES at least to 1. Mesh will use last available
advertising instance for PB-ADV and second to last for PB-GATT and
PROXY

### DIFF
--- a/nimble/host/mesh/include/mesh/glue.h
+++ b/nimble/host/mesh/include/mesh/glue.h
@@ -238,7 +238,6 @@ void net_buf_reserve(struct os_mbuf *om, size_t reserve);
 #define net_buf_add_be32(a, b) net_buf_simple_add_be32(a, b)
 #define net_buf_add_be16(a, b) net_buf_simple_add_be16(a, b)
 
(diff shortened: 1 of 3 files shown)


rymanluk:
With latest commit in this PR and  https://github.com/apache/mynewt-core/pull/1078  it back to work.
Issue we had was that controller code start to be more strict and does not allow to mix legacy advertising HCI command with Advertising Extensions. Previous version of this PR was using Legacy HCI to start advertising but Advertising Extenstions to configure advertising instances. It is fixed now.

rymanluk:
fixed @andrzej-kaczmarek  comments and rebased

PR #9, opened 2018-03-19
nimble/ll: Do not use sec_txdone for legacy adv

andrzej-kaczmarek:


andrzej-kaczmarek:
@simonratner

nimble/controller/src/ble_ll_adv.c
```diff
@@ -1323,7 +1323,9 @@ ble_ll_adv_halt(void)
 
         os_eventq_put(&g_ble_ll_data.ll_evq, &advsm->adv_txdone_ev);
 #if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_EXT_ADV)
-        os_eventq_put(&g_ble_ll_data.ll_evq, &advsm->adv_sec_txdone_ev);
+        if (!(advsm->props & BLE_HCI_LE_SET_EXT_ADV_PROP_LEGACY)) {
```
simonratner: Wondering: why is the check explicitly for legacy adv rather than `ble_ll_adv_active_chanset_is_sec(advsm)` like in `ble_ll_adv_tx_done`?

andrzej-kaczmarek: ble_ll_adv_tx_done() is called from ISR where we need to figure out which chanset has just finished transmission and act accordingly. here we just want to stop everything thus is does not matter which chanset is active - just stop all that could be active.

PR #12, opened 2018-03-21
ble service: Device Information Service

sdalu:
See: https://github.com/apache/mynewt-core/pull/901

### DIFF
--- a/nimble/host/services/dis/src/ble_svc_dis.c
+++ b/nimble/host/services/dis/src/ble_svc_dis.c
@@ -1,0 +1,296 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <assert.h>
+#include <string.h>
+#include "sysinit/sysinit.h"
+#include "host/ble_hs.h"
+#include "services/dis/ble_svc_dis.h"
+
+/* Device information */
+struct ble_svc_dis_data ble_svc_dis_data = {
+    .model_number      = MYNEWT_VAL(BLE_SVC_DIS_MODEL_NUMBER_DEFAULT),
+    .serial_number     = MYNEWT_VAL(BLE_SVC_DIS_SERIAL_NUMBER_DEFAULT),
+    .firmware_revision = MYNEWT_VAL(BLE_SVC_DIS_FIRMWARE_REVISION_DEFAULT),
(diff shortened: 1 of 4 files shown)


sterlinghughes:
szymon, are we ready to merge?

nimble/host/services/dis/src/ble_svc_dis.c
```diff
@@ -0,0 +1,289 @@
...
+ */
+
+#include <assert.h>
+#include <string.h>
+#include "sysinit/sysinit.h"
+#include "host/ble_hs.h"
+#include "services/dis/ble_svc_dis.h"
+
+/* Device information */
+struct ble_svc_dis_data ble_svc_dis_data = { 0 };
```
andrzej-kaczmarek: this should be simply initialized with default data from syscfg

nimble/host/services/dis/src/ble_svc_dis.c
```diff
@@ -0,0 +1,289 @@
...
+	break;
+#endif
+    default:
+	assert(0);
+	return BLE_ATT_ERR_UNLIKELY;
+    }
+
+    if (info == NULL) {
+	info = "";
+    }
```
andrzej-kaczmarek: is this really needed? why not just skip appending to mbuf?

nimble/host/services/dis/syscfg.yml
```diff
@@ -0,0 +1,116 @@
...
+                BLE_GATT_CHR_F_READ_ENC
+                BLE_GATT_CHR_F_READ_AUTHEN
+                BLE_GATT_CHR_F_READ_AUTHOR
+            Set to '-1' to remove this characteristic.
+        value: -1
+    BLE_SVC_DIS_MODEL_NUMBER_DEFAULT:
+        description: >
+            Defines a default value for "Model number" if not set with
+            'ble_svc_dis_model_number_set', leaving empty will use the
+            empty-string as default value.
```
andrzej-kaczmarek: I'd skip remark regarding empty string value since this is clear by looking at value below. Also, if someone updates default value here I bet he will just forget to update this description and they won't match.

nimble/host/services/dis/syscfg.yml
```diff
@@ -0,0 +1,116 @@
...
+syscfg.defs:
+    BLE_SVC_DIS_MODEL_NUMBER_READ_PERM:
+        description: >
+            Defines permissions for reading "Model Number" characteristics. Can
+            be zero to allow read without extra permissions or combination of:
+                BLE_GATT_CHR_F_READ_ENC
+                BLE_GATT_CHR_F_READ_AUTHEN
+                BLE_GATT_CHR_F_READ_AUTHOR
+            Set to '-1' to remove this characteristic.
+        value: -1
```
andrzej-kaczmarek: this is c&p multiple times - I'd make it smarter by adding syscfg val e.g. BLE_SVC_DIS_DEFAULT_READ_PERM with all the description there and then just reference it in descriptions of all other characteristics. additionally, it may be set to 0 to allow read and value of this particular characteristic could be set to MYNEWT_VAL_BLE_SVC_DIS_DEFAULT_READ_PERM as an example so it inherits that common value. other characteristics can be -1 to be disabled by default.

nimble/host/services/dis/syscfg.yml
```diff
@@ -0,0 +1,116 @@
...
+    BLE_SVC_DIS_MODEL_NUMBER_DEFAULT:
+        description: >
+            Defines a default value for "Model number" if not set with
+            'ble_svc_dis_model_number_set', leaving empty will use the
+            empty-string as default value.
+        value:
```
andrzej-kaczmarek: in addition to my suggestion for default permissions, for this one I'd just set default value to "Apache Mynewt NimBLE" or sth. other characteristics are be disabled by default and default value empty.

sdalu:
Applied mentioned changes

PR #14, opened 2018-03-21
ble service: Current Time Service

sdalu:
Main characteristics for the Current Time Service

### DIFF
--- a/nimble/host/services/cts/src/ble_svc_cts.c
+++ b/nimble/host/services/cts/src/ble_svc_cts.c
@@ -1,0 +1,544 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <assert.h>
+#include <string.h>
+
+#include "sysinit/sysinit.h"
+#include "syscfg/syscfg.h"
+#include "host/ble_hs.h"
+#include "host/ble_gap.h"
+#include "services/cts/ble_svc_cts.h"
+#include "datetime/datetime.h"
+
+/* To use for days AND hours if update more than 255 days ago */
+#define BLE_SVC_CTS_UPDATED_SINCE_A_LONG_TIME_AGO 255
+
+#define BLE_SVC_CTS_ADJUST_REASON_MANUAL_TIME_UPDATE               1
+#define BLE_SVC_CTS_ADJUST_REASON_EXTERN_REFERENCE_TIME_UPDATE     2
+#define BLE_SVC_CTS_ADJUST_REASON_CHANGE_OF_TIMEZONE               4
+#define BLE_SVC_CTS_ADJUST_REASON_CHANGE_OF_DST                    8
+
+static bool     ble_svc_cts_reference_time_was_used  = false;
+static uint64_t ble_svc_cts_reference_time_last_used = 0;
+static uint8_t  ble_svc_cts_adjust_reason = 0;
+
+/* Used as buffer */
+static struct {
+    uint8_t time_source;
+    uint8_t time_accuracy;
+    uint8_t days;
+    uint8_t hours;
+} __attribute__((__packed__)) ble_svc_cts_reference_time_information = {
+    .time_source   = BLE_SVC_CTS_TIME_SOURCE_UNKNOWN,
+    .time_accuracy = BLE_SVC_CTS_TIME_ACCURACY_UNKNOWN,
+    .days          = BLE_SVC_CTS_UPDATED_SINCE_A_LONG_TIME_AGO,
+    .hours         = BLE_SVC_CTS_UPDATED_SINCE_A_LONG_TIME_AGO,
+};
+
+static struct {
+    uint16_t year;
+    uint8_t month;
+    uint8_t day;
+    uint8_t hours;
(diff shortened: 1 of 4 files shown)


sterlinghughes:
@sjanc are we ready to merge?

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+    }
+    if (ble_hs_mbuf_to_flat(ctxt->om,
+			    &ble_svc_cts_current_time, om_len, NULL)) {
+        return BLE_ATT_ERR_UNLIKELY;
+    }
+    
+    /* Convert to timeval */
+    struct clocktime   ct;
+    struct os_timezone tz;
+    struct os_timeval  tv;
```
andrzej-kaczmarek: all local variables should be defined at the beginning of function

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+	return BLE_ATT_ERR_UNLIKELY;
+    }
+
+    /* Are we ignoring the day of week */
+    if (ble_svc_cts_current_time.day_of_week) {
+	if (timeval_to_clocktime(&tv, &tz, &ct)) {
+	    return BLE_ATT_ERR_UNLIKELY;
+	}
+	if (ct.dow != (ble_svc_cts_current_time.day_of_week % 7)) {
+	    rc = 0x80; // Data field ignored
```
andrzej-kaczmarek: this should be defined as symbol

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+
+	/* Convert to days and hours format */
+	uint8_t  days, hours;
+	delta /= 3600; 
+	hours  = delta % 24;
+	delta /= 24;
+	days   = delta;
+
+	/* Special case if more than 254 days ago */
+	if (delta > 254) {
```
andrzej-kaczmarek: to be precise here - spec says more than 254 days and 23 hours

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+	    days = hours = BLE_SVC_CTS_UPDATED_SINCE_A_LONG_TIME_AGO;
```
andrzej-kaczmarek: no double assignments

also symbol should have different name, it does not quite fit here imo. I'd make it sth like `BLE_SVC_CTS_UPDATED_255_OR_MORE'

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+    /* Compute new timezone/dst */
+    struct os_timezone tz_n = {
+	.tz_minuteswest = ble_svc_cts_local_time_information.time_zone * 15,
+	.tz_dsttime     = ble_svc_cts_local_time_information.dst_offset != 0
+    };
+
+    /* Ensure conversion is legit 
+     * For DST, see: https://github.com/apache/mynewt-core/issues/944
+     */
+    if (ble_svc_cts_local_time_information.dst_offset != 4) {
```
andrzej-kaczmarek: 4 is not the only valid value for DST offset

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+     */
+    if ((tz.tz_minuteswest %   15) ||
+	(tz.tz_minuteswest < -720) ||
+	(tz.tz_minuteswest >  840)) {
+        return BLE_ATT_ERR_UNLIKELY; /* XXX: better error code? */
+    }
+        
+    /* Prepare data */
+    ble_svc_cts_local_time_information.time_zone  = tz.tz_minuteswest / 15;
+    ble_svc_cts_local_time_information.dst_offset = tz.tz_dsttime ? 4 : 0;
```
andrzej-kaczmarek: DST offset can be different than just 1h. not sure if any region in the world actually uses different offset, but this does not mean we can just assume it will be always 0 or 1.

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+    /* Apply time change */
+    if (os_settimeofday(&tv, NULL)) {
+	return BLE_ATT_ERR_UNLIKELY;
+    }
+
+    /* Notify 
+     *  (Manual change are always notify, don't bother hinting on the offset)
+     */
+    ble_svc_cts_reference_time_updated(BLE_SVC_CTS_TIME_SOURCE_MANUAL,
+				       BLE_SVC_CTS_TIME_ACCURACY_UNKNOWN, 0);
```
andrzej-kaczmarek: I'd say manual update is when user updates time on device, not when it is written by client. also this information should be passed from application, not automatically from service since spec does not define this anywyere. also this could trigger local time characteristic to be notified just after write which is weird, because client already knows it changed it...

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+
+static inline int
+ble_svc_cts_current_time_read_access(struct ble_gatt_access_ctxt *ctxt)
+{
+    /* Get current clock time */
+    struct os_timeval  tv;
+    struct os_timezone tz;
+    struct clocktime   ct;
+    
+    if (os_gettimeofday(&tv, &tz) || timeval_to_clocktime(&tv, &tz, &ct)) {
```
andrzej-kaczmarek: this introduces dependency to yet another Mynewt compoments thus making this service only usable with Mynewt - we would prefer something more generic

nimble/host/services/cts/src/ble_svc_cts.c
```diff
@@ -0,0 +1,544 @@
...
+    uint8_t secondes;
+    uint8_t day_of_week;
+    uint8_t fraction256;
+    uint8_t adjust_reason;
+} __attribute__((__packed__)) ble_svc_cts_current_time;
+
+static struct {
+    int8_t  time_zone;  /* -48 .. 56, -128 = unknown */
+    uint8_t dst_offset; 
+} __attribute__((__packed__)) ble_svc_cts_local_time_information;
```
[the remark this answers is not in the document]
sdalu: As it is a kind of large structure, the idea was to reduce the current thread stack usage by making them global

sjanc:
no action for almost 2 months, closing

PR #18, opened 2018-03-28
nimble/ll: Portion of fixes for the nimble controller.

rymanluk:
This PR fixes couple of bugs found during LL testing.

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -133,6 +133,7 @@ struct ble_ll_adv_sm
 #define BLE_LL_ADV_SM_FLAG_CONN_RSP_TXD         0x08
 #define BLE_LL_ADV_SM_FLAG_ACTIVE_CHANSET_MASK  0x30 /* use helpers! */
 #define BLE_LL_ADV_SM_FLAG_ADV_DATA_INCOMPLETE  0x40
+#define BLE_LL_ADV_SM_FLAG_ADV_TERMINATE_EVT    0x80
 
 #define ADV_DATA_LEN(_advsm) \
                 ((_advsm->adv_data) ? OS_MBUF_PKTLEN(advsm->adv_data) : 0)
@@ -2188,6 +2189,8 @@ ble_ll_adv_ext_set_param(uint8_t *cmdbuf, uint8_t *rspbuf, uint8_t *rsplen)
         advsm->flags &= ~BLE_LL_ADV_SM_FLAG_SCAN_REQ_NOTIF;
     }
 
+    advsm->flags |= BLE_LL_ADV_SM_FLAG_ADV_TERMINATE_EVT;
+
     rspbuf[0] = advsm->adv_txpwr;
     *rsplen = 1;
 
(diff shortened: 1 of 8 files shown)


nimble/controller/src/ble_ll_conn_hci.c
```diff
@@ -1260,13 +1260,12 @@ ble_ll_conn_hci_set_data_len(uint8_t *cmdbuf, uint8_t *rspbuf, uint8_t *rsplen)
             rc = BLE_ERR_SUCCESS;
         }
 
-        /* XXX: should I check against max supported? I think so */
+        if (connsm->max_tx_time != txtime || connsm->max_tx_octets != txoctets) {
```
andrzej-kaczmarek: this should not be executed when command returns an invalid parameters error

nimble/controller/src/ble_ll_ctrl.c
```diff
@@ -1599,6 +1599,16 @@ ble_ll_ctrl_rx_conn_param_req(struct ble_ll_conn_sm *connsm, uint8_t *dptr,
 {
     uint8_t rsp_opcode;
 
+    /* If Host masked out Remote Connection Parameter Request Event, we need to
+     * send Reject back to the remote device
+     */
+    if (!ble_ll_hci_is_le_event_enabled(BLE_HCI_LE_SUBEV_REM_CONN_PARM_REQ)){
```
andrzej-kaczmarek: masked event only matters if event should be indicated to host - you do not know this here. this should be added in `ble_ll_ctrl_conn_param_pdu_proc` - there's fixme comment regarding event masking there.

nimble/controller/src/ble_ll_ctrl.c
```diff
@@ -689,8 +689,8 @@ ble_ll_ctrl_rx_phy_req(struct ble_ll_conn_sm *connsm, uint8_t *req,
         /* XXX: TODO: if we started another procedure with an instant
          * why are we doing this? Need to look into this.*/
 
-        /* Respond to master's phy update procedure */
-        CONN_F_PEER_PHY_UPDATE(connsm) = 1;
+        /* Let us wait for LL_PHY_UPDATE_IND */
+        ble_ll_ctrl_proc_start(connsm, BLE_LL_CTRL_PROC_PHY_UPDATE);
```
andrzej-kaczmarek: seems like this only usually works because current procedure is non-idle when this is called - only bitmask is set in such case which is what you want. in case if above is executed current procedure will be set to idle and this will start an actualy phy update procedure.

nimble/controller/src/ble_ll_adv.c
```diff
@@ -2998,7 +3001,8 @@ ble_ll_adv_done(struct ble_ll_adv_sm *advsm)
         /* Legacy PDUs need to be stop here, for ext adv it will be stopped when
          * AUX is done.
          */
-        if (advsm->props & BLE_HCI_LE_SET_EXT_ADV_PROP_LEGACY) {
+        if (advsm->props & BLE_HCI_LE_SET_EXT_ADV_PROP_LEGACY &&
+                        advsm->flags & BLE_LL_ADV_SM_FLAG_ADV_TERMINATE_EVT) {
```
andrzej-kaczmarek: some extra parenthesis would make all these expression more readable

nimble/controller/src/ble_ll_conn_hci.c
rymanluk: ah right.

nimble/controller/src/ble_ll_ctrl.c
rymanluk: hmm you right.. will remove this patch now. need to figure out how to do it nicely.

wes3:
Looked over the changes and Andrzej comments and they seem good to me.

rymanluk:
actually there was one issue in last patch. We shall not set anything in rspbuf before we parse all the cmdbuf because basically rspbuf uses cmdbuf. Fixed. Can somebody check last patch now? thanks

PR #21, opened 2018-03-29
nimble/ll: Fix scanning handle for directed advertising

rymanluk:
So far Nimble did not sent advertising reports to host for directed
advertising events if RPA was not used. This is fixed with this patch.
If RPA is used in advertising report then LE Directed Advertising Report
Event  is sent to the host, otherwise LE Advertising Report Event.

### DIFF
--- a/nimble/controller/src/ble_ll_scan.c
+++ b/nimble/controller/src/ble_ll_scan.c
@@ -758,12 +758,12 @@ ble_ll_scan_send_adv_report(uint8_t pdu_type, uint8_t txadd, struct os_mbuf *om,
 
     if (pdu_type == BLE_ADV_PDU_TYPE_ADV_DIRECT_IND) {
         inita = rxbuf + BLE_LL_PDU_HDR_LEN + BLE_DEV_ADDR_LEN;
-        if (!(inita[5] & 0x40)) {
-            /* Let's ignore it if address is not resolvable */
-            return;
+        if ((inita[5] & 0x40)) {
+            /* For resolvable we send separate event */
+            subev = BLE_HCI_LE_SUBEV_DIRECT_ADV_RPT;
+        } else {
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
"LE Directed Advertising Report Event" should be sent only if TargetA is a RPA which controller cannot resolve and filter policy is set to accept such reports. However, controller does not even try to resolve TargetA when scanning (at least I don't see where it does it) so it does not have enough data to decide if mentioned event should be sent - so we should implement this first.

rymanluk:
@andrzej-kaczmarek you are looking for this function: ble_ll_scan_chk_filter_policy() and in particular this part:

    /* If this is a directed advertisement, init_addr is not NULL.
     * Check that it is for us */
    if (init_addr) {
        /* Is this for us? If not, is it resolvable */
        if (!ble_ll_is_our_devaddr(init_addr, init_addr_type)) {
            if (!chk_inita || !ble_ll_is_rpa(adv_addr, adv_addr_type)) {
                return 1;
            }
        }
    }

andrzej-kaczmarek:
@rymanluk this code does not resolve TargetA - it just compares it to either public or static random address

rymanluk:
@andrzej-kaczmarek  hmm actually you have right. But this is different issue. The one solved in the patch is that we actually start to send HCI Advertising Report Events when we received ADV_DIRECT_IND to us and TargetA is not RPA.

andrzej-kaczmarek:
ok, so we can assume that if we have TargetA which is an RPA at this point, this patch will guarantee that it will be sent via proper event. what we're missing is some handling in filtering policy most likely that would try to resolve TargetA first so if it can be resolved, at this point we won't have RPA anymore.

so I guess we can merge this fix assuming the other thing will be also fixed (soon)

PR #23, opened 2018-03-29
nimble/ll: Fix HCI Set Resolving Timeout

rymanluk:
With this patch, controller will accept resolving timeout even resolving
is not yet enabled.


rymanluk:
Not sure about error code to return when we fail to reset os_callout

rymanluk:
agreed..updating with a suggestion

PR #27, opened 2018-04-04
nimble/ll: Fix for not mixing BT5.0 and legacy HCI commands

rymanluk:
With this patch Controller makes sure that Host does not mix BT5 and
legacy HCI commands


rymanluk:
This commit is going in along with https://github.com/apache/mynewt-nimble/pull/26

PR #37, opened 2018-04-10
[RFC][WIP] ble service: HID service

samveen:
This is an effort to create an HID service for nimble.
### DONE:
- Skeleton created.

### TODO:
- Get it to a basic working state.
- Fill in everything else.

### Caveat Emptor:
- Needs a **LOT** of review.
- I don't know any of the details of the following protocols:
  - PS/2
  - USB
  - USB HID
  - BLE
  - HID over BLE

### References:
- https://www.bluetooth.com/specifications/gatt/viewer?attributeXmlFile=org.bluetooth.service.human_interface_device.xml

### DIFF
--- a/nimble/host/services/hid/include/services/hid/ble_svc_hid.h
+++ b/nimble/host/services/hid/include/services/hid/ble_svc_hid.h
@@ -48,6 +48,11 @@ typedef enum __attribute__((packed)) {
     BLE_SVC_HID_PROTOCOL_MODE_REPORT
 } ble_svc_hid_protocol_mode_value_t ;
 
+/* HID Service USB HID version
+ * Format: uint16
+ */
+#define USB_HID_VERSION_1_11    (0x0111)
+
 /* HID Service HID Information Flags
  * Format: 8-bit BitField
  * Note:The fields are in the order of LSO to MSO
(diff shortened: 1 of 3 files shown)


samveen:
@andrzej-kaczmarek 
- You are absolutely right that this is probably a horrible horrible idea, but I need an HID service and there isn't one yet, so why not? I don't know enough yet, but I am working on it (which is why it's marked as [WIP]
- I corrected my statement about`basic service creation`.
- There is some HID code in the `bletest` example, which I wanted to try and extract for reuse, but it is written in such a convoluted manner, that I cannot understand much yet. Hopefully I should start understanding some more soon.

andrzej-kaczmarek:
There is no HID code in `bletest`. The only reference to HID there is because sample advertising data have HID Service UUID, but it is just for testing purposes and could be virtually any value. Basically, `bletest` is just some testing code which is not really useful for an actual Mynewt app so it won't be helpful here.

samveen:
@andrzej-kaczmarek I've started reading the specs. I'll add more code as I understand more of the details. Hopefully I should have something to for it show in a week's time.

sjanc:
no action for almost 2 months, closing

PR #38, opened 2018-04-10
[WIP] Mesh storage

michal-narajowski:
This is an early version of Mesh storage functionality.

It allows you to store:
- main Mesh structure
- subnets
- appKeys

For now, the data can be saved by running `provision` command
from shell. To reset the storage run `reset` command. To restore
data frome storage run `restore`.

### DIFF
--- a/nimble/host/mesh/src/net.h
+++ b/nimble/host/mesh/src/net.h
@@ -269,6 +269,21 @@ extern struct bt_mesh_net bt_mesh;
 
 #define BT_MESH_NET_HDR_LEN 9
 
+void bt_mesh_net_set_last_update(s64_t last_udpate);
+s64_t bt_mesh_net_get_last_update(void);
+bool bt_mesh_net_get_valid(void);
+void bt_mesh_net_set_valid(bool valid);
+bool bt_mesh_net_get_iv_update(void);
+void bt_mesh_net_set_iv_update(bool iv_update);
+u32_t bt_mesh_net_get_iv_index(bool *iv_update);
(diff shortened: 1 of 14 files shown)


nimble/host/mesh/src/net.h
```diff
@@ -266,9 +269,27 @@ extern struct bt_mesh_net bt_mesh;
...
+bool bt_mesh_net_get_iv_update(void);
+void bt_mesh_net_set_iv_update(bool iv_update);
+u32_t bt_mesh_net_get_iv_index(bool *iv_update);
+void  bt_mesh_net_set_iv_index(u32_t index, bool update);
+u32_t bt_mesh_net_get_sequence_number(void);
+void  bt_mesh_net_set_sequence_number(u32_t seq_number);
+bool bt_mesh_net_get_pending_update(void);
+void bt_mesh_net_set_pending_update(bool pending);
+bool bt_mesh_net_get_key(u16_t net_idx, u8_t *key);
+bool bt_mesh_net_get_flags(u16_t net_idx, u8_t *out_flags);
```
jhedberg: Why do you need to add all of these new functions? The bt_mesh object is already exported to all mesh-internal c-files through net.h.

michal-narajowski: It's a proposal. Accessing the struct directly from many functions makes it hard to track where the struct is changed so it can be stored.

michal-narajowski:
I'm closing this one and will open a new pull request with a new implementation.

PR #41, opened 2018-04-12
Connection complete fix

sjanc:


nimble/host/src/ble_gap.c
```diff
@@ -449,7 +449,7 @@ ble_gap_extract_conn_cb(uint16_t conn_handle,
 {
     const struct ble_hs_conn *conn;
 
-    BLE_HS_DBG_ASSERT(conn_handle != 0);
+    BLE_HS_DBG_ASSERT(conn_handle <= 0x0eff);
```
rymanluk: Can we make 0x0eff as a define? Maybe BLE_CONN_HANDLE_MAX or similar? Or at least comment in the code

PR #54, opened 2018-04-18
nimble/ll: Fix advertising interval validation

sjanc:
For legacy advertising any interval between 0x0020 and 0x4000 is valid
as long as min interval is not greater than max interval.

For extended advertising any interval between 0x000020 and 0xffffff
is valid as long as min interval is not greater than max interval.

If intervals provided are outside of value allowed by specificication
then Invalid Parmeters error is returned.

If the advertising interval range provided by the Host is valid but
outside the advertising interval range supported by the Controller,
then the Controller shall return the Unsupported Feature or Parameter
Value (0x11) error code. Since for now we support only intervals
in legacy range we use this code in that case.

This was affecting TP/DDI/BI-02-C qualification test case.

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -1370,7 +1370,6 @@ ble_ll_adv_set_adv_params(uint8_t *cmd)
     uint8_t peer_addr_type;
     uint16_t adv_itvl_min;
     uint16_t adv_itvl_max;
-    uint16_t min_itvl;
     struct ble_ll_adv_sm *advsm;
     uint16_t props;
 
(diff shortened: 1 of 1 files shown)


nimble/controller/src/ble_ll_adv.c
```diff
@@ -1415,22 +1410,26 @@ ble_ll_adv_set_adv_params(uint8_t *cmd)
...
-        (adv_itvl_min > BLE_HCI_ADV_ITVL_MAX) ||
-        (adv_itvl_max > BLE_HCI_ADV_ITVL_MAX)) {
-        return BLE_ERR_INV_HCI_CMD_PARMS;
+    /* Make sure intervals values are valid
+     * (HD directed advertising ignores those parameters)
+     */
+    if (!(props & BLE_HCI_LE_SET_EXT_ADV_PROP_HD_DIRECTED)) {
+        if ((adv_itvl_min > adv_itvl_max) ||
+                (adv_itvl_min < BLE_LL_ADV_ITVL_MIN) ||
+                (adv_itvl_min > BLE_HCI_ADV_ITVL_MAX) ||
```
andrzej-kaczmarek: would be good to use the same symbols (either HCI or LL) for both min and max (also in changes below)

PR #55, opened 2018-04-18
nimble/host: Fix build with ext adv disabled

andrzej-kaczmarek:


andrzej-kaczmarek:
trivial build fix after merging #51

PR #58, opened 2018-04-24
call ble_gatts_chr_updated() on battery level change

davidzuhn:
This call doesn't require the connection handle, and thus doesn't
need an additional setup call to pass in that value.

### DIFF
--- a/nimble/host/services/bas/src/ble_svc_bas.c
+++ b/nimble/host/services/bas/src/ble_svc_bas.c
@@ -31,14 +31,6 @@
 static uint16_t ble_svc_bas_battery_handle;
 #endif
 
-/* Connection handle 
- *
- * TODO: In order to support multiple connections we would need to save
- *       the handles for every connection, not just the most recent. Then
- *       we would need to notify each connection when needed.
- * */
-static uint16_t ble_svc_bas_conn_handle;
-
 /* Battery level */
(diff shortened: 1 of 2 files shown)


nimble/host/services/bas/src/ble_svc_bas.c
```diff
@@ -118,17 +118,15 @@ ble_svc_bas_on_gap_connect(uint16_t conn_handle)
...
 ble_svc_bas_battery_level_set(uint8_t level) {
-    int rc = 0;
     if (level > 100)
 	level = 100;
     if (ble_svc_bas_battery_level != level) {
 	ble_svc_bas_battery_level = level;
 #if MYNEWT_VAL(BLE_SVC_BAS_BATTERY_LEVEL_NOTIFY_ENABLE) > 0
-	rc = ble_gattc_notify(ble_svc_bas_conn_handle,
-			      ble_svc_bas_battery_handle);
+        ble_gatts_chr_updated(ble_svc_bas_battery_handle);
```
andrzej-kaczmarek: so if `ble_svc_bas_conn_handle` is not needed you should also remove other code which references it

davidzuhn: That other code now has been removed.

PR #60, opened 2018-04-24
Fix data_status in BLE_GAP_EVENT_EXT_DISC

sjanc:
Since we decode this from HCI props host needs to have dedicated
defines as those no longer match HCI properties mask.


nimble/host/src/ble_hs_hci_evt.c
```diff
@@ -586,8 +588,15 @@ ble_hs_hci_evt_le_ext_adv_rpt(uint8_t subevent, uint8_t *data, int len)
                 continue;
             }
             desc.legacy_event_type = legacy_event_type;
+            desc.data_status = BLE_GAP_EXT_ADV_DATA_STATUS_COMPLETE;
         } else {
-            desc.data_status = (params->evt_type & BLE_HCI_ADV_DATA_STATUS_MASK) >> 5;
+            if (params->evt_type & BLE_HCI_ADV_DATA_STATUS_INCOMPLETE) {
```
andrzej-kaczmarek: data status is a 2-bit field so you need to mask it and compare

PR #72, opened 2018-04-27
Porting NimBLE

andrzej-kaczmarek:
(updated 20180505 to match latest version)

This is initial version of porting layer to have some generic and relatively easy way to run NimBLE on different OS. Also included is working port for FreeRTOS with working example code for nRF52 + nRF5 SDK (needs to be downloaded separately).

Since this is not yet documented (except for some sparse comments in FreeRTOS port), below is short description how it works (I'll use it as base for more detailes docs later). Also make sure to check each commit message as there are some helpful informations there as well.

Porting code consists of 3 parts:
- `nimble/include/nimble/nimble_npl.h' - porting layer API (this is what target OS shall implement)
- `porting/npl` - porting layer APIs implemented for specific target OS (currently Mynewt, FreeRTOS and dummy variants are available)
- `porting/nimble` - parts of `kernel/os` and few headers from Mynewt which are required to build/run NimBLE, "startup" code to initialize NimBLE easily and makefile helpers (this basically should not be touched when porting to target OS)

Assuming target OS is built using own makefile, process of porting NimBLE is as follows (see `porting/examples/freertos_nrf5_sdk` where this process is implemented in fully working app):
1. get target OS
2. add following to makefile
```
# set mynet-nimble root
NIMBLE_ROOT := /here/is/my/mynewt-nimble
# enable controller (host is by default)
NIMBLE_CFG_CONTROLLER := 1
# enable local TinyCrypt (can be omitted if target OS has TinyCrypt with AES and ECC)
NIMBLE_CFG_TINYCRYPT := 1
# setup NimBLE defs
include $(NIMBLE_ROOT)/porting/nimble/Makefile.defs 
```
3. including `Makefile.defs` defines few variable which should be used to add NimBLE files to build:
* `NIMBLE_SRC` - sources to be included in build
* `NIMBLE_OBJ` - same as above, but includes objects insteads
* `NIMBLE_INCLUDE` - include paths to be added (only paths, not prefixed with `-I`)
* `NIMBLE_CFLAGS` - extra flags to use when building (at least for NimBLE files)
4. add implementation of interfaces defined in `nimble/include/nimble/nimble_npl.h` (see `porting/npl/dummy`) and include this in build
5. if building host-only, add HCI transport to build (see `porting/npl/dummy` for template, do not forget to initialize in code)
6. in application, call `nimble_port_init()` before calling anything from NimBLE code (preferably during startup)
7. in application, create task for NimBLE host (app priority) and, if required, for NimBLE controller (highest prio)
* host task should do more or less the same as `main()` in Mynewt app, i.e. configure host and run default queue
* controller task should just call `nimble_port_ll_task_func()`
8. build and run

Note on configuration: since Mynewt (and thus NimBLE) in configured using syscfg mechanism provided by `newt` tool, the configuration is provided by `syscfg.h` file generated by this tool. It can be either modified by hand or substituted by other configuration file. To make things a bit easier, `porting/nimble` has an app package which NimBLE dependencies - just create target for this app, change syscfg values, build (it won't link since there is no `main()` etc. but it does not matter) and extract generated `syscfg.h` file.

What should be done next:
- [ ] document porting process
- [ ] make configuration easier
- [ ] figure our sysinit configuration (which extra packages should be inited, like GATT services)
- [ ] parts of `porting/nimble` is a bit of a mess so need to remove unused stuff
- [x] ideally NPL should have own API for mutexes (`struct npl_mutex`), semaphores (`struct npl_sem`), queues etc. used everywhere in NimBLE; currently some structures have to include certain fields only because they are accessed directly in Mynewt code - own APIs would resolve this; this means also shim porting layer for Mynewt would be necessary though
- [ ] ???


sjanc:
Hi,

this looks good to me, I have a bit of mixed feelings about copy of tinycrypt (ie this should be supplied by OS) but if that makes porting easier I could live with that.

andrzej-kaczmarek:
fixed issue with dropped connections

porting/examples/freertos_nrf5_sdk/main.c
```diff
@@ -0,0 +1,128 @@
...
+
+    /* Initialize NimBLE porting layer */
+    nimble_port_init();
+
+    /*
+     * Create task where NimBLE host will run. It is not strictly necessary to
+     * have separate task for NimBLE host, but since something needs to handle
+     * default queue it is just easier to make separate task which does this.
+     */
+    xTaskCreate(nimble_host_task, "nh", configMINIMAL_STACK_SIZE + 400,
```
turon: Shouldn't the task creation be abstracted by os_task_create API, similar to os_mutex_create, etc?

porting/examples/freertos_nrf5_sdk/main.c
```diff
@@ -0,0 +1,128 @@
+/*
```
turon: Could this example be made completely portable and put in examples/common or examples/dummy?

porting/npl/freertos/include/npl/npl.h
```diff
@@ -0,0 +1,180 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef _NPL_H_
+#define _NPL_H_
```
turon: I preferred when os_<module> function call API was in common headers, and platform defines the specific struct os_<module>.  Is it possible to return to that model?  This scheme has no enforcement of common osal API.

npl.h can define the structs and then include all the individual os_<module>.h headers: os_eventq.h, os_mutex.h, ...

porting/npl/freertos/include/npl/npl.h
```diff
@@ -0,0 +1,180 @@
+/*
```
turon: Why isn't there a mynewt/include/npl/npl.h?

All platforms should use the exact same osal, including mynewt.

porting/examples/dummy/Makefile
```diff
@@ -0,0 +1,33 @@
+# Configure NimBLE variables
```
turon: Can this app share the name with its npl?
Either call the app "template" or rename the npl it uses "dummy".
Is there a way to make an app completely generic, and compile it against your npl of choice?

porting/examples/freertos_nrf5_sdk/ble.c
```diff
@@ -0,0 +1,109 @@
...
+    case BLE_GAP_EVENT_DISCONNECT:
+        start_advertise();
+        break;
+    }
+
+    return 0;
+}
+
+static void
+start_advertise(void)
```
turon: Is code in this file specific to the example app "main.c"?
If so, it may be more clear to simply include everything for the example app in one file: main.c.

porting/examples/freertos_nrf5_sdk/main.c
andrzej-kaczmarek: There's only one task created by NimBLE (LL task) so I don't think it makes sense to create abstraction for a single call. Instead, application can simply create this task using OS native API. Host task (as above) is actually optional and not created by NimBLE - in Mynewt this code runs in default app task which handles default event queue. Also there API semantics differences between e.g. Mynewt (stack passed as an argument) and FreeRTOS (stack allocated internally from heap) so basically it will take more code to abstract this properly than just create a task in application.

porting/npl/freertos/include/npl/npl.h
andrzej-kaczmarek: I also preferred it that way and will try to fix this. The problem was mostly due to the way NimBLE includes various OS headers and in some cases two files needed different ordering of the same includes to build properly. I made some cleanup in the meantime so it may be better now, will give it a try.

Another small advantage of having declarations in OS NPL is that some functions can be implemented as `static inline`.

porting/examples/dummy/Makefile
andrzej-kaczmarek: Dummy app and template NPL are not really related. The purpose of this app is only to verify that porting layer builds. Once we have NPL that can be build on Linux we can create simple app using that NPL.

There's no way to make application completely generic since we do not make abstraction of complete OS. NPL only abstracts calls which are required to run NimBLE on other OS.

porting/npl/freertos/include/npl/npl.h
andrzej-kaczmarek: As I wrote in description, the best would be if NimBLE defined its own interface for mutexes, semaphores etc. as true NPL and have all platforms implement this NPL, including Mynewt. This is not possible at the moment since it requires some more changes in NimBLE but it should be possible eventually. At the moment we need to use API imposed by Mynewt and just use this as NPL (or "NPL").

porting/examples/freertos_nrf5_sdk/ble.c
andrzej-kaczmarek: Well, `ble.c` is actually one piece of code that has NimBLE-only code and could be shared with other apps, but I guess we should have more examples for different platforms to make some code common (we only have 1 now).

andrzej-kaczmarek:
Last two commits update NPL to include "native" API, i.e. functions and structs are not the same as Mynewt but are specific to NimBLE. This means we do not explicitly depend on Mynewt API and also Mynewt will need its porting layer (although new function APIs are pretty similar to Mynewt's so it should not be a big problem here).

This means we have single `porting/nimble/include/nimble/nimble_npl.h` file which is a common definition for all APIs and structures are defined in `nimble/nimble_npl_os.h` for each NPL in OS-specific manner. I guess this is what is expected here.

However, some changes are still required to NimBLE code to fully utilize these APIs since some of them do not exist in Mynewt and basically wrap operations that are now done by accessing various structures directly.

andrzej-kaczmarek:
Now there's `nimble/include/nimble/nimble_npl.h` with complete porting interface declarations, OS-specific part is in `porting/npl/<something>/include/nimble/nimble_npl_os.h`.

Also, Mynewt is built with its own NPL implementation (which is just shim layer on top of `kernel/os` APIs).

sjanc:
looks OK, but sample with SDK scripts needs to be clarified about licence.

andrzej-kaczmarek:
I removed patch with nRF5 SDK sample from this PR - will push it to another PR where we can resolve licensing issues.

PR #73, opened 2018-04-30
Refactor RPA handling in controller

sjanc:
This set do a bit of refactoring on how peer RPAs are handled. How those are similar to local RPA when it is regenerated only from timer and user are taking current RPA. Advertising code is also adjusted to make use of that.

This is needed to implement LE Read Peer Resolvable Address Command where test spec assumes address read after LE Connection Complete Event (within RPA timeout) is same as one used over the radio.

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -87,8 +87,8 @@ struct ble_ll_adv_sm
     uint8_t adv_chan;
     uint8_t adv_pdu_len;
     int8_t adv_rpa_index;
-    uint8_t flags;
     int8_t adv_txpwr;
+    uint16_t flags;
     uint16_t props;
     uint16_t adv_itvl_min;
     uint16_t adv_itvl_max;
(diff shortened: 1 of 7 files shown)


nimble/controller/src/ble_ll_adv.c
```diff
@@ -239,16 +240,29 @@ ble_ll_adv_rpa_update(struct ble_ll_adv_sm *advsm)
...
+void
+ble_ll_adv_rpa_timeout(void)
+{
+    struct ble_ll_adv_sm *advsm;
+    int i;
+
+    for (i = 0; i < BLE_ADV_INSTANCES; i++) {
+        advsm = &g_ble_ll_adv_sm[i];
+
+        if (advsm->adv_enabled) {
```
rymanluk: Would it make sense to check if advertising instance is using RPA before setting this flag?

nimble/controller/src/ble_ll_resolv.c
```diff
@@ -406,15 +406,53 @@ ble_ll_resolv_enable_cmd(uint8_t *cmdbuf)
...
+    int rc;
+
+    addr_type = cmdbuf[0];
+    ident_addr = cmdbuf + 1;
+
+    rl = ble_ll_resolv_list_find(ident_addr, addr_type);
+    if (rl) {
+        memcpy(rspbuf, rl->rl_local_rpa, BLE_DEV_ADDR_LEN);
+
+
```
rymanluk: double empty line

PR #86, opened 2018-05-10
nimble/mesh: Add bt_mesh_register_gatt to pkg init

rymanluk:
Whenever mesh is initialized we need to make sure GATT services are
registered for GATT proxy


rymanluk:
@kolodgrz  could you have a look

PR #87, opened 2018-05-10
nimble/gap: Fix assert on slave connection with ext advertising

sjanc:
When extended advertising is enabled and advertising instance is
configured to own_address_type=random, it is using per instance
random address and not global random address. We need to store
address set in instance and pass it to connection object so that
it can be returned as ID address.


sjanc:
@kolodgrz  FYI

PR #90, opened 2018-05-12
[linux] Add linux port of nimble using npl.

turon:
Created linux port and npl (nimble platform layer) adapter: task, sem, mutex, callout, eventq, etc.

* Added unit tests for npl.
* Mods to support pointer alignment on 64-bit architectures.
* Fixes to ble_hci_sock to support linux port

Tested advertising, scan response, connections and GATT database operations.


nimble/host/src/ble_hs_startup.c
```diff
@@ -307,7 +307,7 @@ ble_hs_startup_set_evmask_tx(void)
                                                 BLE_HCI_OCF_CB_SET_EVENT_MASK2),
                                      buf, sizeof(buf));
     if (rc != 0) {
-        return rc;
+        BLE_HS_LOG(WARN, "ble_hs_startup_set_evmask_tx() failed\n");
```
andrzej-kaczmarek: why need this change?

nimble/include/nimble/nimble_npl.h
```diff
@@ -28,8 +28,10 @@
 extern "C" {
 #endif
 
-#ifndef BLE_NPL_STORAGE_DECL
+#ifdef  BLE_NPL_DECL_INLINE
 #define BLE_NPL_STORAGE_DECL static inline
+#else
+#define BLE_NPL_STORAGE_DECL
```
andrzej-kaczmarek: why need this change?

nimble/include/nimble/nimble_npl.h
```diff
@@ -125,7 +127,8 @@ ble_npl_error_t ble_npl_sem_init(struct ble_npl_sem *sem, uint16_t tokens);
 
 BLE_NPL_STORAGE_DECL
 ble_npl_error_t ble_npl_sem_pend(struct ble_npl_sem *sem,
-                                 ble_npl_time_t timeout);
+                                 ble_npl_time_t 
+                                 out);
```
andrzej-kaczmarek: why need this change?

nimble/include/nimble/nimble_npl.h
```diff
@@ -172,6 +175,20 @@ ble_npl_time_t ble_npl_time_ms_to_ticks32(uint32_t ms);
 BLE_NPL_STORAGE_DECL
 uint32_t ble_npl_time_ticks_to_ms32(ble_npl_time_t ticks);
 
+typedef void (*ble_npl_task_func_t)(void *);
+struct ble_npl_task;
+
+BLE_NPL_STORAGE_DECL
+int ble_npl_task_init(struct ble_npl_task *t, const char *name, ble_npl_task_func_t func,
+		 void *arg, uint8_t prio, ble_npl_time_t sanity_itvl,
+		 ble_npl_stack_t *stack_bottom, uint16_t stack_size);
```
andrzej-kaczmarek: we do not have tasks in NPL on purpose - required tasks should be created by target os/application (we'll have this described in doc)

the reason is that there only a handful of tasks required depending on what features should be enabled. the API above is copied straight from Mynewt and is not really portable. for example, FreeRTOS API does allocate stack internally instead of having it passed as a parameter. so unless there is good API proposal for this I'd stick with our current approach.

nimble/transport/socket/src/ble_hci_socket.c
```diff
@@ -142,27 +144,27 @@ STATS_NAME_END(hci_sock_stats)
 #define BLE_SOCK_STACK_SIZE         \
     OS_STACK_ALIGN(MYNEWT_VAL(BLE_SOCK_STACK_SIZE))
 
-struct os_task ble_sock_task;
+struct ble_npl_task ble_sock_task;
 
-static struct os_mempool ble_hci_sock_evt_hi_pool;
+static struct os_mempool s_hci_sock_evt_hi_pool;
```
andrzej-kaczmarek: why need to change static variable names?

nimble/transport/socket/src/ble_hci_socket.c
```diff
@@ -420,15 +422,15 @@ ble_hci_sock_rx_msg(void)
...
 {
     int rc;
 
     rc = ble_hci_sock_rx_msg();
     if (rc == 0) {
-        os_eventq_put(&ble_hci_sock_state.evq, &ble_hci_sock_state.ev);
+        ble_npl_eventq_put(&ble_hci_sock_state.evq, &ble_hci_sock_state.ev);
     } else {
-        os_callout_reset(&ble_hci_sock_state.timer, OS_TICKS_PER_SEC / 100);
+        ble_npl_callout_reset(&ble_hci_sock_state.timer, OS_TICKS_PER_SEC / 100);
```
andrzej-kaczmarek: explicit usage of `OS_TICKS_PER_SEC` should be removed in favor of NPL APIs

nimble/transport/socket/src/ble_hci_socket.c
```diff
@@ -510,6 +512,10 @@ ble_hci_sock_config(void)
         goto err;
     }
 
+    // HCI User Channel requires exclusive access to the device.
+    // The device has to be down at the time of binding.
```
andrzej-kaczmarek: do not use C++ style comments

nimble/transport/socket/src/ble_hci_socket.c
```diff
@@ -743,26 +749,27 @@ static void
...
-    os_callout_stop(&ble_hci_sock_state.timer);
-    os_callout_init(&ble_hci_sock_state.timer, &ble_hci_sock_state.evq,
+    ble_npl_eventq_init(&ble_hci_sock_state.evq);
+    ble_npl_callout_stop(&ble_hci_sock_state.timer);
+    ble_npl_callout_init(&ble_hci_sock_state.timer, &ble_hci_sock_state.evq,
                     ble_hci_sock_rx_ev, NULL);
+
+    ble_npl_task_init(&ble_sock_task, "hci_sock", ble_hci_sock_ack_handler, NULL,
+                      MYNEWT_VAL(BLE_SOCK_TASK_PRIO), OS_WAIT_FOREVER, pstack,
+                      BLE_SOCK_STACK_SIZE);
```
andrzej-kaczmarek: this should be Mynewt call with `#ifdef MYNEWT` as in other components

porting/examples/dummy/Makefile
```diff
@@ -17,7 +17,7 @@ INC = \
 
 OBJ := $(SRC:.c=.o)
 
-CFLAGS := $(NIMBLE_CFLAGS) -DBLE_NPL_STORAGE_DECL=
+CFLAGS := $(NIMBLE_CFLAGS)
```
andrzej-kaczmarek: no need to change this, it was on purpose as we default to `static inline` declarations

porting/examples/linux/Makefile
```diff
@@ -0,0 +1,179 @@
...
+
+include Makefile.common
+
+$(foreach target, $(TARGETS), $(call define_target, $(target)))
+
+# Flash the program
+flash: $(OUTPUT_DIRECTORY)/nrf52832_xxaa.hex
+	@echo Flashing: $<
+	nrfjprog --program $< -f nrf52 --sectorerase
+	nrfjprog --reset -f nrf52
```
andrzej-kaczmarek: these makefiles are obviously copied from nRF5 SDK - we cannot have it like this in repo. I'd expect that `Makefile` for Linux port is similar to what we have for dummy example

porting/nimble/src/os_mempool.c
```diff
@@ -30,7 +30,7 @@ STAILQ_HEAD(, os_mempool) g_os_mempool_list =
     STAILQ_HEAD_INITIALIZER(g_os_mempool_list);
 
 #if MYNEWT_VAL(OS_MEMPOOL_POISON)
-static uint32_t os_mem_poison = 0xde7ec7ed;
+static uintptr_t os_mem_poison = 0xde7ec7ed;
```
andrzej-kaczmarek: this is not a pointer value - it's just a regular value

porting/npl/linux/include/nimble/nimble_npl_os.h
```diff
@@ -0,0 +1,78 @@
...
+typedef int32_t ble_npl_stime_t;
+
+#define SYSINIT_PANIC_MSG(msg) __assert_fail(msg, __FILE__, __LINE__, __func__)
+
+#define SYSINIT_PANIC_ASSERT_MSG(rc, msg) do \
+{                                            \
+    if (!(rc)) {                             \
+        SYSINIT_PANIC_MSG(msg);              \
+    }                                        \
+} while (0)
```
andrzej-kaczmarek: most of the things above should not be part of NPL as they redefine Mynewt symbols - we do not do this in NPL anymore. check dummy example.

porting/npl/linux/include/nimble/nimble_npl_os.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+//#include "npl_freertos.h"
+
+
+#if NIMBLE_CFG_CONTROLLER
+static inline void
+ble_npl_hw_set_isr(int irqn, uint32_t addr)
+{
+    npl_freertos_hw_set_isr(irqn, addr);
+}
```
andrzej-kaczmarek: this is c&p from FreeRTOS - remove. we won't have controller on Linux.

porting/npl/linux/src/os_atomic.c
```diff
@@ -0,0 +1,36 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <stdint.h>
+#include <pthread.h>
+
+#include "osal.h"
```
andrzej-kaczmarek: do not use `osal.h` name - it's common with different vendors to call their own OS porting layer like this so may cause name collisions. this is why our porting layer is called NPL instead.

porting/npl/linux/src/os_callout.c
```diff
@@ -0,0 +1,120 @@
...
+
+
+void ble_npl_callout_init(struct ble_npl_callout *c, 
+                          struct ble_npl_eventq *evq,
+                          ble_npl_event_fn *ev_cb, 
+                          void *ev_arg)
+{
+    struct sigevent         event;
+
+    // Initialize the callout.
```
andrzej-kaczmarek: no C++ style comments

porting/npl/linux/src/os_mutex.c
```diff
@@ -0,0 +1,81 @@
...
+
+#include <pthread.h>
+
+ble_npl_error_t
+ble_npl_mutex_init(struct ble_npl_mutex *mu)
+{
+    pthread_mutexattr_t muAttr;
+
+    if (!mu)
+    {
```
andrzej-kaczmarek: should be on the same line as `if`

porting/npl/linux/src/os_mutex.c
```diff
@@ -0,0 +1,81 @@
...
+#include <stdint.h>
+#include <string.h>
+#include "os/os.h"
+
+#include <pthread.h>
+
+ble_npl_error_t
+ble_npl_mutex_init(struct ble_npl_mutex *mu)
+{
+    pthread_mutexattr_t muAttr;
```
andrzej-kaczmarek: no camel case in names

porting/npl/linux/src/os_sem.c
```diff
@@ -0,0 +1,102 @@
...
+    int err = 0;
+    struct timespec wait;
+
+    if (!sem)
+    {
+        return OS_INVALID_PARM;
+    }
+
+    err = clock_gettime(CLOCK_REALTIME, &wait);
+    if (err) return OS_ERROR;
```
andrzej-kaczmarek: return should be in `{ }` on separate line

porting/npl/linux/src/wqueue.h
```diff
@@ -0,0 +1,76 @@
...
+
+#ifndef __wqueue_h__
+#define __wqueue_h__
+
+#include <pthread.h>
+#include <list>
+
+using namespace std;
+
+template <typename T> class wqueue
```
andrzej-kaczmarek: do not mix C++ into C code - this has to be rewritten

we can have C++ port for something, but this should be separate NPL

tests/unit/porting/npl/Makefile
```diff
@@ -0,0 +1,111 @@
+# Makefile
```
andrzej-kaczmarek: I believe these are unit tests for Linux NPL so should be in `porting/npl/linux/tests`
if we need to have unit tests for NPL itself, there is already test framework which can be run via `newt` tool and it should go to `porting/test`

nimble/include/nimble/nimble_npl.h
turon: This was a rebase error.  Will revert.

nimble/include/nimble/nimble_npl.h
turon: Nimble is creating tasks internally, specifically `nimble/transport/socket/src/ble_hci_socket.c`, so should I sought to use npl for this.  FreeRTOS10 can be configured to take passed in pre-allocated stack.  I think npl should provide a complete osal abstraction, which would include a task api.

nimble/transport/socket/src/ble_hci_socket.c
turon: I thought `s_` was the naming convention used for static variables within nimble.  It looks like the convention may be `g_`.  Either way, when working in this code, I was running into cases where simple searches would not disambiguate between static variables and function names, so I moved these statics to follow a more common naming convention.

nimble/transport/socket/src/ble_hci_socket.c
turon: Agreed.  This got missed -- the older osal used these defines directly and I missed the change to new NPL_ names.

nimble/transport/socket/src/ble_hci_socket.c
turon: Ok

nimble/transport/socket/src/ble_hci_socket.c
turon: But that will fail when compiled for Linux...  Abstracting this call is exactly why NPL exists!

porting/examples/dummy/Makefile
turon: But then when the define is missing, NPL using simple function calls fails to link...
Better is to default to a working, simple solution and then add defines to enable optimizations.

porting/examples/linux/Makefile
turon: Ok, I'll adapt the dummy makefile.

nimble/include/nimble/nimble_npl.h
andrzej-kaczmarek: tasks are only created once and "just work" so for now this is done by app with provided task entry function:
https://github.com/apache/mynewt-nimble/blob/master/nimble/controller/src/ble_ll.c#L1543
https://github.com/andrzej-kaczmarek/apache-mynewt-nimble/blob/f4d909e2d6c0624b8936963e5dd204a219ae78cd/porting/examples/freertos_nrf5_sdk/main.c#L118

we can add task creation abstraction later (other APIs for tasks are not needed) as I'd prefer to have this defined in some better way than just c&p from Mynewt to cover differences between OS-es.

turon: In the case of hci sock, there are order-of-operation issues with your proposed technique.  main can't safely create the hci task before or after calling `ble_hci_sock_init`.  

If called before, event queue isn't setup.

If called after, config fails (as it tries to open socket and setup hci before event handler task is started).

Both failures result in failure setting up socket:
```
bind() failed 19
nimble-linux: ../../../nimble/transport/socket/src/ble_hci_socket.c:852: ble_hci_sock_init: Assertion `Failure configuring socket HCI' failed.
Aborted (core dumped)
```

nimble/host/src/ble_hs_startup.c
turon: ok, good point.  I'll investigate making such a patch.

porting/npl/linux/include/nimble/nimble_npl_os.h
turon: hci_sock is using `SYSINIT_PANIC_ASSERT_MSG` in addition to `SYSINIT_PANIC_ASSERT` provided by NPL and shown in dummy example.

nimble/host/src/ble_hs_startup.c
turon: Oh, I see #93 is the patch.  How soon can this be merged?

porting/npl/linux/src/wqueue.h
turon: Why require the busy work to rewrite this working, tested C++ queue implementation?
It would only be used by the Linux NPL porting layer.

porting/npl/linux/include/nimble/nimble_npl_os.h
andrzej-kaczmarek: so this should go to `porting/nimble/include/sysinit/sysinit.h` instead of OS code

porting/npl/linux/src/wqueue.h
andrzej-kaczmarek: because this is C project and it should build with C compiler.
as I said, we can add C++ linux port, but in this case I'd expect it also uses C++ standard library for threads, mutexes etc.

turon: @andrzej-kaczmarek Is removing the C++ the last item blocking acceptance of this PR?

andrzej-kaczmarek: IIRC there were some more comments to be fixed, like macros defined in NPL instead of common code (perhaps some more, I do not remember now)

PR #96, opened 2018-05-15
[64-bit] Patches to support running on 64-bit architectures.

turon:


porting/nimble/src/os_mempool.c
```diff
@@ -94,7 +94,7 @@ os_mempool_init(struct os_mempool *mp, uint16_t blocks, uint32_t block_size,
     mp->mp_min_free = blocks;
     mp->mp_flags = 0;
     mp->mp_num_blocks = blocks;
-    mp->mp_membuf_addr = (uint32_t)membuf;
+    mp->mp_membuf_addr = (uintptr_t)membuf;
```
sjanc: Shouldn't mp_membuf_addr member in struct os_mempool be also changed to uintptr_t? Otherwise it will be clipped to 32bits, right?

sjanc:
could you prefix and name commits similar to rest of the code?  ie
"porting: Add support for running on 64-bit architectures"

PR #98, opened 2018-05-15
Update Mesh to use NPL

michal-narajowski:


nimble/include/nimble/nimble_npl.h
```diff
@@ -68,8 +68,13 @@ struct ble_npl_eventq *ble_npl_eventq_dflt_get(void);
 
 void ble_npl_eventq_init(struct ble_npl_eventq *evq);
 
+struct ble_npl_event *ble_npl_eventq_get_tmo(struct ble_npl_eventq *evq,
+                                             ble_npl_time_t tmo);
+
 struct ble_npl_event *ble_npl_eventq_get(struct ble_npl_eventq *evq);
 
+struct ble_npl_event *ble_npl_eventq_get_no_wait(struct ble_npl_eventq *evq);
```
andrzej-kaczmarek: this can be actually `ble_npl_eventq_get_tmo(evq, 0)` - no need to create too many functions
(but I'd keep `ble_npl_evqnt_get` for "historical" reasons ;-) )

porting/npl/freertos/include/nimble/nimble_npl_os.h
```diff
@@ -99,12 +99,24 @@ ble_npl_eventq_init(struct ble_npl_eventq *evq)
     evq->q = xQueueCreate(32, sizeof(struct ble_npl_eventq *));
 }
 
+static inline struct ble_npl_event *
+ble_npl_eventq_get_tmo(struct ble_npl_eventq *evq, ble_npl_time_t tmo)
+{
+    return npl_freertos_eventq_get_tmo(evq, tmo);
```
andrzej-kaczmarek: this is no longer necessary - you can just define `ble_npl_eventq_get_tmo` in `.c` file without need to "jump" through `static inline` stub. we'll fix other calls later.

michal-narajowski: We decided to not change this now and fix all calls later.

michal-narajowski:
@sjanc @andrzej-kaczmarek Thanks, fixed.

PR #100, opened 2018-05-17
nimble/ll: Use simply memcpy for copying in resolv

andrzej-kaczmarek:
Copying data by words requires both src and dst to be word aligned, but
fc4497b78 modified ble_ll_resolv_entry struct so neither rl_local_irk
nor rl_peer_irk are aligned properly so copying does not work on nRF51
(Cortex-M0).

Let's use memcpy() here which is a bit slower, but since this code is
not used very often it does not really matter.

### DIFF
--- a/nimble/controller/src/ble_ll_resolv.c
+++ b/nimble/controller/src/ble_ll_resolv.c
@@ -84,9 +84,6 @@ ble_ll_resolv_gen_priv_addr(struct ble_ll_resolv_entry *rl, int local)
 {
     uint8_t *irk;
     uint8_t *prand;
-    uint32_t *irk32;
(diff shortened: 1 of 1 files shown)


wes3:
Not sure it matters any longer, but I have recollections that there was not enough time in the past and that this code was specifically modified to copy by words. Things might have changed enough so that this is no longer needed and I might be mis-remembering, but looking at the code it seemed to ring a bell :-)

andrzej-kaczmarek:
@wes3 I also thought this optimization was for a reason, but in the current version we use this only for handling HCI command or from RPA-timeout timer so both are not time critical.

PR #102, opened 2018-05-18
Documentation update

sjanc:
This moves and updated documentation from -core repo to -nimble repo. Also some initial work on doxygen generation is done.


sjanc:
Hi,

I added those to keep headers consistent with others so that there is no need to edit .h file while extending documentation. I wasn't planning to add those myself TBH :-)

aditihilbert:
Oh, I see  👍

PR #104, opened 2018-05-21
Add tracing via SystemView

andrzej-kaczmarek:
This makes use of new features provided by https://github.com/apache/mynewt-core/pull/1107 and adds option to trace NimBLE LL and PHY events via SystemView. It effectively replaces existing LL logging which was temporary anyway.


andrzej-kaczmarek:
Note that with default tracing options (i.e. OS task activity also enabled) this causes connection timeouts on nRF51. Not sure if it would be any better with less logging (did not try it yet) but it could be just that MCU is too slow to handle extra stuff during connection.

On nRF52840 everything works fine with tracing enabled:
image

PR #105, opened 2018-05-22
porting: Cleanup and fix NPL

andrzej-kaczmarek:


andrzej-kaczmarek:
@turon @MariuszSkamra

porting/npl/freertos/include/nimble/nimble_npl_os.h
```diff
@@ -128,7 +122,7 @@ ble_npl_eventq_run(struct ble_npl_eventq *evq)
 {
     struct ble_npl_event *ev;
 
-    ev = ble_npl_eventq_get(evq);
+    ev = ble_npl_eventq_get_tmo(evq, BLE_NPL_TIME_FOREVER);
```
rymanluk: it probably does not build as ble_npl_eventq_get_tmo is removed

porting/nimble/src/nimble_port.c
```diff
@@ -44,14 +38,6 @@ nimble_port_init(void)
 
     ble_hs_init();
 
-    /* XXX These should be configurable somehow */
-    ble_svc_gap_init();
```
rymanluk: I'm not sure about gap service. I would keep it here imho as this is needed but not everybody is aware of it.

andrzej-kaczmarek: it's mandatory for peripheral and central so should only be included if needed. in Mynewt app needs to explicitly add dep to package for GAP (and also GATT) service to be included so it works pretty much the same here.

porting/npl/freertos/include/nimble/nimble_npl_os.h
andrzej-kaczmarek: it builds because I removed `ble_npl_eventq_get` instead from this NPL :-)
but mesh probably does not build on FreeRTOS since it uses `ble_npl_eventq_get`.
will fix.

porting/nimble/src/nimble_port.c
rymanluk: I agree

andrzej-kaczmarek:
@rymanluk fixed FreeRTOS NPL issue

PR #108, opened 2018-05-24
nimble/ll: Vendor error event on assert

rymanluk:
With this patch controller will send Vendor Debug event
on asserts.
First octet of this event is Debug event type
1) Debug event type (0x00 - assert information)
2) Assert information (file and line)

   Example below shows my test asser in ble_ll_hci.c line 764
    
    > HCI Event: Vendor (0xff) plen 66
       00 72 65 70 6f 73 2f 61 70 61 63 68 65 2d 6d 79 .repos/apache-my
       6e 65 77 74 2d 6e 69 6d 62 6c 65 2f 6e 69 6d 62 newt-nimble/nimb
       6c 65 2f 63 6f 6e 74 72 6f 6c 6c 65 72 2f 73 72 le/controller/sr
       63 2f 62 6c 65 5f 6c 6c 5f 68 63 69 2e 63 fc 02 c/ble_ll_hci.c..
       00 00

### DIFF
--- a/nimble/controller/include/controller/ble_ll.h
+++ b/nimble/controller/include/controller/ble_ll.h
@@ -26,6 +26,11 @@
 #include "nimble/nimble_npl.h"
 #include "controller/ble_phy.h"
 
+#ifdef MYNEWT
+#include "controller/ble_ll_ctrl.h"
+#include "hal/hal_system.h"
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -34,6 +39,23 @@ extern "C" {
 #error 32.768kHz clock required
 #endif
 
+#ifdef MYNEWT
+#ifdef NDEBUG
+#define BLE_LL_ASSERT(cond) (void(0))
+#else
+#define BLE_LL_ASSERT(cond) \
+    if (!(cond)) { \
+        if (hal_debugger_connected()) { \
+            assert(0);\
+        } else {\
+            ble_ll_hci_ev_send_vendor_err(__FILE__, __LINE__); \
+            while(1) {}\
+        }\
+    }
+#endif
+#else
+#define BLE_LL_ASSERT(cond) assert(cond)
+#endif
 /*
  * XXX:
  * I guess this should not depend on the 32768 crystal to be honest. This
--- a/nimble/controller/include/controller/ble_ll_ctrl.h
+++ b/nimble/controller/include/controller/ble_ll_ctrl.h
@@ -279,6 +279,7 @@ int ble_ll_hci_ev_phy_update(struct ble_ll_conn_sm *connsm, uint8_t status);
 void ble_ll_calc_session_key(struct ble_ll_conn_sm *connsm);
 void ble_ll_ctrl_phy_update_proc_complete(struct ble_ll_conn_sm *connsm);
 void ble_ll_ctrl_initiate_dle(struct ble_ll_conn_sm *connsm);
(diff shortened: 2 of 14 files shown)


nimble/controller/include/controller/ble_ll.h
```diff
@@ -34,6 +35,14 @@ extern "C" {
 #error 32.768kHz clock required
 #endif
 
+#define BLE_LL_ASSERT(cond) \
+    if (!(cond)) { \
```
andrzej-kaczmarek: this should be enabled only for debug builds

nimble/controller/include/controller/ble_ll.h
```diff
@@ -34,6 +35,14 @@ extern "C" {
...
+        ble_ll_hci_ev_send_vendor_err(__FILE__, __LINE__); \
+        if (hal_debugger_connected()) { \
```
andrzej-kaczmarek: this is not portable

nimble/controller/include/controller/ble_ll.h
```diff
@@ -34,6 +35,14 @@ extern "C" {
...
+            assert(0);\
```
andrzej-kaczmarek: bkpt instead of assert

nimble/controller/include/controller/ble_ll.h
rymanluk: Well I had such version but then got to conclusion that it would be nice to always know if assert happens. However in order to keep current behavior I will enable it only for debug.

nimble/controller/include/controller/ble_ll.h
rymanluk: and why this is that?  It is HAL.

andrzej-kaczmarek: we do not port HAL and LL should be independent of hw

nimble/controller/include/controller/ble_ll.h
andrzej-kaczmarek: leaving this always enabled means extra code and branching included in build

nimble/controller/include/controller/ble_ll.h
rymanluk: well,  actually I will remove hal_debugger_connected - each architecture will do on assert what  is needed.

rymanluk:
@andrzej-kaczmarek  updated with macro enabled only for debug.

nimble/controller/include/controller/ble_ll.h
andrzej-kaczmarek: perhaps cleaner way to do this is to send an event, flush transport or wait for everything to be sent to host (may need some update in transport) and then just assert. or just send and do infinite loop - calling hal is not really needed here. or simply add `#if MYNEWT` around hal call.

nimble/controller/include/controller/ble_ll.h
```diff
@@ -25,6 +25,8 @@
 #include "nimble/nimble_opt.h"
 #include "nimble/nimble_npl.h"
 #include "controller/ble_phy.h"
+#include "controller/ble_ll_ctrl.h"
```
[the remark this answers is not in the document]
rymanluk: needed

nimble/controller/include/controller/ble_ll.h
```diff
@@ -25,6 +25,8 @@
...
+#include "hal/hal_system.h"
```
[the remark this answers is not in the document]
rymanluk: fixed

PR #115, opened 2018-05-25
debug: Add missing debug logs for LE Meta Events

MariuszSkamra:
This adds debug logs for LE Meta Events.
Those are not too heavy to limit the console output.

@sjanc @rymanluk

### DIFF
--- a/nimble/host/src/ble_hs_dbg.c
+++ b/nimble/host/src/ble_hs_dbg.c
@@ -149,6 +149,188 @@ ble_hs_dbg_le_event_disp(uint8_t subev, uint8_t len, uint8_t *evdata)
                        get_le16(evdata + 1), evdata[3], evdata[4]);
         break;
 
+    case BLE_HCI_LE_SUBEV_DIRECT_ADV_RPT:
+    {
+        struct hci_le_subev_direct_adv_rpt *data = (void *) evdata;
+        struct hci_le_subev_direct_adv_rpt_param *params = data->params;
+
+        if (len < sizeof(*data) ||
+                len < sizeof(*data) + data->num_reports * sizeof(*params)) {
+            BLE_HS_LOG(DEBUG, "Corrupted LE Directed Advertising Report "
+                       "len=%u\n", len);
+            break;
+        }
+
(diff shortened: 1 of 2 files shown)


nimble/host/src/ble_hs_dbg.c
```diff
@@ -149,6 +149,184 @@ ble_hs_dbg_le_event_disp(uint8_t subev, uint8_t len, uint8_t *evdata)
...
+            break;
+        }
+
+        BLE_HS_LOG(DEBUG, "LE Extended Advertising Report len=%u num=0x%02x ",
+                   len, data->num_reports);
+
+        for (i = 0; i < data->num_reports; i++) {
+            BLE_HS_LOG(DEBUG, "[%d]={evttype=0x%04x advlen=%u}\n",
+                       i, le16toh(params->evt_type), params->adv_data_len);
+            params += 1;
```
sjanc: you need to adjust this with adv_data_len so params +=1 is not correct here.

MariuszSkamra: done

MariuszSkamra:
* Added conditional preprocessor directives to exclude Extended Advertising related events if BLE_EXT_ADV is not set

nimble/host/src/ble_hs_dbg.c
```diff
@@ -149,6 +149,188 @@ ble_hs_dbg_le_event_disp(uint8_t subev, uint8_t len, uint8_t *evdata)
...
+        struct hci_le_subev_direct_adv_rpt_param *params = data->params;
+
+        if (len < sizeof(*data) ||
+                len < sizeof(*data) + data->num_reports * sizeof(*params)) {
+            BLE_HS_LOG(DEBUG, "Corrupted LE Directed Advertising Report "
+                       "len=%u\n", len);
+            break;
+        }
+
+        BLE_HS_LOG(DEBUG, "LE Extended Advertising Report len=%u "
```
sjanc: didn't notice that before,  this message should be fixed :)

MariuszSkamra: good catch!

PR #116, opened 2018-05-25
nimble/mesh: Don't send scannable adv events if non-connectable

kolodgrz:
Extended advertising events should be send as non-scannable events.


kolodgrz:
@rymanluk @sjanc @andrzej-kaczmarek @michal-narajowski

PR #120, opened 2018-05-30
transport: uart: Prioritize events

MariuszSkamra:
This adds prioritization of events received. This allows to allocate
membuf for LE Advertising events from low prio mempool.
This patch is needed for host to work in high poluted environment,
where advertisements received may block the host from receiving
other more important events from controller.


MariuszSkamra:
@andrzej-kaczmarek FYI

MariuszSkamra:
Tested with 
* blehci application + Bluez
* custom board with blehci and blemesh application running on separate cores

andrzej-kaczmarek:
lgtm

long term I think we should split UART transport into two separate transports: for host and for controller. this could make code easier to maintain.

PR #122, opened 2018-06-04
Add sample applications from mynewt-core repository

sjanc:


andrzej-kaczmarek:
I would remove `bletest` or move it to some separate folder so someone does not try to use it as a sample app...

andrzej-kaczmarek:
after some discussion offline we'll leave this as it is now - after release we'll do some more cleanup and organize this into samples, tools, test apps etc.

PR #133, opened 2018-06-18
nimble/ll: Return error for unconfigured advertising instances

sjanc:
Trying to enable, remove, set advertising data or scan response to
set that was previoously not configured should result in "Unknown
Advertising Identifier (0x42)" error.


nimble/controller/include/controller/ble_ll_hci.h
```diff
@@ -57,6 +57,8 @@ int ble_ll_hci_send_noop(void);
 /* Checks the preferref phy masks from set default phy and set phy commands */
 int ble_ll_hci_chk_phy_masks(uint8_t *cmdbuf, uint8_t *txphy, uint8_t *rxphy);
```
rymanluk: missing comment here

PR #134, opened 2018-06-20
[RFC] mesh: Check if a model is subscribed to the destination address

michal-narajowski:
Previously we only checked if any of the models within the element have
the destination address in their subscription list. After that check the
specific model that the message was addressed to was not verified to have that
address in it's subscription list. This patch fixes this problem.

### DIFF
--- a/nimble/host/mesh/src/access.c
+++ b/nimble/host/mesh/src/access.c
@@ -390,7 +390,7 @@ static bool model_has_key(struct bt_mesh_model *mod, u16_t key)
 }
 
 static const struct bt_mesh_model_op *find_op(struct bt_mesh_model *models,
-					      u8_t model_count,
+					      u8_t model_count, u16_t dst,
 					      u16_t app_idx, u32_t opcode,
 					      struct bt_mesh_model **model)
 {
@@ -400,6 +400,13 @@ static const struct bt_mesh_model_op *find_op(struct bt_mesh_model *models,
 		const struct bt_mesh_model_op *op;
 
 		*model = &models[i];
+
+		if (BT_MESH_ADDR_IS_GROUP(dst) ||
+		    BT_MESH_ADDR_IS_VIRTUAL(dst)) {
+			if (!bt_mesh_model_find_group(*model, dst)) {
+				continue;
+			}
+		}
 
 		if (!model_has_key(*model, app_idx)) {
 			continue;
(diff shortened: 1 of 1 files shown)


rymanluk:
@jhedberg

nimble/host/mesh/src/access.c
```diff
@@ -401,6 +401,13 @@ static const struct bt_mesh_model_op *find_op(struct bt_mesh_model *models,
 
 		*model = &models[i];
 
+		if (BT_MESH_ADDR_IS_GROUP(dst) ||
+		    BT_MESH_ADDR_IS_VIRTUAL(dst)) {
+			if (!bt_mesh_model_find_group(*model, dst)) {
+				continue;
+			}
+		}
```
jhedberg: Looks good, but should the earlier `bt_mesh_model_find_group()` call in `bt_mesh_model_recv()` be removed? It seems redundant with this addition.

michal-narajowski: Yes, I removed that call now.

nimble/host/mesh/src/access.c
```diff
@@ -494,11 +501,6 @@ void bt_mesh_model_recv(struct bt_mesh_net_rx *rx, struct os_mbuf *buf)
 			if (elem->addr != rx->dst) {
 				continue;
 			}
-		} else if (BT_MESH_ADDR_IS_GROUP(rx->dst) ||
-			   BT_MESH_ADDR_IS_VIRTUAL(rx->dst)) {
-			if (!bt_mesh_elem_find_group(elem, rx->dst)) {
-				continue;
-			}
 		} else if (i != 0 || !bt_mesh_fixed_group_match(rx->dst)) {
```
jhedberg: This is now broken. `bt_mesh_fixed_group_match()` is supposed to be used when you already know the address is a fixed group one. Now that you removed the intermediate checks for `IS_GROUP()` and `IS_VIRTUAL()` those types of addresses will get passed to `bt_mesh_fixed_group_match()` which will always return false for them, causing messages to be ignored before `find_op()` has a chance to be called. What I meant when I said you should remove the redundant check is not just removing the middle `else if` branch, but moving the adress type checks to the last branch so that `fixed_group_match()` is still only called for fixed group addresses.

jhedberg:
See my comment. This should not have been merged. Did you not run any tests on this? As far as I see the code will now ignore all group and virtual destination addresses.

michal-narajowski:
Yeah, my bad. Fixing it now.

PR #136, opened 2018-06-21
mesh: Fix ignoring group and virtual addresses

michal-narajowski:
Previous "fix" broke address processing by ignoring group and virtual addresses.
This one ensures that we correctly handle these addresses and move on to finding
model that subscribed to this type of address inside find_op() function.

### DIFF
--- a/nimble/host/mesh/src/access.c
+++ b/nimble/host/mesh/src/access.c
@@ -501,6 +501,9 @@ void bt_mesh_model_recv(struct bt_mesh_net_rx *rx, struct os_mbuf *buf)
 			if (elem->addr != rx->dst) {
 				continue;
 			}
+		} else if (BT_MESH_ADDR_IS_GROUP(rx->dst) ||
+			   BT_MESH_ADDR_IS_VIRTUAL(rx->dst)) {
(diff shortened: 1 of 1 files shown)


michal-narajowski:
@jhedberg FYI

nimble/host/mesh/src/access.c
```diff
@@ -501,10 +501,15 @@ void bt_mesh_model_recv(struct bt_mesh_net_rx *rx, struct os_mbuf *buf)
 			if (elem->addr != rx->dst) {
 				continue;
 			}
+		} else if (BT_MESH_ADDR_IS_GROUP(rx->dst) ||
+			   BT_MESH_ADDR_IS_VIRTUAL(rx->dst)) {
+			/* find_op will find the correct model for the group */
+			goto proceed;
```
jhedberg: This will work, but I'm not sure I'd have bothered with a goto/label. Instead, I'd just leave this branch empty (but with the code comment).

michal-narajowski: I'm ok with dropping the goto. I wanted to make it clear what's the intention, but a comment is enough.

PR #137, opened 2018-06-27
porting: Use real function pointer types for ISR functions

aykevl:
It appears the only reason `uint32_t` was used, was because of the CMSIS `NVIC_SetVector` function which takes an address as an `uint32_t`. Clean this up for the porting layer.

To be honest, casting to `uint32_t` seems very non-portable to me so using function pointers directly should be better for the future.

Note: I haven't actually tested the FreeRTOS and RIOT changes, but I've used almost identical code in my own port (WIP) so it should be okay.

### DIFF
--- a/nimble/include/nimble/nimble_npl.h
+++ b/nimble/include/nimble/nimble_npl.h
@@ -157,7 +157,7 @@ void ble_npl_time_delay(ble_npl_time_t ticks);
 
 #if NIMBLE_CFG_CONTROLLER
 
-void ble_npl_hw_set_isr(int irqn, uint32_t addr);
+void ble_npl_hw_set_isr(int irqn, void (*addr)());
 
 #endif
 
--- a/nimble/include/nimble/nimble_npl.h
+++ b/nimble/include/nimble/nimble_npl.h
@@ -157,7 +157,7 @@ void ble_npl_time_delay(ble_npl_time_t ticks);
 
 #if NIMBLE_CFG_CONTROLLER
 
-void ble_npl_hw_set_isr(int irqn, void (*addr)());
+void ble_npl_hw_set_isr(int irqn, void (*addr)(void));
 
 #endif
 
--- a/porting/npl/riot/src/nrf5x_isr.c
+++ b/porting/npl/riot/src/nrf5x_isr.c
@@ -19,9 +19,9 @@
 
 #include "cpu.h"
 
-static void (*radio_isr_addr)();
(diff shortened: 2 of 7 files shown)


nimble/include/nimble/nimble_npl.h
```diff
@@ -157,7 +157,7 @@ void ble_npl_time_delay(ble_npl_time_t ticks);
 
 #if NIMBLE_CFG_CONTROLLER
 
-void ble_npl_hw_set_isr(int irqn, uint32_t addr);
+void ble_npl_hw_set_isr(int irqn, void (*addr)());
```
haukepetersen: This triggers a `[-Werror=strict-prototypes]` when building for RIOT. 

Using instead
```c
void ble_npl_hw_set_isr(int irqn, void (*addr)(void));
```
fixes that issue.

porting/npl/riot/src/nrf5x_isr.c
```diff
@@ -19,30 +19,30 @@
 
 #include "cpu.h"
 
-static uint32_t radio_isr_addr;
-static uint32_t rng_isr_addr;
-static uint32_t rtc0_isr_addr;
+static void (*radio_isr_addr)();
+static void (*rng_isr_addr)();
+static void (*rtc0_isr_addr)();
```
haukepetersen: Same here, RIOT complains about `[-Werror=strict-prototypes]`, so please add the missing `void`:
```c
static void (*radio_isr_addr)(void);
static void (*rng_isr_addr)(void);
static void (*rtc0_isr_addr)(void);
```

porting/npl/riot/src/nrf5x_isr.c
```diff
@@ -19,30 +19,30 @@
...
 void
 isr_rtc0(void)
 {
-    ((void (*)(void))rtc0_isr_addr)();
+    rtc0_isr_addr();
 }
 
 void
-ble_npl_hw_set_isr(int irqn, uint32_t addr)
+ble_npl_hw_set_isr(int irqn, void (*addr)())
```
haukepetersen: and here as well:
```c
ble_npl_hw_set_isr(int irqn, void (*addr)(void))
```

andrzej-kaczmarek:
@haukepetersen thanks, for checking this!
@aykevl sorry I forgot about this PR, could you please fix above comments and I'll merge this.

aykevl:
Right. I totally forgot about that oddity in the C language. Fixed it, also in other places.

PR #143, opened 2018-07-10
nimble/host: Add support for SM debug mode

andrzej-kaczmarek:
New syscfg option allows to enable SM debug mode which uses predefined
DH key pair, as per core spec.


andrzej-kaczmarek:
@sjanc yeah, we can handle this in separate PR :)

PR #153, opened 2018-07-29
nimble/ll: Make vnd event on assert configurable

andrzej-kaczmarek:
This event is only useful when talking to external hosts, so no need to
have it enabled in combined builds. Now it's configurable and disabled
by default on all builds except for builds without NimBLE host.


andrzej-kaczmarek:
@jacobrosenthal

PR #159, opened 2018-08-10
Fix build error when data length ext. disabled

ccollins476ad:


nimble/controller/src/ble_ll_hci.c
```diff
@@ -454,6 +454,7 @@ ble_ll_hci_le_rd_sugg_data_len(uint8_t *rspbuf, uint8_t *rsplen)
  *
  * @return int BLE error code
  */
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_DATA_LEN_EXT)
```
andrzej-kaczmarek: you could just move `#endif` above to keep all DLE functions in single block

ccollins476ad: Thanks, Andrzej.  I have made the suggested change.

PR #160, opened 2018-08-10
 Added fixes to make NimBLE work with all ported OS

mehulhirpara:
Suggesting following changes to make NimBLE compile and work correctly with all ported OS. I have tested it with FreeRTOS.

### DIFF
--- a/nimble/host/mesh/src/settings.c
+++ b/nimble/host/mesh/src/settings.c
@@ -10,6 +10,9 @@
 
 #define BT_DBG_ENABLED MYNEWT_VAL(BLE_MESH_DEBUG_SETTINGS)
 
+#include "sysinit/sysinit.h"
+#include "stdio.h"
+#include "stdlib.h"
 #include "mesh/mesh.h"
 #include "mesh/glue.h"
 #include "net.h"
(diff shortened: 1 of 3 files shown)


andrzej-kaczmarek:
I am not sure how this works if we don't have `config/config.h` in porting code anywhere and iirc the `BLE_MESH_SETTINGS` setting added because of this to make. I guess you either have some extra modifications in your port or just do not have `BLE_MESH_SETTINGS`.

Also the commit which changes repo version should not be merged to master - it's from `1_0_0_dev` branch.

mehulhirpara:
Yes, you are right. I do have extra code in my porting layer to handle minimum requirements of config module. Following is the structure for the same:

npl/<os>/include/config/config.h
npl/<os>/src/nimble_config_settings.c

nimble_config_settings.c implements only following APIs in different way than config module in Mynewt.

conf_init()
conf_register()
conf_save_one()
conf_load()

Apart from this, glue.c also needs base64.h from base64 module. Since base64 module is independent of  OS, I am use it as it is from ext/base64.

Regarding version change, I could not able to understand how it went with pull request as I was seeing only two file changes while committing. Let me check how I can exclude it.

mehulhirpara:
After discussing above changes with @rymanluk, I found that above changes are specific to a particular porting case and can be shifted to porting layer (nimble_npl_os.h and numble_port_<os>.c). Therefore, closing PR.

PR #163, opened 2018-08-17
Add support for Nordic RAAL

andrzej-kaczmarek:
Nordic RAAL (Radio Arbiter Abstraction Layer) allows to share nRF5x radio with other drivers like nRF 802.15.4. See: https://github.com/NordicSemiconductor/nRF-IEEE-802.15.4-radio-driver/wiki/Multiprotocol-support

This is initial implementation - it should work, but there are many things that need to be optimized. For now it just schedules 6.4ms slots (configurable) for external driver on a best-effort basis:
image

To enable set `BLE_LL_NRF_RAAL_ENABLE: 1`.

### DIFF
--- a/apps/nrf_raal_test/pkg.yml
+++ b/apps/nrf_raal_test/pkg.yml
@@ -1,0 +1,39 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: apps/nrf_raal_test
+pkg.type: app
+pkg.description: nRF RAAL integration test
+pkg.author: "Andrzej Kaczmarek"
+pkg.email: "andrzej.kaczmarek@codecoup.pl"
+pkg.homepage: "http://mynewt.apache.org/"
+pkg.keywords:
+
+pkg.deps:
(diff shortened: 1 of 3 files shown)


nimble/controller/src/ble_ll_sched.c
```diff
@@ -1237,6 +1252,81 @@ ble_ll_sched_next_time(uint32_t *next_event_time)
     return rc;
 }
 
+uint8_t
+ble_ll_sched_get_current_type(void)
+{
+    if ((int32_t)(g_ble_ll_sched_current.end_time - os_cputime_get32()) > 0) {
```
nutka7: What is the meaning of BLE_LL_SCHED_TYPE_NONE? Shouldn't the condition check for <= 0?

andrzej-kaczmarek: none means that currently no scheduled item is being executed.
the check above is required since we do not have trigger when scheduled item finishes execution thus `g_ble_ll_sched_current` is not reset automatically - we do this here if its end time is past current time.

nimble/controller/src/ble_ll_nrf_raal.c
```diff
@@ -0,0 +1,243 @@
...
+        g_ble_ll_nrf_raal_pending |= BLE_LL_NRF_RAAL_PENDING_SLOT_EXIT;
+    }
+}
+
+static void
+ble_ll_nrf_raal_slot_end_tmr_cb(void *arg)
+{
+    ble_ll_nrf_raal_slot_exit();
+    ble_phy_nrf_raal_slot_exit();
+    /* XXX fixme :) */
```
nutka7: Hey, can you give me some detail on what is the issue here?

andrzej-kaczmarek: I was not sure what to do with rfclk here (i.e. if we can just disable it) so left it as fixme to figure out this later

nimble/controller/src/ble_ll_sched.c
```diff
@@ -1198,7 +1206,14 @@ ble_ll_sched_run(void *arg)
         /* Remove schedule item and execute the callback */
         TAILQ_REMOVE(&g_ble_ll_sched_q, sch, link);
         sch->enqueued = 0;
-        ble_ll_sched_execute_item(sch);
+        rc = ble_ll_sched_execute_item(sch);
```
turon: Should `ble_ll_sched_execute_item` be called *after* updating current slot bounds below?

PR #168, opened 2018-08-24
[build] Adjust makefiles to support autotools integration.

turon:


ccollins476ad:
Looks good (aside from the conflict).  If #169 gets merged, we can close this one, right?

andrzej-kaczmarek:
indeed, looks like the same commit was also part of #169 so I'll close this one - feel free to reopen if something is missing

PR #169, opened 2018-08-24
[linux] Cleanup npl layer in linux port.

turon:
Some minor cleanup to linux port to follow npl include patterns better.
This also fixes a build issue in the npl/example apps.

This PR is prepared as an addendum to #168.


turon:
@andrzej-kaczmarek or @ccollins476ad, can this and #168 be considered for merge?  They are small changes that only affect the Linux port and have been in the queue for 2 months.  Thank you.

PR #177, opened 2018-08-30
nimble: Add API to query the L2CAP channel for details

michal-narajowski:


nimble/host/include/host/ble_l2cap.h
```diff
@@ -188,6 +188,10 @@ int ble_l2cap_connect(uint16_t conn_handle, uint16_t psm, uint16_t mtu,
 int ble_l2cap_disconnect(struct ble_l2cap_chan *chan);
 int ble_l2cap_send(struct ble_l2cap_chan *chan, struct os_mbuf *sdu_tx);
 void ble_l2cap_recv_ready(struct ble_l2cap_chan *chan, struct os_mbuf *sdu_rx);
+uint16_t ble_l2cap_get_scid(struct ble_l2cap_chan *chan);
+uint16_t ble_l2cap_get_dcid(struct ble_l2cap_chan *chan);
+uint16_t ble_l2cap_get_my_mtu(struct ble_l2cap_chan *chan);
```
andrzej-kaczmarek: I did not check if spec has some names for both MTU values but if it does, we should just follow in API names
if it doesn't, this should be e.g. `our_mtu` (as we have `our_addr_type`) or `local_mtu`

nimble/host/src/ble_l2cap_coc.c
```diff
@@ -499,6 +499,34 @@ ble_l2cap_coc_send(struct ble_l2cap_chan *chan, struct os_mbuf *sdu_tx)
     return ble_l2cap_coc_continue_tx(chan);
 }
 
+uint16_t
+ble_l2cap_get_scid(struct ble_l2cap_chan *chan)
```
rymanluk: I suggest to change return type from uint16_t to int and in case of chan==NULL return -1.
This is now API for the application so let's be on the safe side.

rymanluk:
lgtm

PR #178, opened 2018-09-04
host: Add API to unpair a peer device

michal-narajowski:
This API allows to unpair a device with the specified address. The keys
related to that peer device are removed from storage and peer address
is removed from the resolve list from the controller. If a peer
is connected, the connection is terminated.

This function is used in policy for SM key overflow. Instead of just deleting
the keys from storage, the host performs tasks described above to
ensure correct behavior.


michal-narajowski:
@andrzej-kaczmarek Thanks, fixed

PR #181, opened 2018-09-05
nimble/ll: Certification fixes - part 2

rymanluk:


nimble/controller/src/ble_ll_conn_hci.c
```diff
@@ -850,8 +850,11 @@ ble_ll_conn_hci_update(uint8_t *cmdbuf)
         return BLE_ERR_CMD_DISALLOWED;
     }
 
-    /* See if this feature is supported on both sides */
-    if ((connsm->conn_features & BLE_LL_FEAT_CONN_PARM_REQ) == 0) {
+    /* See if this feature is supported on both sides.
+     * If features has not be exchanged let us give a try and send request
+     */
+    if (connsm->conn_features &&
+        (connsm->conn_features & BLE_LL_FEAT_CONN_PARM_REQ) == 0) {
```
andrzej-kaczmarek: I don't understand why this is required - it does not seem right.

nimble/controller/src/ble_ll_conn.c
```diff
@@ -3907,7 +3907,8 @@ ble_ll_conn_rx_isr_end(uint8_t *rxbuf, struct ble_mbuf_hdr *rxhdr)
         }
 
         /* If we received a terminate IND, we must set some flags */
-        if (is_ctrl && (opcode == BLE_LL_CTRL_TERMINATE_IND)) {
+        if (is_ctrl && (opcode == BLE_LL_CTRL_TERMINATE_IND)
+                    && (rx_pyld_len == BLE_LL_CTRL_TERMINATE_IND_LEN)) {
```
andrzej-kaczmarek: I think check for length should be in nested `if` since we should not go to `else` statements in this case

nimble/controller/include/controller/ble_ll_scan.h
```diff
@@ -103,7 +103,7 @@ struct ble_ll_aux_data {
     uint8_t mode;
     uint8_t scanning;
     uint8_t flags;
-    uint16_t did;
+    uint16_t aid;
```
andrzej-kaczmarek: `adi`

nimble/controller/src/ble_ll_scan.c
```diff
@@ -1809,6 +1809,8 @@ ble_ll_scan_parse_ext_hdr(struct os_mbuf *om, struct ble_mbuf_hdr *ble_hdr,
     if (ext_hdr_flags & (1 << BLE_LL_EXT_ADV_DATA_INFO_BIT)) {
         ble_ll_ext_scan_parse_adv_info(scansm, out_evt, (ext_hdr + i));
         i += BLE_LL_EXT_ADV_DATA_INFO_SIZE;
+    } else if ((out_evt->evt_type & BLE_HCI_ADV_SCAN_RSP_MASK)) {
```
andrzej-kaczmarek: no need for double parentheses

nimble/controller/src/ble_ll_scan.c
```diff
@@ -1773,6 +1773,17 @@ ble_ll_scan_parse_ext_hdr(struct os_mbuf *om, struct ble_mbuf_hdr *ble_hdr,
         return -1;
     }
 
+    if (BLE_MBUF_HDR_SCAN_RSP_RCV(ble_hdr)) {
+        out_evt->evt_type |= BLE_HCI_ADV_SCAN_RSP_MASK;
+    }
+
+    if (aux_data) {
+        /* Store event type */
+        aux_data->evt_type |= out_evt->evt_type;
```
andrzej-kaczmarek: I think if you want to store, you should do `=` (unless this does something else)

nimble/controller/src/ble_ll_scan.c
```diff
@@ -1773,6 +1773,17 @@ ble_ll_scan_parse_ext_hdr(struct os_mbuf *om, struct ble_mbuf_hdr *ble_hdr,
...
+    }
+
+    out_evt->evt_type = aux_data->evt_type;
```
andrzej-kaczmarek: this should be inside `if`

nimble/controller/src/ble_ll_conn_hci.c
rymanluk: well this patch is indeed not needed as our controller always does feature exchange. Will remove it.

nimble/controller/src/ble_ll_conn.c
rymanluk: As discussed off line, it make sense to do so

nimble/controller/src/ble_ll_scan.c
rymanluk: actually I should "store" it later as "directed" bit is added below in this function.

PR #183, opened 2018-09-05
nimble: Build error fixes after PR 164 and PR 182

rymanluk:


rymanluk:
Merging as those are trivial build fixes

PR #184, opened 2018-09-06
DTM: few improvements

andrzej-kaczmarek:
Few improvements so we do not assert in cases which can happen during DTM.

### DIFF
--- a/nimble/controller/src/ble_ll_dtm.c
+++ b/nimble/controller/src/ble_ll_dtm.c
@@ -578,4 +578,16 @@ ble_ll_dtm_reset(void)
 {
     ble_ll_dtm_ctx_free(&g_ble_ll_dtm_ctx);
 }
+
+void
+ble_ll_dtm_init(void)
+{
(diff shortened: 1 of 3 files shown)


nimble/controller/src/ble_ll_dtm.c
```diff
@@ -23,12 +23,24 @@
...
 #include "assert.h"
 #include "os/os.h"
+#include "stats/stats.h"
 #include "controller/ble_ll.h"
 #include "controller/ble_phy.h"
 #include "controller/ble_ll_sched.h"
 #include "controller/ble_ll_xcvr.h"
 #include "ble_ll_dtm_priv.h"
 
+STATS_SECT_START(ble_ll_dtm_stats)
```
mkiiskila: These stats are not getting registered.

andrzej-kaczmarek: right, I usually view stats directly on debugger so forgot about this. will fix :)

PR #185, opened 2018-09-10
nimble/host: Add listeners for GAP events

andrzej-kaczmarek:
This allow any package to registere listener which will receive GAP
events. The events applicable for listener are any "notifications",
i.e. events which are informational only. This means that e.g. events
like connection state changes are notified while pairing request is
not.

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -185,6 +185,8 @@ struct ble_gap_snapshot {
     void *cb_arg;
 };
 
+static SLIST_HEAD(ble_gap_hook_list, ble_gap_event_listener) ble_gap_event_listener_list;
+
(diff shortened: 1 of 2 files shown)


nimble/host/src/ble_gap.c
```diff
@@ -4640,6 +4659,76 @@ ble_gap_preempt_done(void)
...
+        }
+    } else {
+        rc = BLE_HS_EALREADY;
+    }
+
+    return rc;
+}
+
+int
+ble_gap_event_listener_unregiter(struct ble_gap_event_listener *listener)
```
ccollins476ad: Typo: missing an 's'.

andrzej-kaczmarek:
@rymanluk I will, in separate PR.

PR #192, opened 2018-09-21
nimble/gap: When BLE_EXT_ADV is on use only EXT ADV HCI commands

rymanluk:


rymanluk:
rebased to trigger Travis

PR #193, opened 2018-09-22
nimble/host: Improve BLE uuid comparison, handle invalid uuid's

jeremywood:
This PR contains a couple of changes to UUID checking and comparison to address issues we've seen with malformed UUID's in the field. We also needed to check UUID's in our application, so we made that function available externally.

Rename verify_uuid to ble_uuid_check and make it available to other files. Improve ble_uuid_cmp so that it returns unequal if the UUID type is unrecognized. Use UUID check in ble_gattc.c to resolve errors with invalid UUID's.

### DIFF
--- a/nimble/host/src/ble_gattc.c
+++ b/nimble/host/src/ble_gattc.c
@@ -2453,7 +2453,8 @@ done:
     if (rc != 0) {
         /* Failure. */
         cbrc = ble_gattc_disc_chr_uuid_cb(proc, rc, 0, NULL);
-    } else if (ble_uuid_cmp(&chr.uuid.u, &proc->disc_chr_uuid.chr_uuid.u) == 0) {
+    } else if (ble_uuid_check(&chr.uuid.u) == 0 &&
+               ble_uuid_cmp(&chr.uuid.u, &proc->disc_chr_uuid.chr_uuid.u) == 0) {
         /* Requested characteristic discovered. */
         cbrc = ble_gattc_disc_chr_uuid_cb(proc, 0, 0, &chr);
     } else {
--- a/nimble/host/include/host/ble_uuid.h
+++ b/nimble/host/include/host/ble_uuid.h
@@ -171,6 +171,15 @@ char *ble_uuid_to_str(const ble_uuid_t *uuid, char *dst);
  */
 uint16_t ble_uuid_u16(const ble_uuid_t *uuid);
 
+/** @brief Checks a UUID object for validity.
+ *
+ * @param uuid      The UUID object to check.
+ *
+ * @return          0 on success;
+ *                  BLE_HS_EBADDATA if the provided UUID object is not valid.
+ */
+int ble_uuid_check(const ble_uuid_t *uuid);
+
 #ifdef __cplusplus
 }
 #endif
--- a/nimble/host/src/ble_uuid.c
+++ b/nimble/host/src/ble_uuid.c
@@ -32,8 +32,8 @@ static uint8_t ble_uuid_base[16] = {
     0x00, 0x10, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
 };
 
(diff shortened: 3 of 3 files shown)


nimble/host/src/ble_gattc.c
```diff
@@ -2453,7 +2453,8 @@ ble_gattc_disc_chr_uuid_rx_adata(struct ble_gattc_proc *proc,
     if (rc != 0) {
         /* Failure. */
         cbrc = ble_gattc_disc_chr_uuid_cb(proc, rc, 0, NULL);
-    } else if (ble_uuid_cmp(&chr.uuid.u, &proc->disc_chr_uuid.chr_uuid.u) == 0) {
+    } else if (ble_uuid_check(&chr.uuid.u) == 0 &&
+               ble_uuid_cmp(&chr.uuid.u, &proc->disc_chr_uuid.chr_uuid.u) == 0) {
```
andrzej-kaczmarek: What does it mean 'malformed UUID' in this case? UUID received over the air is checked earlier so malformed UUID should be already rejected. One thing that `ble_uuid_init_from_buf` does not check is whether 16-bit UUID was not transferred as 128-bit UUID over the air which is incorrect and this is what should be fixed.

jeremywood:
This is a solution to a problem we've encountered with invalid UUID's in the field, where the invalid UUID, from an external source, caused issues in the stack and in our application.

If the correct place to fix this is in `ble_uuid_init_from_buf`, that would be fine, assuming that's the only point of entry for an external UUID.

andrzej-kaczmarek:
I do not know if fixing `ble_uuid_init_from_buf` will resolve your problem because you did not provide any details... what is "invalid UUID" here? What was the request/response to/from peer?

We either need fix with these information to be able to tell whether it's correct or just create an issue with all the details and we'll try to fix it.

jeremywood:
I’ve asked my colleague @simonratner to chime in on this, as he worked this issue originally.

(I took responsibility to make this PR as part of our efforts to update to the latest Nimble, which want to do without any patches.)

simonratner:
@andrzej-kaczmarek This is something we saw in field deployment tests, so unfortunately we do not have a capture of request/response to go with this. Here's what I did do:
* Track down the problem to the `verify_uuid` check inside `ble_uuid_cmp`, based on `pc` values and stack traces from the field.
* Test a local reproduction by running an app that advertises a 16-bit-in-128-bit UUID, which is invalid, and observe nearby mynewt devices fail. Note that this is not necessarily the only way in which this can fail, but it was just one incarnation we could repro (I had to disable some of the checks you mention to make mynewt advertise an invalid UUID in the first place.)
* Test a local fix whereby every call to `ble_uuid_cmp` that operates on external data (as opposed to app-generated data) is guarded by a check for UUID validity; this is the change you see here.
* Confirm that deploying this change to the field has stopped the problem.

Again, I want to emphasize that I have no guarantee that 16-bit-in-128-bit UUID advertisement is the only way this can fail. In theory, any input that would result in `ble_uuid_cmp` asserting in a debug build will be a problem, and should be addressed. Verifying that there are no other entry points for an invalid external UUID is not something I could grep for, so your input here would be appreciated. Some things that spring to mind: disc of service by UUID, disc of characteristic by UUID (incl. read/write by UUID), as well as their gatt server counterparts whereby a peer requests a read/write of an invalid UUID.

So with this information, what would you suggest as an appropriate fix?

andrzej-kaczmarek:
@simonratner Thanks for explanation. So the problem is indeed 16-bit UUID received as 128-bit one because it's allowed to "enter" the stack and then fails on sanity check. To solve this we'll just convert such UUID to 16-bit on receive so that it will pass any sanity check later. I'll prepare patch for this soon.

simonratner:
Let me know if we can assist.

jeremywood:
This was superseded by PR #198.

PR #197, opened 2018-09-26
Fix the wrong way to detect the trailing space

forgithub123:
Assignment n bytes to  link.rx.buf->om_len will lead to n bytes decrease of OS_MBUF_TRAILINGSPACE(link.rx.buf), so we should get the OS_MBUF_TRAILINGSPACE(link.rx.buf) before assignment net_buf_simple_pull_be16(buf) to link.rx.buf->om_len.

### DIFF
--- a/nimble/host/mesh/src/prov.c
+++ b/nimble/host/mesh/src/prov.c
@@ -1375,6 +1375,8 @@ static void gen_prov_ack(struct prov_rx *rx, struct os_mbuf *buf)
 
 static void gen_prov_start(struct prov_rx *rx, struct os_mbuf *buf)
 {
+	u16_t trailing_space = 0;
+	
 	if (link.rx.seg) {
 		BT_WARN("Got Start while there are unreceived segments");
 		return;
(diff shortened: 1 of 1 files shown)


nimble/host/mesh/src/prov.c
```diff
@@ -1385,6 +1385,8 @@ static void gen_prov_start(struct prov_rx *rx, struct os_mbuf *buf)
 		gen_prov_ack_send(rx->xact_id);
 		return;
 	}
+	
+	u16_t trailing_space = OS_MBUF_TRAILINGSPACE(link.rx.buf);
```
rymanluk: Good catch! Thanks. I have couple of comments:
1) Please prefix commit message with "mesh:". Having this it would bring my attention much faster :)
2) Please explain in commit message why we need this change. Comment from PR will not stay in git.
3) Please move `trailing_space` declaration on top of the function

forgithub123: Thanks for your comments, I will add the comments to the commit message.

PR #198, opened 2018-09-27
nimble/host: Fixes for UUID handling

andrzej-kaczmarek:
This is more generic approach for issue described in https://github.com/apache/mynewt-nimble/pull/193.

### DIFF
--- a/nimble/host/src/ble_gattc.c
+++ b/nimble/host/src/ble_gattc.c
@@ -1441,7 +1441,8 @@ ble_gattc_disc_all_svcs_rx_adata(struct ble_gattc_proc *proc,
     switch (adata->value_len) {
     case 2:
     case 16:
-        rc = ble_uuid_init_from_buf(&service.uuid, adata->value, adata->value_len);
+        rc = ble_uuid_init_from_att_buf(&service.uuid, adata->value,
+                                        adata->value_len);
         if (rc != 0) {
             rc = BLE_HS_EBADDATA;
             goto done;
@@ -1887,7 +1888,7 @@ ble_gattc_find_inc_svcs_rx_read_rsp(struct ble_gattc_proc *proc, int status,
 
     ble_gattc_dbg_assert_proc_not_inserted(proc);
 
-    rc = ble_uuid_init_from_mbuf(&service.uuid, *om, 0, 16);
+    rc = ble_uuid_init_from_att_mbuf(&service.uuid, *om, 0, 16);
     os_mbuf_free_chain(*om);
     *om = NULL;
 
@@ -1964,6 +1965,8 @@ ble_gattc_find_inc_svcs_rx_adata(struct ble_gattc_proc *proc,
 
     proc->find_inc_svcs.prev_handle = adata->att_handle;
 
+    rc = 0;
+
     switch (adata->value_len) {
     case BLE_GATTS_INC_SVC_LEN_NO_UUID:
         proc->find_inc_svcs.cur_start = get_le16(adata->value + 0);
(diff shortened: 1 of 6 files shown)


nimble/host/src/ble_gattc.c
```diff
@@ -1974,7 +1975,7 @@ ble_gattc_find_inc_svcs_rx_adata(struct ble_gattc_proc *proc,
     case BLE_GATTS_INC_SVC_LEN_UUID:
         service.start_handle = get_le16(adata->value + 0);
         service.end_handle = get_le16(adata->value + 2);
-        ble_uuid_init_from_buf(&service.uuid, adata->value + 4, 2);
+        ble_uuid_init_from_att_buf(&service.uuid, adata->value + 4, 2);
```
simonratner: Should this check return code? Technically the length is always `2` so will always succeed, but seems safer to check in case `ble_uuid_init_from_att_buf` is changed in the future.

nimble/host/src/ble_uuid.c
```diff
@@ -74,8 +65,8 @@ ble_uuid_init_from_buf(ble_uuid_any_t *uuid, const void *buf, size_t len)
 int
 ble_uuid_cmp(const ble_uuid_t *uuid1, const ble_uuid_t *uuid2)
```
simonratner: Also, this change is small, but important: https://github.com/apache/mynewt-nimble/pull/193/files#diff-fa8c3d9c49d86b38eeeca3239b039642R94

It feels safer to compare things as unequal if you don't know how to compare them (cf. Java comparing things of incompatible types as unequal). If you are running without `BLE_HS_DEBUG`, returning zero by default will have surprising results downstream.

simonratner:
@andrzej-kaczmarek Any way i can help get this merged? Looks like the tests are failing because they didn't get updated for the function rename:
```
nimble/host/test/src/ble_hs_test_util.c: In function ‘ble_hs_test_util_verify_tx_find_info_rsp’:
nimble/host/test/src/ble_hs_test_util.c:1268:13: error: implicit declaration of function ‘ble_uuid_init_from_mbuf’; did you mean ‘ble_uuid_init_from_buf’? [-Werror=implicit-function-declaration]
             ble_uuid_init_from_mbuf(&uuid, om, off, 2);
             ^~~~~~~~~~~~~~~~~~~~~~~
             ble_uuid_init_from_buf
```

nimble/host/src/ble_uuid.c
```diff
@@ -33,21 +33,12 @@ static uint8_t ble_uuid_base[16] = {
...
 
 #if MYNEWT_VAL(BLE_HS_DEBUG)
-static int verify_uuid(const ble_uuid_t *uuid)
-{
-    switch (uuid->type) {
-    case BLE_UUID_TYPE_16:
-    case BLE_UUID_TYPE_32:
-        return 0;
-    case BLE_UUID_TYPE_128:
-        if (memcmp(BLE_UUID128(uuid)->value, ble_uuid_base, 12) != 0) {
```
rymanluk: since `verify_uuid` was under `BLE_HS_DEBUG` I'm not sure why we want to remove this `memcmp`.
Maybe instead of removing it we could print some error? It would be nice to catch that remote behaves incorrectly when sending Bluetooth UUIDs as 128bits value.

andrzej-kaczmarek: if remote behaves incorrectly it will just fail and you will find the root cause in logs anyway

rymanluk: that is fine then

PR #200, opened 2018-09-28
Modified iBeacon and Eddystone

adityaxavier:
Modified ble_ibeacon_set_adv_data function to take in struct ble_hs_adv_fields as an argument, similar to ble_eddystone_set_adv_data_url function for Eddystone. This would enable users to set ble_hs_adv_fields with any additional parameters like name etc.

### DIFF
--- a/apps/btshell/src/cmd.c
+++ b/apps/btshell/src/cmd.c
@@ -1752,6 +1752,12 @@ cmd_set_adv_data_or_scan_rsp(int argc, char **argv, bool scan_rsp)
         return rc;
     }
 
+    int8_t measured_power = parse_arg_uint16("measured_power", &rc);
+    if (rc != 0 && rc != ENOENT) {
+        console_printf("invalid 'measured_power' parameter\n");
+        return rc;
+    }
+
     eddystone_url_full = parse_arg_extract("eddystone_url");
     if (eddystone_url_full != NULL) {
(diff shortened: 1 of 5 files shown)


andrzej-kaczmarek:
iBeacon uses 30 (out of 31) bytes of advertising data so not sure how would you like to fit more data there?

adityaxavier:
> iBeacon uses 30 (out of 31) bytes of advertising data so not sure how would you like to fit more data there?

Am sorry I don't understand what you are saying. Previously, tx_pwr (intt8_t) was being appended in the buf for both iBeacon and Eddystone. This was as protocol (byte 25-26) were meant for Measured Power. However the problem was it was being generated automatically instead of taking as an argument from the User. This should have been a user defined parameter as this value depends on HW , antenna design, Casing etc. This change is done on both iBeacon and Eddystone.
Apart from that commit, I have reused the ble_hs_adv_fields in iBeacon similar to what was done in Eddystone. This allows us to pass along ble_hs_adv_fields parameters instead of overwriting them, for e.g in my use case I wanted BLE to have a name.

I hope this clears this up.

adityaxavier:
Probably could have written the commit message better..

adityaxavier:
You are correct. (Though I don’t know what’s the use case in Eddystone) I will close this commit and send another PR for TX_Power.

PR #201, opened 2018-09-30
Implemented Measured Power as an Argument

adityaxavier:
Measured Power was being generated automatically and appended into the advertising buffer in both Eddystone and iBeacon.
Changed it to take it as an argument from the User, as it depends on Antenna Hardware, Case design etc.

Modified btshell application to utilize the same.

### DIFF
--- a/apps/btshell/src/cmd.c
+++ b/apps/btshell/src/cmd.c
@@ -1752,6 +1752,15 @@ cmd_set_adv_data_or_scan_rsp(int argc, char **argv, bool scan_rsp)
         return rc;
     }
 
+    int8_t measured_power;
+    uint8_t measured_power_uint = parse_arg_uint8_dflt("measured_power", 0, rc);
+    if(rc == 0){
+        measured_power = -1 * measured_power_uint;
+    }else if(rc != ENOENT){
+        console_printf("invalid 'measured_power' parameter\n");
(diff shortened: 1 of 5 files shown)


apps/btshell/src/cmd.c
```diff
@@ -1752,6 +1752,15 @@ cmd_set_adv_data_or_scan_rsp(int argc, char **argv, bool scan_rsp)
...
     }
 
+    int8_t measured_power;
+    uint8_t measured_power_uint = parse_arg_uint8_dflt("measured_power", 0, rc);
+    if(rc == 0){
+        measured_power = -1 * measured_power_uint;
+    }else if(rc != ENOENT){
+        console_printf("invalid 'measured_power' parameter\n");
+        return rc;
+    }
```
andrzej-kaczmarek: there are plenty of coding style issues in this small piece of code - please check `CODING_STANDARDS.md` file in repository and fix everywhere

adityaxavier: So you want me to send another PR with all the changes in one commit ? or would this do for now ?
I'll make changes wrt coding standards.

andrzej-kaczmarek: just fix and cleanup commits on your local branch and force push to the same branch on github to update PR

PR #206, opened 2018-10-04
Add support for Eddystone TLM Plain

adityaxavier:
Added support for Eddystone TLM Plain.
PDU Interweaving with Eddystone URL / UID, PDU Count and Time since power up are left for the user to handle to keep it platform agnostic.

### DIFF
--- a/nimble/host/include/host/ble_eddystone.h
+++ b/nimble/host/include/host/ble_eddystone.h
@@ -106,6 +106,29 @@ int ble_eddystone_set_adv_data_url(struct ble_hs_adv_fields *adv_fields,
                                    uint8_t url_body_len, uint8_t suffix,
                                    int8_t measured_power);
 
+/**
+ * Configures the device to advertise Eddystone TLM (Unencrypted) beacons.
+ *
+ * @param adv_fields            The base advertisement fields to transform into
+ *                                  an eddystone beacon.  All configured fields
+ *                                  are preserved; you probably want to clear
+ *                                  this struct before calling this function.
+ * @param battery_voltage       The 2-byte battery voltage in mV to advertise.
+ * @param temperature           The 2-byte temperature to advertise.
+ * @param adv_pdu_count         The 4-byte total packet count to advertise.
+ * @param time_since_power_up   The 4-byte count in 0.1 Seconds since the device 
+ *                                  started / rebooted.
+ *
+ * @return                      0 on success;
+ *                              BLE_HS_EBUSY if advertising is in progress;
+ *                              BLE_HS_EMSGSIZE if the specified data is too
+ *                                  large to fit in an advertisement;
+ *                              Other nonzero on failure.
+ */
+int ble_eddystone_set_adv_data_tlm_unencrypted(struct ble_hs_adv_fields *adv_fields,
+                                               uint16_t battery_voltage, float temperature,
+                                               uint32_t adv_pdu_count, uint32_t time_since_power_up);
+
 #ifdef __cplusplus
 }
 #endif
--- a/nimble/host/src/ble_eddystone.c
+++ b/nimble/host/src/ble_eddystone.c
@@ -29,6 +29,7 @@
 
 #define BLE_EDDYSTONE_FRAME_TYPE_UID    0x00
 #define BLE_EDDYSTONE_FRAME_TYPE_URL    0x10
+#define BLE_EDDYSTONE_FRAME_TYPE_TLM_UNENCRYPTED    0x20
 
 static ble_uuid16_t ble_eddystone_uuids16[BLE_EDDYSTONE_MAX_UUIDS16 + 1];
 static uint8_t ble_eddystone_svc_data[BLE_EDDYSTONE_MAX_SVC_DATA_LEN];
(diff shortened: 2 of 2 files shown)


nimble/host/src/ble_eddystone.c
```diff
@@ -29,6 +29,7 @@
 
 #define BLE_EDDYSTONE_FRAME_TYPE_UID    0x00
 #define BLE_EDDYSTONE_FRAME_TYPE_URL    0x10
+#define BLE_EDDYSTONE_FRAME_TYPE_TLM_UNENCRYPTED    0x20
```
andrzej-kaczmarek: frame type is the same for any TLM so name is misleading

nimble/host/src/ble_eddystone.c
```diff
@@ -175,3 +176,44 @@ ble_eddystone_set_adv_data_url(struct ble_hs_adv_fields *adv_fields,
...
+    int rc;
+
+    svc_data = ble_eddystone_set_svc_data_base(BLE_EDDYSTONE_FRAME_TYPE_TLM_UNENCRYPTED);
+
+    /* First byte is TLM Version - 00 for Unencrypted. */
+    svc_data[0] = 00;
+
+    /* Second paramter is Battery Voltage in mV. */
+    put_be16(temp_data, battery_voltage);
+    memcpy(svc_data + 1, temp_data, 2);
```
andrzej-kaczmarek: you can write data directly to destination buffer, not need for extra copy

nimble/host/src/ble_eddystone.c
```diff
@@ -175,3 +176,44 @@ ble_eddystone_set_adv_data_url(struct ble_hs_adv_fields *adv_fields,
...
+    /* Fifth paramter is Alive Time (Total number of 0.1 Seconds since the device booted / rebooted). */
+    put_be32(temp_data, time_since_power_up);
+    memcpy(svc_data + 9, temp_data, 4);
+
+    rc = ble_eddystone_set_adv_data_gen(adv_fields, 13);
+    if (rc != 0) {
+        return rc;
+    }
+
+    return 0;
```
andrzej-kaczmarek: `return rc` will do exactly the same

nimble/host/include/host/ble_eddystone.h
```diff
@@ -106,6 +106,29 @@ int ble_eddystone_set_adv_data_url(struct ble_hs_adv_fields *adv_fields,
...
+ *
+ * @param adv_fields            The base advertisement fields to transform into
+ *                                  an eddystone beacon.  All configured fields
+ *                                  are preserved; you probably want to clear
+ *                                  this struct before calling this function.
+ * @param battery_voltage       The 2-byte battery voltage in mV to advertise.
+ * @param temperature           The 2-byte temperature to advertise.
+ * @param adv_pdu_count         The 4-byte total packet count to advertise.
+ * @param time_since_power_up   The 4-byte count in 0.1 Seconds since the device 
+ *                                  started / rebooted.
```
andrzej-kaczmarek: internal byte count is irrelevant in API description (and confusing)

nimble/host/include/host/ble_eddystone.h
```diff
@@ -106,6 +106,29 @@ int ble_eddystone_set_adv_data_url(struct ble_hs_adv_fields *adv_fields,
...
+ *
+ * @return                      0 on success;
+ *                              BLE_HS_EBUSY if advertising is in progress;
```
andrzej-kaczmarek: I don't think this will ever return `BLE_HS_EBUSY` since it's perfectly fine to change advertising data while advertising is in progress

nimble/host/include/host/ble_eddystone.h
```diff
@@ -106,6 +106,29 @@ int ble_eddystone_set_adv_data_url(struct ble_hs_adv_fields *adv_fields,
...
+ *                              BLE_HS_EMSGSIZE if the specified data is too
+ *                                  large to fit in an advertisement;
+ *                              Other nonzero on failure.
+ */
+int ble_eddystone_set_adv_data_tlm_unencrypted(struct ble_hs_adv_fields *adv_fields,
+                                               uint16_t battery_voltage, float temperature,
```
andrzej-kaczmarek: using `float` is probably not the best idea since this can run on devices without fpu so I'd either use original 8.8 fixed point notation or just use .001 deg units.

nimble/host/src/ble_eddystone.c
adityaxavier: Again, I only followed what other implementations in Eddystone are doing. Though I agree with you that return rc is better than doing return 0.

nimble/host/include/host/ble_eddystone.h
adityaxavier: I assumed it is possible that the User stops advertising and advertises TLM data, hence the similarity in returns.

nimble/host/include/host/ble_eddystone.h
adityaxavier: Could you explain what you mean by original 8.8 fixed point ? 8.8 is a float value right ? with 1 byte being an integer and another byte being the fraction ? 
So do you want me to take 2 uin8_t from the user? 
Also I didn't understand what you mean by .001 deg unit .

nimble/host/include/host/ble_eddystone.h
adityaxavier: Understood, will change it.

nimble/host/src/ble_eddystone.c
adityaxavier: Understood, will change it.

PR #215, opened 2018-10-09
host/gap: Add connection find by address API

michal-narajowski:
This may be useful in some cases.

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -436,6 +436,28 @@ ble_gap_conn_find(uint16_t handle, struct ble_gap_conn_desc *out_desc)
     }
 }
 
+int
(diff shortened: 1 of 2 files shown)


nimble/host/src/ble_gap.c
```diff
@@ -436,6 +436,28 @@ ble_gap_conn_find(uint16_t handle, struct ble_gap_conn_desc *out_desc)
...
+	conn = ble_hs_conn_find_by_addr(addr);
+	if (conn != NULL && out_desc != NULL) {
+		ble_gap_fill_conn_desc(conn, out_desc);
+	}
+
+	ble_hs_unlock();
+
+	if (conn == NULL) {
+		return BLE_HS_ENOTCONN;
+	} else {
```
rymanluk: nitpick: no need for else

PR #218, opened 2018-10-10
nimble/ll: Certification fixes part3

rymanluk:


nimble/controller/src/ble_ll_ctrl.c
```diff
@@ -1584,7 +1584,7 @@ ble_ll_ctrl_rx_conn_update(struct ble_ll_conn_sm *connsm, uint8_t *dptr)
 
     /* Only a slave should receive this */
     if (connsm->conn_role == BLE_LL_CONN_ROLE_MASTER) {
-        return BLE_ERR_MAX;
+        return BLE_LL_CTRL_UNKNOWN_RSP;
```
andrzej-kaczmarek: I think we should reply with `LL_REJECT_IND` since this is not an unknown PDU, it's just invalid in current role

nimble/controller/src/ble_ll_ctrl.c
```diff
@@ -1848,24 +1848,28 @@ ble_ll_ctrl_rx_version_ind(struct ble_ll_conn_sm *connsm, uint8_t *dptr,
...
  */
-static void
+static int
 ble_ll_ctrl_rx_chanmap_req(struct ble_ll_conn_sm *connsm, uint8_t *dptr)
 {
     uint16_t instant;
     uint16_t conn_events;
 
+    if (connsm->conn_role == BLE_LL_CONN_ROLE_MASTER) {
+        return BLE_LL_CTRL_UNKNOWN_RSP;
```
andrzej-kaczmarek: same here

nimble/controller/src/ble_ll_conn.c
```diff
@@ -1038,8 +1038,19 @@ ble_ll_conn_adjust_pyld_len(struct ble_ll_conn_sm *connsm, uint16_t pyld_len)
     uint16_t ret;
 
 #if (BLE_LL_BT5_PHY_SUPPORTED == 1)
+    uint8_t phy_mode;
+
+    if (connsm->phy_tx_transition != BLE_PHY_TRANSITION_INVALID) {
+        phy_mode = ble_ll_phy_to_phy_mode(connsm->phy_tx_transition,
+                                          connsm->phy_data.phy_options);
+    } else {
+        phy_mode = ble_ll_phy_to_phy_mode(connsm->phy_data.tx_phy_mode,
```
andrzej-kaczmarek: this does not seem right since `tx_phy_mode` already stores PHY mode so it should not be used as input to `ble_ll_phy_to_phy_mode`

nimble/controller/src/ble_ll_ctrl.c
```diff
@@ -2540,6 +2542,17 @@ ble_ll_ctrl_tx_done(struct os_mbuf *txpdu, struct ble_ll_conn_sm *connsm)
             connsm->enc_data.enc_state = CONN_ENC_S_PAUSE_ENC_RSP_WAIT;
         }
         break;
+    case BLE_LL_CTRL_PHY_REQ:
+        if (connsm->conn_role == BLE_LL_CONN_ROLE_SLAVE) {
+            if (connsm->phy_data.req_pref_tx_phys_mask & BLE_PHY_MASK_1M) {
+                connsm->phy_tx_mod_transition = BLE_PHY_1M;
```
andrzej-kaczmarek: `phy_tx_mod_transition` is not defined anywhere in this patch.
also what `mod` means? if this is typo in `mode` then you should not use `BLE_PHY_1M` here because this is not a mode

nimble/controller/src/ble_ll_conn.c
```diff
@@ -3698,6 +3698,19 @@ ble_ll_conn_rx_data_pdu(struct os_mbuf *rxpdu, struct ble_mbuf_hdr *hdr)
...
+ble_ll_conn_check_phy_update(struct ble_ll_conn_sm *connsm, uint8_t opcode, uint8_t *ctrdata)
+{
+    uint8_t tx_phy;
+
+    tx_phy = ctrdata[0];
+
+    if (opcode == BLE_LL_CTRL_PHY_UPDATE_IND) {
+        connsm->phy_tx_transition = tx_phy;
+        return;
+    }
```
andrzej-kaczmarek: this function is a bit weird and not only because of this redundant `return`, but also it does not return anything so in case `opcode` does not match it won't touch `ctrdata` so not sure how caller should know about this

nimble/controller/src/ble_ll_conn.c
rymanluk: it is not right for sure. It should be phy_mode = connsm->phy_data.tx_phy_mode
will fix

nimble/controller/src/ble_ll_ctrl.c
rymanluk: well and what error code would you used then?

On the other side you probably agree that this LL_CTRL_PDU is not used in the given role.
That said look at this from Bluetooth 5 spec:

"If an LL Control PDU is received that is not used or not supported, the Link
Layer shall respond with an LL_UNKNOWN_RSP PDU"

andrzej-kaczmarek: so let's leave `LL_UNKNOWN_RSP` then - the quotation from spec is what I was missing here and could not find this myself

nimble/controller/include/controller/ble_ll.h
```diff
@@ -229,6 +229,15 @@ extern STATS_SECT_DECL(ble_ll_stats) ble_ll_stats;
 #define BLE_LL_FEAT_LE_POWER_CLASS_1 (0x00008000)
 #define BLE_LL_FEAT_MIN_USED_CHAN    (0x00010000)
 
+/* This is initial mask, so if feature exchange will not happen,
+ * but host will want to use this procedure, we will try. If not
+ * succeed, feature bit will be cleared.
+ * Look at LL Features above to find out what is allowed
+ */
+#define BLE_LL_FEAT_INITIAL          (0x00000002)
```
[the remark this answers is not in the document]
rymanluk: ok

nimble/controller/src/ble_ll_scan.c
```diff
@@ -1924,6 +1934,19 @@ ble_ll_scan_get_addr_from_ext_adv(uint8_t *rxbuf, struct ble_mbuf_hdr *ble_hdr,
         *inita_type =
                 ble_ll_get_addr_type(rxbuf[0] & BLE_ADV_PDU_HDR_RXADD_MASK);
         i += BLE_LL_EXT_ADV_TARGETA_SIZE;
+        if (aux_data) {
+            /* Lets copy addr to aux_data. Need it for e.g. chaining */
+            /* XXX add sanity checks */
```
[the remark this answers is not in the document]
rymanluk: don't remember already :) so yeah, will remove it

nimble/controller/src/ble_ll_conn.c
rymanluk: This function end up here because previous versions of the fix used it in couple of places and I agree name is bad. Looks I forgot to clean this patch. I'm going to get rid of it

rymanluk:
thanks for all the comments @andrzej-kaczmarek . PR is updated now.

nimble/controller/src/ble_ll_ctrl.c
rymanluk: that makes me encourage to add this sentence to commit message. Let me do it.

rymanluk: done

rymanluk:
Now it is rebased and some build issues reported by Travis solved

PR #219, opened 2018-10-11
Allow parallel tests

utzig:
The build matrix was updated to support running tests on multiple machines. Brings testing time down to somewhere between 5 min and 10 min usually, depending on amount of travis resources available.

Other small changes to formatting and style where applied, plus clone git repos without history. Some tests that are just "duplicate" on OSX and Linux were disabled on OSX, like newt build <blinkies> and newt build <targets>.

This depends on runtimeco/mynewt-travis-ci#11. Results can be seen here: https://travis-ci.org/apache/mynewt-nimble/builds/440297541

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,54 +28,102 @@ _addons: &addon_conf
 go:
   - "1.10"
 
+git:
+  depth: false
+
 matrix:
   include:
+    # newt build <targets>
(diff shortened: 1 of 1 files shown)


utzig:
This mirrors the changes from `mynewt-core`...

PR #220, opened 2018-10-11
apps/btshell: Add option to restart advertising after disconnect

michal-narajowski:
This patch takes into account both legacy and extended advertising. When using
legacy advertising (BLE_EXT_ADV: 0) all advertising parameters are remembered
to use when restarting. When using extended advertising the parameters
per instance are remembered by the host. For ext adv the app remembers
the handle of the connection that issued Advertising Complete Event.
Later, at a Disconnect Event, the code checks if we should restart advertising
for an instance related to that connection.

### DIFF
--- a/apps/btshell/src/cmd.c
+++ b/apps/btshell/src/cmd.c
@@ -338,6 +338,7 @@ cmd_advertise_start(int argc, char **argv)
     int max_events;
     uint8_t instance;
     int duration;
+    bool restart;
     int rc;
 
     rc = parse_arg_all(argc - 1, argv + 1);
@@ -368,7 +369,13 @@ cmd_advertise_start(int argc, char **argv)
         return rc;
     }
 
-    rc = ble_gap_ext_adv_start(instance, duration, max_events);
+    restart = parse_arg_bool_dflt("restart", 0, &rc);
+    if (rc != 0) {
+        console_printf("invalid 'restart' parameter\n");
+        return rc;
+    }
+
+    rc = btshell_ext_adv_start(instance, duration, max_events, restart);
     if (rc) {
         console_printf("failed to start advertising instance\n");
         return rc;
@@ -399,7 +406,7 @@ cmd_advertise_stop(int argc, char **argv)
         return rc;
     }
 
-    rc = ble_gap_ext_adv_stop(instance);
+    rc = btshell_ext_adv_stop(instance);
     if (rc) {
         console_printf("failed to stop advertising instance\n");
         return rc;
@@ -488,6 +495,7 @@ static const struct shell_param advertise_start_params[] = {
     {"instance", "default: 0"},
     {"duration", "advertising duration in 10ms units, default: 0 (forever)"},
     {"max_events", "max number of advertising events, default: 0 (no limit)"},
+    {"restart", "restart advertising after disconnect, usage: =[0-1], default: 0"},
     {NULL, NULL}
 };
 
@@ -543,6 +551,7 @@ cmd_advertise(int argc, char **argv)
     ble_addr_t peer_addr;
     ble_addr_t *peer_addr_param = &peer_addr;
     uint8_t own_addr_type;
+    bool restart;
     int rc;
 
     rc = parse_arg_all(argc - 1, argv + 1);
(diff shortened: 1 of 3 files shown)


apps/btshell/src/cmd.c
```diff
@@ -368,7 +369,13 @@ cmd_advertise_start(int argc, char **argv)
         return rc;
     }
 
-    rc = ble_gap_ext_adv_start(instance, duration, max_events);
+    restart_adv = parse_arg_bool_dflt("restart_adv", 0, &rc);
```
andrzej-kaczmarek: nitpick: I'd call this just `restart` since it's kind of obvious that you will restart advertising. and it's faster to type ;-)

apps/btshell/src/main.c
```diff
@@ -129,6 +129,21 @@ int btshell_full_disc_prev_chr_val;
 #define BTSHELL_AUTO_DEVICE_NAME    ""
 #endif
 
+#if MYNEWT_VAL(BLE_EXT_ADV)
+struct {
+    bool restart;
+    int conn_handle;
```
andrzej-kaczmarek: make this `uint16_t` and use `BLE_HS_CONN_HANDLE_NONE` instead of `-1` for invalid handle

apps/btshell/src/main.c
```diff
@@ -985,6 +1000,50 @@ btshell_decode_event_type(struct ble_gap_ext_disc_desc *desc, void *arg)
...
+    int i;
+#endif
+
+    if (event->type != BLE_GAP_EVENT_DISCONNECT) {
+        return -1;
+    }
+
+#if MYNEWT_VAL(BLE_EXT_ADV)
+    for (i = 0; i < BLE_ADV_INSTANCES; ++i) {
+    	if (ext_adv_restart[i].conn_handle == -1) {
```
andrzej-kaczmarek: indentation is not quite right here

apps/btshell/src/main.c
```diff
@@ -985,6 +1000,50 @@ btshell_decode_event_type(struct ble_gap_ext_disc_desc *desc, void *arg)
...
+            continue;
+    	}
+
+        if (ext_adv_restart[i].conn_handle !=
+            event->disconnect.conn.conn_handle) {
+            continue;
```
andrzej-kaczmarek: seems like this `if` also works for condition which the one above checks

apps/btshell/src/main.c
```diff
@@ -985,6 +1000,50 @@ btshell_decode_event_type(struct ble_gap_ext_disc_desc *desc, void *arg)
...
+        }
+
+        if (ext_adv_restart[i].restart) {
```
andrzej-kaczmarek: so you could actually write
```
if (ext_adv_restart[i].restart && (ext_adv_restart[i].conn_handle == event->disconnect.conn.conn_handle)) {
    rc = ble_gap_ext_adv_start(i, 0, 0);
    break;
}
```
and avoid this break at the end of loop below

apps/btshell/src/main.c
```diff
@@ -1439,23 +1504,70 @@ btshell_ext_adv_configure(uint8_t instance,
...
+int
+btshell_ext_adv_start(uint8_t instance, int duration,
+                      int max_events, bool restart_adv)
+{
+    int rc;
+
+    /* Advertising restart doesn't make sense
+     * with limited duration or events
+     */
+    if (restart_adv && duration == 0 && max_events == 0) {
```
andrzej-kaczmarek: add `()` to condition

apps/btshell/src/main.c
```diff
@@ -1439,23 +1504,70 @@ btshell_ext_adv_configure(uint8_t instance,
...
 int
 btshell_adv_start(uint8_t own_addr_type, const ble_addr_t *direct_addr,
-                  int32_t duration_ms, const struct ble_gap_adv_params *params)
+                  int32_t duration_ms, const struct ble_gap_adv_params *params,
+                  bool restart_adv)
 {
     int rc;
 
+    if (restart_adv) {
+    	adv_params.restart_adv = restart_adv;
```
andrzej-kaczmarek: indentation

apps/btshell/src/main.c
```diff
@@ -1074,6 +1134,11 @@ btshell_gap_event(struct ble_gap_event *event, void *arg)
         console_printf("advertise complete; reason=%d, instance=%u, handle=%d\n",
                        event->adv_complete.reason, event->adv_complete.instance,
                        event->adv_complete.conn_handle);
+
+        if (ext_adv_restart[event->adv_complete.instance].restart) {
```
andrzej-kaczmarek: I think you should either skip `if` here (so we always have proper `conn_handle` set in array for debugging) or assign invalid handle in `else`, just to have some consistent state here

michal-narajowski:
@andrzej-kaczmarek Thanks for the comments. Fixed.

PR #221, opened 2018-10-15
mesh: Fix memory leaks

forgithub123:
Function return without os_mbuf_free_chain(buf) after buf = NET_BUF_SIMPLE(n)
will cause memory leaks

### DIFF
--- a/nimble/host/mesh/src/cfg_srv.c
+++ b/nimble/host/mesh/src/cfg_srv.c
@@ -725,7 +725,7 @@ static void beacon_set(struct bt_mesh_model *model,
 		}
 	} else {
 		BT_WARN("Invalid Config Beacon value 0x%02x", buf->om_data[0]);
-		return;
+		goto done;
 	}
 
 	bt_mesh_model_msg_init(msg, OP_BEACON_STATUS);
@@ -735,6 +735,7 @@ static void beacon_set(struct bt_mesh_model *model,
 		BT_ERR("Unable to send Config Beacon Status response");
 	}
 
+done:
 	os_mbuf_free_chain(msg);
 
 }
@@ -1698,7 +1699,7 @@ static void mod_sub_get(struct bt_mesh_model *model,
 	addr = net_buf_simple_pull_le16(buf);
 	if (!BT_MESH_ADDR_IS_UNICAST(addr)) {
 		BT_WARN("Prohibited element address");
-		return;
+		goto done;
 	}
 
 	id = net_buf_simple_pull_le16(buf);
@@ -1739,6 +1740,7 @@ send_list:
 		BT_ERR("Unable to send Model Subscription List");
 	}
 
+done:
 	os_mbuf_free_chain(msg);
 
 }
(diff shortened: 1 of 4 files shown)


nimble/host/mesh/src/cfg_srv.c
```diff
@@ -735,6 +735,7 @@ static void beacon_set(struct bt_mesh_model *model,
 		BT_ERR("Unable to send Config Beacon Status response");
 	}
 
+fail:
```
michal-narajowski: Let's rename it to "done" as in other places. We should free mbuf irrespective of the result.

nimble/host/mesh/src/friend.c
```diff
@@ -453,6 +453,8 @@ static void enqueue_sub_cfm(struct bt_mesh_friend *frnd, u8_t xact)
 
 	frnd->last = buf;
 	frnd->send_last = 1;
+
+fail:
```
michal-narajowski: Again: "done"

nimble/host/mesh/src/friend.c
```diff
@@ -760,6 +762,8 @@ static void enqueue_offer(struct bt_mesh_friend *frnd, s8_t rssi)
...
```
michal-narajowski: Again: "done"

nimble/host/mesh/src/transport.c
```diff
@@ -590,8 +590,9 @@ static int sdu_recv(struct bt_mesh_net_rx *rx, u32_t seq, u8_t hdr,
 
 	if (IS_ENABLED(CONFIG_BT_MESH_FRIEND) && !rx->local_match) {
 		BT_DBG("Ignoring PDU for LPN 0x%04x of this Friend",
-		       rx->ctx.recv_dst);
-		return 0;
+			   rx->ctx.recv_dst);
```
michal-narajowski: The indentation was fine here.

forgithub123:
Renamed 'fail' as 'done', and the indentation is my mistake.

michal-narajowski:
@forgithub123 Please squash these two commits into one. You can use `git rebase -i master` and then follow help.

nimble/host/mesh/src/transport.c
```diff
@@ -590,8 +590,9 @@ static int sdu_recv(struct bt_mesh_net_rx *rx, u32_t seq, u8_t hdr,
 
 	if (IS_ENABLED(CONFIG_BT_MESH_FRIEND) && !rx->local_match) {
 		BT_DBG("Ignoring PDU for LPN 0x%04x of this Friend",
-		       rx->ctx.recv_dst);
-		return 0;
+			rx->ctx.recv_dst);
```
michal-narajowski: This part still needs whitespace fix:

```suggestion
		       rx->ctx.recv_dst);
```

michal-narajowski: In other words, mesh code was ported from Zephyr and the coding style should be the same. So you should use tabs and then align to the beginning of parenthesis with spaces.

PR #226, opened 2018-10-18
Fix c&p issues in .yml files

andrzej-kaczmarek:


andrzej-kaczmarek:
Same as https://github.com/apache/mynewt-core/pull/1467 - I'll just remove these comments everywhere.

PR #228, opened 2018-10-23
Connection request rx while already connected

cedric84:
Hi,
In the situation I have to face, my slave allows several masters to create a connection with it.
The slave uses extended advertising and sends it continuously until a connection request is received.
A first connection is created and the slave transmits a new extended advertising. The master is reset (to simulate an issue) and restarts before the slave detects the disconnection. The master sends a new connection request as it receives the extended advertising and the slave stops the advertising state to enter the connection state. From the slave point of view, the connection already exists so the connection creation (done in the ble_ll_conn_slave_start function) fails but the advertising does not restart.
The update I propose is to ignore the connection request if the connection is already created before trying to create the connection. This way, the advertising state continues as if the frame was reject by the white list.
According to me, this update conforms with the Bluetooth Core Specification Version 5.0, Volume 6, Part B, Section 4.5, page 2637 "If an advertiser receives a connection request from an initiator it is already connected to, it shall ignore that request".
What do you think ?

Note : The code I insert in my "ble_ll_adv_already_connected" function is very similar from the one in the start of the "ble_ll_conn_slave_start" function, so it would perhaps be better to create this new function in the "ble_ll_conn.c" and to call it from the "ble_ll_conn_slave_start" and "ble_ll_adv_rx_req" functions.

Best regards,
Cedric

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -2674,11 +2674,6 @@ ble_ll_adv_rx_req(uint8_t pdu_type, struct os_mbuf *rxpdu)
     }
 #endif
 
-    /* See if the device is already connected */
-    if (ble_ll_adv_already_connected(peer, peer_addr_type)) {
-        return -1;
-    }
-
     /* Set device match bit if we are whitelisting */
     if (chk_wl && !ble_ll_whitelist_match(peer, peer_addr_type, resolved)) {
         return -1;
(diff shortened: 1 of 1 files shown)


nimble/controller/src/ble_ll_adv.c
```diff
@@ -2646,6 +2674,11 @@ ble_ll_adv_rx_req(uint8_t pdu_type, struct os_mbuf *rxpdu)
     }
 #endif
 
+    /* See if the device is already connected */
+    if (ble_ll_adv_already_connected(peer, peer_addr_type)) {
```
andrzej-kaczmarek: as per spec you quoted, this should be only checked when we receive CONNECT_IND or AUX_CONNECT_REQ - we should still be able to scan peer we are connected to (unless spec does not allow for this)

cedric84:
I think you are right.
The test is now performed only in case of an initiating PDU reception so the frame won't be ignored in case of a scan request anymore.

PR #232, opened 2018-10-26
Use syscfg settings for sysinit stage numbers

ccollins476ad:
This allows the app or target to rearrange package initialization order via syscfg overrides.

**Note: This requires an updated version of newt
(https://github.com/apache/mynewt-newt/pull/230).  Older versions of newt fail to parse the updated `pkg.yml` files.

I wanted to make this change backwards compatible via an injected newt syscfg setting, e.g.,
```
pkg.init.NEWT_FEATURE_SYSCFG_STAGES:
    log_init: 'MYNEWT_VAL(LOG_SYSINIT_STAGE)'

pkg.init.!NEWT_FEATURE_SYSCFG_STAGES:
    log_init: 100
```

Unfortunately, this is not possible, as the old newt does not allow conditionals to be applied to `pkg.init`.  So, as soon as this PR is merged, everything will break for users using older versions of newt.

I think it is still worth it to get this change in.  Before we merge, we can send a note to the dev list and the slack channel.  When 1.6.0 is released, we can make sure to add the appropriate newt version restriction.

### DIFF
--- a/nimble/host/mesh/pkg.yml
+++ b/nimble/host/mesh/pkg.yml
@@ -45,5 +45,5 @@ pkg.req_apis:
     - stats
 
 pkg.init:
-    bt_mesh_register_gatt: 500
-    ble_mesh_shell_init: 1000
+    bt_mesh_register_gatt: 'MYNEWT_VAL(BLE_MESH_SYSINIT_STAGE)'
+    ble_mesh_shell_init: 'MYNEWT_VAL(BLE_MESH_SYSINIT_STAGE_2)'
--- a/nimble/host/mesh/pkg.yml
+++ b/nimble/host/mesh/pkg.yml
@@ -46,4 +46,4 @@ pkg.req_apis:
 
 pkg.init:
     bt_mesh_register_gatt: 'MYNEWT_VAL(BLE_MESH_SYSINIT_STAGE)'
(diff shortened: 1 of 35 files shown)


nimble/host/mesh/pkg.yml
```diff
@@ -45,5 +45,5 @@ pkg.req_apis:
     - stats
 
 pkg.init:
-    bt_mesh_register_gatt: 500
-    ble_mesh_shell_init: 1000
+    bt_mesh_register_gatt: 'MYNEWT_VAL(BLE_MESH_SYSINIT_STAGE)'
+    ble_mesh_shell_init: 'MYNEWT_VAL(BLE_MESH_SYSINIT_STAGE_2)'
```
rymanluk: Can we use BLE_MESH_SHELL_SYSINIT_STAGE instead of BLE_MESH_SYSINIT_STAGE_2?

ccollins476ad: Thanks, @rymanluk.  I agree that `BLE_MESH_SYSINIT_STAGE_2` is a poor name.

I renamed it to `BLE_MESH_SYSINIT_STAGE_SHELL` rather than your suggestion of `BLE_MESH_SHELL_SYSINIT_STAGE`.  I don't have a preference either way, but I already chose this arbitrary convention for other packages with multiple sysinit stages.

I pushed this commit separately with the intent to fixup before merging.

PR #233, opened 2018-10-28
fix length comparison bug

InternetofAwesome:
This looks like a bug. Fails consistently for me without this fix, and it looks like we are using this for the specified length for outgoing csa requests.


InternetofAwesome:
@andrzej-kaczmarek , is this what you mean? Seems to work on my end, though I have not tested every type of message that can be passed.

PR #236, opened 2018-11-02
nimble/host: Allow auto-start to be disabled

ccollins476ad:
#### Before PR: 
The host always starts during system initialization.

#### After PR:
A Mynewt syscfg setting (`BLE_HS_AUTO_START`) was added. When this setting is disabled, the host does not start automatically. Instead, the application starts the host using the new
`ble_hs_sched_start()` function.

`BLE_HS_AUTO_START` is enabled by default.

### DIFF
--- a/nimble/host/src/ble_hs.c
+++ b/nimble/host/src/ble_hs.c
@@ -455,6 +455,17 @@ ble_hs_timer_resched(void)
      */
     ble_hs_timer_reset(0);
 }
+ 
+void
+ble_hs_sched_start(void)
+{
+#ifdef MYNEWT
+    ble_npl_eventq_put((struct ble_npl_eventq *)os_eventq_dflt_get(),
+                       &ble_hs_ev_start);
+#else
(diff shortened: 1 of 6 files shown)


nimble/host/src/ble_hs.c
```diff
@@ -699,7 +710,10 @@ ble_hs_init(void)
      * the application to configure this package in the meantime.
      */
 #ifdef MYNEWT
-    ble_npl_eventq_put((struct ble_npl_eventq *)os_eventq_dflt_get(), &ble_hs_ev_start);
+#if MYNEWT_VAL(BLE_HS_AUTO_START)
```
andrzej-kaczmarek: I guess this `#if` should wrap complete `#if #else` instead of only mynewt-specific part since you also updated ports to work with this new setting

ccollins476ad: Thanks, @andrzej-kaczmarek.  I have fixed it.

PR #237, opened 2018-11-03
nimble/ll: Request/release hfxo using mynewt API

wes3:
This change is to use a new API committed to mynewt core that
will request the hfxo to be turned on and release it when the
hfxo is no longer desired. One reason to do this is that an
application may desire the hfxo to be on and nimble would
blindly turn it on/off when it wanted to. Note that the API
requires that the request/release be paired; you cannot release
without a request and if you request without a release the hfxo
will be on continuously.

A final note: the nimble link layer assumes that when it requests
the hfxo it needs to settle. If the hfxo was already on the code
would not need to wait for the settling time. The code has not
been updated to handle that case. This is not going to cause
any issues; it will operate as it did before. However, future
improvements could be to check if the hfxo was already on and
settled and thus not have to wait for the settling time.

### DIFF
--- a/nimble/controller/src/ble_ll_xcvr.c
+++ b/nimble/controller/src/ble_ll_xcvr.c
@@ -46,15 +46,19 @@ ble_ll_xcvr_rfclk_state(void)
 void
 ble_ll_xcvr_rfclk_enable(void)
 {
-    g_ble_ll_data.ll_rfclk_state = BLE_RFCLK_STATE_ON;
-    ble_phy_rfclk_enable();
+    if (g_ble_ll_data.ll_rfclk_state == BLE_RFCLK_STATE_OFF) {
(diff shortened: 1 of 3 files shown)


andrzej-kaczmarek:
changes look good, but new `nrf5x` APIs should be only used when build for Mynewt since ports do not support this and they should still use direct register access for now (and I guess no need to restore code for initial hfxo settle - they will be fine without it).

wes3:
@andrzej-kaczmarek I was wondering about adding #ifdef MYNEWT to the code when doing the PR but was not sure. OK; I see the code that should be changed for this and will do it. Thanks for the review.

andrzej-kaczmarek:
lgtm

btw, not sure why this does not build on travis, perhaps there was some issue on commit this is based on since error does not seem related

PR #243, opened 2018-11-15
[linux] Stability fixes to linux port: npl mutex and task.

turon:
Fixes to allow the npl unit tests to pass reliably on linux.  Also lower the set of 4.1 feature checks when running on linux so as not to hang when using btvirt virtual controller or older 4.0 controllers.


turon:
@ccollins476ad @andrzej-kaczmarek 
These fixes greatly increase that stability of the Linux port.  
Please review when you have time.

PR #244, opened 2018-11-16
nimble/ll: Qualification and performance issues

rymanluk:
Yet another set of fixes for the qualification along with fixes for stress testing on extended scanner.
One of the major change here is a ref counter for aux_data and handling truncated reports.

I stress tested it quite a lot with big chaining in scannable and non-scannable events. Also tested extended connect and all seems to work fine.


nimble/controller/src/ble_ll_scan.c
```diff
@@ -2447,8 +2445,13 @@ ble_ll_hci_send_ext_adv_report(uint8_t ptype, uint8_t *adva, uint8_t adva_type,
...
     BLE_LL_ASSERT(offset <= datalen);
 
-    /* Store any event left for later use */
-    aux_data->evt = evt;
+    if (aux_data){
+        /* Store any event left for later use */
+        aux_data->evt = evt;
+    } else {
+        /* If it is empty beacon, evt shall be NULL */
+        BLE_LL_ASSERT(!evt);
```
andrzej-kaczmarek: `BLE_LL_ASSERT(evt)` ?

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2587,12 +2590,13 @@ ble_ll_scan_rx_pkt_in(uint8_t ptype, struct os_mbuf *om, struct ble_mbuf_hdr *hd
...
 
 
-        if (BLE_LL_CHECK_AUX_FLAG(aux_data, BLE_LL_AUX_IGNORE_BIT)) {
+        if (aux_data && BLE_LL_CHECK_AUX_FLAG(aux_data, BLE_LL_AUX_IGNORE_BIT)) {
             goto scan_continue;
         }
 
         /* Let's see if that packet contains aux ptr*/
         if (BLE_MBUF_HDR_WAIT_AUX(hdr)) {
+           BLE_LL_ASSERT(aux_data);
```
andrzej-kaczmarek: indentation

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2447,8 +2445,13 @@ ble_ll_hci_send_ext_adv_report(uint8_t ptype, uint8_t *adva, uint8_t adva_type,
 
     BLE_LL_ASSERT(offset <= datalen);
 
-    /* Store any event left for later use */
-    aux_data->evt = evt;
+    if (aux_data){
```
andrzej-kaczmarek: missing space

nimble/controller/src/ble_ll_scan.c
```diff
@@ -1113,6 +1114,60 @@ ble_ll_scan_window_chk(struct ble_ll_scan_sm *scansm, uint32_t cputime)
...
+    os_sr_t sr;
+
+    OS_ENTER_CRITICAL(sr);
+    if (!aux_data) {
+        OS_EXIT_CRITICAL(sr);
+        return;
+    }
+
+    aux_data->ref_cnt++;
+    STATS_INC(ble_ll_stats, aux_ref);
```
andrzej-kaczmarek: tbh I do not see how such stat could be useful

nimble/controller/src/ble_ll_scan.c
```diff
@@ -1113,6 +1114,60 @@ ble_ll_scan_window_chk(struct ble_ll_scan_sm *scansm, uint32_t cputime)
...
+}
+
+int
+ble_ll_scan_aux_data_unref(struct ble_ll_aux_data *aux_data)
+{
+    os_sr_t sr;
+    int ret;
+
+    OS_ENTER_CRITICAL(sr);
+    if (!aux_data) {
```
andrzej-kaczmarek: probably it would be better to just assert when we try to ref/unref aux_data which is NULL - this is clearly an error

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2238,6 +2299,12 @@ ble_ll_scan_rx_isr_end(struct os_mbuf *rxpdu, uint8_t crcok)
     if (rc) {
         ble_ll_state_set(BLE_LL_STATE_STANDBY);
     }
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_EXT_ADV)
+    /* On handover lets increase ref count */
+    if (ble_hdr->rxinfo.user_data) {
+        ble_ll_scan_aux_data_ref(ble_hdr->rxinfo.user_data);
```
andrzej-kaczmarek: I think you should ref/unref at the same place where copying takes place, otherwise it will be just hard to follow and prone to errors

nimble/controller/src/ble_ll_scan.c
```diff
@@ -408,6 +408,104 @@ ble_ll_scan_find_dup_adv(uint8_t *addr, uint8_t txadd)
...
+ble_ll_scan_end_adv_evt(struct ble_ll_aux_data *aux_data)
+{
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_EXT_ADV)
+    /*
+     * Normally reference counter here should be 2. 1 for outstanding
+     * complete event and one for ongoing scanning. If there is only 1
+     * that means, advertising event is already completed
+     * (truncated was sent to the host) and we just need to drop last reference.
+     * Otherwise we should try to send truncated event to the host.
+     */
```
andrzej-kaczmarek: I don't think using reference counter for taking some action here is a good solution here - if something changes in future these assumptions can be simply wrong. I think this should be solved with e.g. flags or sth.

nimble/controller/src/ble_ll_adv.c
```diff
@@ -2657,17 +2660,28 @@ ble_ll_adv_rx_req(uint8_t pdu_type, struct os_mbuf *rxpdu)
...
+                    resolved = 1;
+                }
+            } else {
+                if (chk_wl) {
+                    return -1;
+                }
             }
         } else {
-            if (chk_wl) {
+            /* Make sure privacy mode */
```
andrzej-kaczmarek: seems like something is missing here

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2384,15 +2385,15 @@ ble_ll_hci_send_ext_adv_report(uint8_t ptype, uint8_t *adva, uint8_t adva_type,
...
     do {
         need_event = false;
         next_evt = NULL;
 
-        evt->adv_data_len = min(BLE_LL_MAX_EVT_LEN - sizeof(*evt),
+        evt->adv_data_len = min(max_event_size - sizeof(*evt),
                                 datalen - offset);
-        evt->event_len = (sizeof(*evt) - BLE_HCI_EVENT_HDR_LEN) +
-                         evt->adv_data_len;
+        evt->event_len = sizeof(*evt) + evt->adv_data_len;
```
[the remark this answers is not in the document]
rymanluk: good catch with `- BLE_HCI_EVENT_HDR_LEN`. Also as explained offline, there is no change needed in the line above.

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2384,15 +2385,15 @@ ble_ll_hci_send_ext_adv_report(uint8_t ptype, uint8_t *adva, uint8_t adva_type,
     }
 
     offset = 0;
+    max_event_size = min(UINT8_MAX, BLE_LL_MAX_EVT_LEN);
```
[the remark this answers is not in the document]
rymanluk: ok

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2429,15 +2427,15 @@ ble_ll_hci_send_ext_adv_report(uint8_t ptype, uint8_t *adva, uint8_t adva_type,
                 evt->evt_type |= (BLE_HCI_ADV_DATA_STATUS_TRUNCATED);
                 rc = -1;
             }
-        } else if (aux_data) { 
-            if (BLE_LL_CHECK_AUX_FLAG(aux_data, BLE_LL_AUX_INCOMPLETE_ERR_BIT)) {
+        } else {
```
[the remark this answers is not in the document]
rymanluk: ok, can do that, however there will be one more `else` :)

nimble/controller/src/ble_ll_scan.c
rymanluk: evt shall be NULL so  this check is good.

nimble/controller/src/ble_ll_scan.c
rymanluk: I agree this has limited usage but can be useful for debugging. When it is not changing when scanning ext advertisement, you know that you are out of aux_pool. Also when you set aux_pool to 1, it can help with debugging. However, I think that better would be to use system view for this, I will remove this stats and add system view calls instead.

nimble/controller/src/ble_ll_scan.c
rymanluk: well is some part of code this is called  without checking aux_data and this is ok that  aux_data is null then. But maybe I will change that so this function is called only for valid aux_data ... I will see

nimble/controller/src/ble_ll_scan.c
rymanluk: I would keep it here as this is single exit place from this function and we know that aux_data will be handover to LL.
In opposite there is no single place in this functions where aux_data is created/copied. If you look from the top of this function you will see that there is at least 2 paths to get to the end of this function with valid aux_data, so we have to do decide to ref in those 2 places  - I think that would complicate code. Single ref in the end of this function looks much better for me.

nimble/controller/src/ble_ll_scan.c
rymanluk: hmm, In general I agree - will try to figure out something.

nimble/controller/src/ble_ll_adv.c
rymanluk: will change to -> "Verify privacy mode"

nimble/controller/src/ble_ll_scan.c
andrzej-kaczmarek: so I think other parts of code should check for null - if you want to ref something, I'd expect it's not null.

nimble/controller/src/ble_ll_scan.c
andrzej-kaczmarek: imo the point of using refs is that you ref immediately when making copy, this way you always know how many refs you have and you can't make mistake here. copying reference in one place and increasing ref count in other place makes this hard to follow, especially that here it's based on some assumption about contents of `ble_hdr->rxinfo.user_data` which for me is counter-intuitive, but I guess it's ok

rymanluk:
@andrzej-kaczmarek fixed.
I put it as fixup - before merge will squash them.

nimble/controller/src/ble_ll_scan.c
rymanluk: fixed

rymanluk:
thanks @andrzej-kaczmarek  - squashed and waiting for Travis before merging.

rymanluk:
It is RIOT build which fails as it tries to apply patch which is already in the master 
`Applying: L2CAP sm: do not discard alignment information.
error: patch failed: nimble/host/src/ble_sm.c:261
`

Addressed to RIOT team (@haukepetersen )

Merging.

PR #248, opened 2018-11-28
nimble host stack should not consider connection handle 0 as invalid.

sagb2015:
As per Bluetooth specification, connection handle can have value in the
range 0x000-0xEFF. So 0 can actually be a valid handle for a connection.
This change fixes the mesh code to not consider 0 as invalid handle.

### DIFF
--- a/nimble/host/mesh/src/prov.c
+++ b/nimble/host/mesh/src/prov.c
@@ -237,7 +237,9 @@ static void reset_link(void)
 
 	/* Clear everything except the retransmit delayed work config */
 	memset(&link, 0, offsetof(struct prov_link, tx.retransmit));
-
+#if (MYNEWT_VAL(BLE_MESH_PB_GATT))
+	link.conn_handle = BLE_HS_CONN_HANDLE_NONE;
+#endif
 	link.rx.prev_id = XACT_NVAL;
 
 	if (bt_pub_key_get()) {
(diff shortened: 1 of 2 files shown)


nimble/host/mesh/src/prov.c
```diff
@@ -237,7 +237,7 @@ static void reset_link(void)
 
 	/* Clear everything except the retransmit delayed work config */
 	memset(&link, 0, offsetof(struct prov_link, tx.retransmit));
-
+    link.conn_handle = BLE_HS_CONN_HANDLE_NONE;
```
rymanluk: I think it should be guarded with #if (MYNEWT_VAL(BLE_MESH_PB_GATT))

also please check indentation here

nimble/host/mesh/src/proxy.c
```diff
@@ -877,7 +878,7 @@ bool bt_mesh_proxy_relay(struct os_mbuf *buf, u16_t dst)
 		struct bt_mesh_proxy_client *client = &clients[i];
 		struct os_mbuf *msg;
 
-		if (!client->conn_handle) {
+		if (clients[i].conn_handle == BLE_HS_CONN_HANDLE_NONE) {
```
rymanluk: since there is `*client`, let us use it:
 if (client->conn_handle == BLE_HS_CONN_HANDLE_NONE) {}

sagb2015:
@rymanluk -  Incorporated the suggested changes. Please review.

yang325:
This change is great. When will it be merged into master?

rymanluk:
Thanks, all looks great. Merged

PR #250, opened 2018-11-29
blehr: Fix state notifications

prasad-alatkar:
Use conn_handle in ble_gattc_notify_custom() instead o incorrect notify_state

### DIFF
--- a/apps/blehr/src/main.c
+++ b/apps/blehr/src/main.c
@@ -32,6 +32,9 @@
 
 static bool notify_state;
 
+/* Connection handle */
+static uint16_t conn_handle;
+
 static const char *device_name = "blehr_sensor";
 
 static int blehr_gap_event(struct ble_gap_event *event, void *arg);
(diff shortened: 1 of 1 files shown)


apps/blehr/src/main.c
```diff
@@ -162,11 +165,17 @@ blehr_gap_event(struct ble_gap_event *event, void *arg)
...
+        }
+        else {
+          conn_handle = event->connect.conn_handle;
         }
+
         break;
 
     case BLE_GAP_EVENT_DISCONNECT:
         MODLOG_DFLT(INFO, "disconnect; reason=%d\n", event->disconnect.reason);
+        conn_handle = 0; // reset conn_handle
```
andrzej-kaczmarek: 0 is a valid connection handle, we have symbol defined iirc in GAP for invalid conn handle
also do not use C++ style comments

prasad-alatkar: Could you point me to the symbol definition `iirc` ? I could not find it gap services.

rymanluk:
@prasad-alatkar I meant commit message and not PR comment. The thing is the it is commit message which stays in the git tree, and PR comment disappears, that is why we care about commit message.

prasad-alatkar:
Hi @rymanluk , got it :+1:

PR #253, opened 2018-11-30
Fix the issue of removing multiple procs having same opcode

dhrishi:
Fixed ble_gattc_extract_by_conn_op() to accept max_procs and extract the appropriate number of procs, as required from each caller:
  i. First matching proc in ble_gattc_extract_first_by_conn_op()
  ii. All matching procs in ble_gattc_fail_procs()


dhrishi:
Thanks @rymanluk. I have made the requested changes.

PR #256, opened 2018-12-03
porting: Check FreeRTOS timer status before stopping it

yang325:


andrzej-kaczmarek:
I am not sure if this is really needed - it means that we try to stop some timer (callout) which was not initialized before so this possibly just hides some other issue.

did you encounter any problems without this fix?

yang325:
In the function ble_hs_start, the timer ble_hs_timer is stopped before initializing it. In my case, the FreeRTOS will assert for the timer is NULL.

sterlinghughes:
can we merge?

yang325:
I highly recommend it.

andrzej-kaczmarek:
I'd really prefer that we fix mentioned NimBLE code which clearly does stop before init which looks just weird in code. I'm not sure yet how to do this since I do not know why this was needed there (apparently something with unittests).

yang325:
In my opinion, ble_npl_callout_stop is trying to stop a item no matter whether it is running. If it is not running, do nothing.

andrzej-kaczmarek:
agree but it should be initialized before it's used - in this case it clearly is stopped before it's initialized which does not seem right

yang325:
So what shall we do next? Do you have a internal discussion?

dhrishi:
We are also seeing the same issue

andrzej-kaczmarek:
imo what is needed is to fix stop-before-init code instead of working this around in NPL.
feel free to provide fix, I may not have spare time in the next few days to look into this.

dhrishi:
@andrzej-kaczmarek @sterlinghughes @yang325 It looks like this was added with commit 65d827c5938ccf644e9845bbd56063dc1a630142, to fix the unit tests?

If you see the code, in `ble_hs_start` it stops/resets before doing init. See here: https://github.com/apache/mynewt-nimble/blob/a1bfe3a6ae4adabbac9c9da7de8c7966569ae6dd/nimble/host/src/ble_hs.c#L647

Let me know, if it is okay to remove this. It should solve the problem in this case.

yang325:
@sterlinghughes @andrzej-kaczmarek What else shall I do before this PR is accepted ?

andrzej-kaczmarek:
@yang325 @dhrishi I think I explained this already, but let me clarify once more: in my opinion there is no issue in FreeRTOS NPL which assumes that callout which we try to stop was previously initialized (otherwise we need to modify every single NPL API to make checks if objects are valid to be consistent here). This is apparently not the case due to change made in mentioned commit so proper solution is to rework that fix so it will not try to stop callout before it was initialized. Simply removing that code is not a proper solution since it will most likely cause new regression.

I do not know what is the proper fix in the code here since I did not have time to analyze this. Feel free to do this and provide fix which resolves mentioned issue and will not cause regression. Also, in case you disagree with my analysis please simply provide a reasonable arguments and I'll reconsider the workaround for sure.

yang325:
Got it. There should be a better solution. Maybe I should have posed an issue.

dhrishi:
@andrzej-kaczmarek Thanks for your response. 

>  proper solution is to rework that fix so it will not try to stop callout before it was initialized.

I agree to this partially. I mean, this will help us fix this issue, no doubt. But this will be specific to this scenario only. If tomorrow someone calls these APIs in sequence from somewhere else then that will have to be fixed again.

May be that is the same reason why implementation of `ble_npl_callout_stop` in NPL for Linux has a check to see if it was inited before stopping it. So, I think the change in this pull request should be good. 
Although, the check can be made through an API (just like the Linux implementation does), which can be used from other places going ahead, if required.

Let me know if you feel this is not the right approach or if I am missing something.

andrzej-kaczmarek:
@dhrishi calling any other API on an unitialized callout should be simply considered invalid so of course someone can do this in future, but it will need to be fixed.

I believe this PR should fix the issue without breaking unit tests: https://github.com/apache/mynewt-nimble/pull/279

aditihilbert:
Hi @yang325 - so is it all working for you now? Can this PR be closed? Thanks!

PR #259, opened 2018-12-05
nimble/ll: Fix handling LL_PHY_UPDATE_IND as a master

rymanluk:
We should response with LL_UNKNOWN_RSP

This should fix LL/PAC/MAS/BV-01

Note: Patch looks big but in fact it is not. Most of it is indentation change since  ` if (connsm->conn_role == BLE_LL_CONN_ROLE_SLAVE) {`  has been removed as it is not needed anymore.


rymanluk:
Travis is complaining of RTOS build (known issues)

Issue #213, opened 2018-10-08
[Ext adv] Send scan req to peer device only once per scanning session unless ADI has changed.


rymanluk:

rymanluk:
https://github.com/apache/mynewt-nimble/pull/262

rymanluk:
Fixed

PR #262, opened 2018-12-07
nimble/ll: Do not send SCAN REQ to already scanned ext advertising

rymanluk:
Solving https://github.com/apache/mynewt-nimble/issues/213

### DIFF
--- a/nimble/controller/src/ble_ll_scan.c
+++ b/nimble/controller/src/ble_ll_scan.c
@@ -126,6 +126,7 @@ struct ble_ll_ext_adv_report {
 struct ble_ll_scan_advertisers
 {
     uint16_t            sc_adv_flags;
+    uint16_t            adi;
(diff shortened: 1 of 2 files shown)


nimble/controller/src/ble_ll_scan.c
```diff
@@ -2103,6 +2121,20 @@ ble_ll_scan_get_addr_from_ext_adv(uint8_t *rxbuf, struct ble_mbuf_hdr *ble_hdr,
 
     return 0;
 }
+
+static int
+ble_ll_scan_get_adi(struct ble_ll_aux_data *aux_data)
```
sjanc: I'd make adi optional output parameter instead of combining it in return code

rymanluk: agreed

PR #272, opened 2018-12-17
nimble/gap: Improve connection parameter update

rymanluk:
This should fix the scenario for Slave when connection parameter update takes longer than GAP_UPDATE_TIMEOUT ms due to connection interval and the instant chosen by the Master


rymanluk:
Travis complains about unit tests, I need to update it before merging. But will do it once PR is confirmed to work.

rymanluk:
When looking into unit tests I found 1 issue in the code which is now fixed. Unit tests are not yet fixed as this is always bit more complicated :)

rymanluk:
Unit tests should be good now

PR #273, opened 2018-12-18
0 is a possible value for conn_handle.

foldl:
use 0xffff as invalid value.


foldl:
More comments:

For GATT bearer, conn_handle can be 0. So it is not proper to use "if (conn_handle)" to judge if connection has been established or not.

rymanluk:
I think it should be fixed now with PR #248. Please verify and it it is a case, please close this PR

foldl:
yes. cool.

PR #284, opened 2019-01-10
nimble/ll: Couple of fixes for extended scanner

rymanluk:
This PR contains:
1) Extended scanner fix for truncated advertising event
2) Fir for address type in the advertising report when OTA addresses are RPA
2) Minor cleaning patches

### DIFF
--- a/nimble/controller/include/controller/ble_ll_scan.h
+++ b/nimble/controller/include/controller/ble_ll_scan.h
@@ -98,6 +98,7 @@ struct ble_ll_scan_params
 #define BLE_LL_AUX_HAS_ADI              0x80
 
 #define BLE_LL_SET_AUX_FLAG(aux_data, flag) ((aux_data)->flags |= flag)
+#define BLE_LL_CLEAR_AUX_FLAG(aux_data, flag) ((aux_data)->flags &= ~flag)
 #define BLE_LL_CHECK_AUX_FLAG(aux_data, flag) (!!((aux_data)->flags & flag))
 
 struct ble_ll_aux_data {
--- a/nimble/controller/include/controller/ble_ll_scan.h
+++ b/nimble/controller/include/controller/ble_ll_scan.h
@@ -97,9 +97,9 @@ struct ble_ll_scan_params
 #define BLE_LL_AUX_TRUNCATED_SENT       0x40
 #define BLE_LL_AUX_HAS_ADI              0x80
 
-#define BLE_LL_SET_AUX_FLAG(aux_data, flag) ((aux_data)->flags |= flag)
-#define BLE_LL_CLEAR_AUX_FLAG(aux_data, flag) ((aux_data)->flags &= ~flag)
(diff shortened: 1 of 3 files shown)


nimble/controller/include/controller/ble_ll_scan.h
```diff
@@ -97,9 +97,9 @@ struct ble_ll_scan_params
 #define BLE_LL_AUX_TRUNCATED_SENT       0x40
 #define BLE_LL_AUX_HAS_ADI              0x80
 
-#define BLE_LL_SET_AUX_FLAG(aux_data, flag) ((aux_data)->flags |= flag)
-#define BLE_LL_CLEAR_AUX_FLAG(aux_data, flag) ((aux_data)->flags &= ~flag)
-#define BLE_LL_CHECK_AUX_FLAG(aux_data, flag) (!!((aux_data)->flags & flag))
+#define BLE_LL_AUX_SET_FLAG(aux_data, flag) ((aux_data)->flags |= flag)
+#define BLE_LL_AUX_CLEAR_FLAG(aux_data, flag) ((aux_data)->flags &= ~flag)
+#define BLE_LL_AUX_CHECK_FLAG(aux_data, flag) (!!((aux_data)->flags & flag))
```
andrzej-kaczmarek: since you modify those macros anyway, it would be safer to use `(flag)` instead of `flag` and there's no need for `!!` as it seems to be used only in `if` statements (?)

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2086,8 +2100,10 @@ ble_ll_scan_parse_ext_hdr(struct os_mbuf *om,
     out_evt->sec_phy = aux_data->aux_phy;
     out_evt->prim_phy = aux_data->aux_primary_phy;
 
-    /* Adjust mbuf to contain advertising data only */
-    os_mbuf_adj(om, i);
+    if (i) {
```
andrzej-kaczmarek: I would check `ext_hdr_len` here so it's easier to follow (there's unconditional `i = ext_hdr_len` few lines earlier)

nimble/controller/include/controller/ble_ll_scan.h
rymanluk: agree

nimble/controller/src/ble_ll_scan.c
rymanluk: agree

rymanluk:
There is some problem with updating status on Travis. When looking into Details it says PASS as below

``` Pull Request #284 nimble/ll: Couple of fixes for extended scanner

This patch makes sure that corret address type is used in the
advertising report when RPA addresses has been used over the air.

This fixes LL/DDI/SCN/BV-33-C

 Commit 23e0a77 
 #284: nimble/ll: Couple of fixes for extended scanner 
 Branch master 
  avatar rymanluk

 #247 passed
```

PR #288, opened 2019-01-11
Host fixes needed for Bluetooth Tester application

michal-narajowski:


nimble/host/src/ble_gap.c
```diff
@@ -1448,8 +1448,17 @@ ble_gap_rx_conn_complete(struct hci_le_conn_complete *evt, uint8_t instance)
...
+    /* If peer RPA is not set in the event and peer address
+     * is RPA then store the peer RPA address so when the peer
+     * address is resolved, the RPA is not forgotten.
+     */
+    if ((memcmp(BLE_ADDR_ANY->val, evt->peer_rpa, 6) == 0) &&
+        (BLE_ADDR_IS_RPA(&conn->bhc_peer_addr))) {
+        conn->bhc_peer_rpa_addr = conn->bhc_peer_addr;
+    } else {
+        conn->bhc_peer_rpa_addr.type = BLE_ADDR_RANDOM;
+        memcpy(conn->bhc_peer_rpa_addr.val, evt->peer_rpa, 6);
```
rymanluk: I think now we can end up with conn->bhc_peer_rpa_addr.val = 00:00:...00 and type RANDOM

Maybe better do:
```
 if ((memcmp(BLE_ADDR_ANY->val, evt->peer_rpa, 6) == 0))  {
        if (BLE_ADDR_IS_RPA(&conn->bhc_peer_addr))) {
            conn->bhc_peer_rpa_addr = conn->bhc_peer_addr;
        }
    } else { ...
```

nimble/host/src/ble_sm.c
```diff
@@ -524,15 +524,14 @@ ble_sm_persist_keys(struct ble_sm_proc *proc)
         memcpy(peer_addr.val, proc->peer_keys.addr, sizeof peer_addr.val);
 
         conn->bhc_peer_addr = peer_addr;
+
         /* Update identity address in conn.
          * If peer's address was an RPA, we store it as RPA since peer's address
```
rymanluk: This commend needs to be updated now.

PR #290, opened 2019-01-13
nrf52: Add support for retrieving u-blox NINA-B1 public address

jnettlet:
u-blox NINA-B1 is an nrf52 based radio module with a public address
stored in UICR registers.  It differs from the existing code in that
it is stored in Big Endian and marked with the upper 16-bits of
CUSTOMER[1] all set to 1.

This patch checks for this and converts and assembles the Public
MAC accordingly without changing the previous behaviour for other
NRF52 modules that mark the upper 16-bits with all zeros.

Tested on the u-blox NINA-B1 device integrated on the SolidRun
i.MX8M based SOM.

Signed-off-by: Jon Nettleton <jon@solid-run.com>

### DIFF
--- a/nimble/drivers/nrf52/src/ble_hw.c
+++ b/nimble/drivers/nrf52/src/ble_hw.c
@@ -67,6 +67,7 @@ int
 ble_hw_get_public_addr(ble_addr_t *addr)
 {
     int rc;
+    int type = 1;
     uint32_t addr_high;
     uint32_t addr_low;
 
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
afaik already existing code to read device public address was just added as an option for Mynewt-based device to store it there but since there's no standard defined for this it's not really reliable (as can be seen here).

instead of extending UICR code in driver we need a way to set controller public address with custom value read in runtime from vendor-specific location. I created https://github.com/apache/mynewt-nimble/pull/291 which addresses this issue by adding new API and removing existing code in driver. for devices with known location of device public address (like nina-b1) I think the best way to use new API is to read public address and set it in `hal_bsp_init` (if BLE controller is present in build).

jnettlet:
We are carrying on the discussion on how to implement this in #291 I will convert it to a BSP integration and then integrate it into whatever API that we decide on in that thread.

PR #291, opened 2019-01-14
nimble/ll: Add API to set controller public address

andrzej-kaczmarek:
ble_ll_set_public_addr() can be used to set controller public address.
This API can be called in e.g. hal_bsp_init() to set public address
loaded from some vendor specific location. Address set in this way will
take precedence over address configured in syscfg or stored in HW.

NOTE: This is *NOT* intended to be used to modify public address on
already running device. Using it for such purpose may yield unknown
results.

### DIFF
--- a/nimble/controller/include/controller/ble_ll.h
+++ b/nimble/controller/include/controller/ble_ll.h
@@ -483,6 +483,15 @@ void ble_ll_event_send(struct ble_npl_event *ev);
 /* Hand received pdu's to LL task  */
 void ble_ll_rx_pdu_in(struct os_mbuf *rxpdu);
 
+/*
+ * Set public address
+ *
+ * This can be used to set controller public address from vendor specific storage,
+ * usually should be done in hal_bsp_init().
+ * Shall be *only* called before LL is initialized, i.e. before sysinit stage.
+ */
+int ble_ll_set_public_addr(const uint8_t *addr);
+
 /* Set random address */
 int ble_ll_set_random_addr(uint8_t *addr, bool hci_adv_ext);
 
--- a/nimble/controller/src/ble_ll.c
+++ b/nimble/controller/src/ble_ll.c
@@ -458,6 +458,14 @@ ble_ll_is_valid_random_addr(uint8_t *addr)
     return rc;
 }
 
+int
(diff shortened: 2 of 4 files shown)


jnettlet:
It seems backwards to me to have the hal_bsp calling into the nimble stack.  If you look at mynewt_core nothing else is calling any ble_ll functions.  I do think the bsp is the correct place for this initialization code to live.  Perhaps we can export a weak hal_bsp_get_public_addr() function that individual bsp packages can override with custom code?  Then let ble_hw_get_public_addr() call into that?

jnettlet:
Looking through the code nothing else in the base bsp calls back into the Nimble stack.  This seems correct since mynewt-core and additional bsp layers may not even run a bluetooth application, which to my understanding is one of the reasons Nimble was broken out into its own project also.  However it does make sense for the bsp to expose access to the underlying hardware since it should know the specifics implementation of the device.

That being said.  how we get from point A to point B I also don't have a strong opinion on.  Should the ble layer pull in the information regarding the hardware, or is it up to the app to use the bsp abstraction to populate the information for the ble stack.  I think the former makes apps simpler to write with less boiler plate, but the latter would allow most flexibility for the app to initialize the hardware in the state it wants.

again, I am only about a week into mynewt and nimble integration.  I do have our bsp setup and working with our hardware and that led to my pull request and this proposed alternative fix.  I do agree that the code to read the public address belongs in the bsp.  Beyond that I will punt to devs that have more experience with the stack.

andrzej-kaczmarek:
We should not make any new calls to Mynewt HAL in LL since this won't be portable. The function name of course can be non-HAL but then there's problem with weak functions which do not work well in Mynewt (and possible other OS-es) since packages are first linked to static libraries and then they are linked into final binary so our weak symbol in LL package won't be overridden by strong symbol from other package. So while I agree that it would be nice to just fetch address automatically from LL, I do not yet see how this could be done in a clean way which works reliably.

With new API things are simple and for generic BSP-s we just need to put this call inside `#ifdef MYNEWT_VAL(BLE_DEVICE)` so it does not get called when LL is not available. For custom BSP-s I think it's just like any other hw available, i.e. if you have sensor in your device then you just call `os_dev_create` in `hal_bsp_init` and now just do the same if your design uses LL.

PR #300, opened 2019-01-17
nimble/ll: Fix scheduling restart

rymanluk:
Remove not needed cast which could cause that time_till_next is negative
when cputime is wrapping around.

This could cause that radio is enabled longer than required.


rymanluk:
it turns out this is not needed

PR #303, opened 2019-01-22
nimble/ll: Fix access address computation

andrzej-kaczmarek:
Core 5.0, Vol 6, Part B, section 2.1.2:
    The Access Address shall be a 32-bit value. Each time it needs a
    new Access Address, the Link Layer shall generate a new random value
    that meets the following requirements:
      (...)
      * It shall have a minimum of two transitions in the most
        significant six bits.

Current implementation of the above only checks if there are no
transitions in 6 msb so it can create invalid access address.

### DIFF
--- a/nimble/controller/src/ble_ll_conn.c
+++ b/nimble/controller/src/ble_ll_conn.c
@@ -511,6 +511,7 @@ ble_ll_conn_calc_access_addr(void)
     uint8_t consecutive;
     uint8_t transitions;
     uint8_t ones;
+    int tmp;
 
     /* Calculate a random access address */
     aa = 0;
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
not sure if this "tricky" way is any faster/slower than just a plain loop, but should work I guess...

andrzej-kaczmarek:
leading zeroes also include zeroes "added" by extending `uint16_t` to `unsigned int` so numbers would not be that obvious

andrzej-kaczmarek:
agree, although with proposed modification we do quite a lot of unnecessary bit counting. instead we can just find all transitions with a simple xor and then just count bits once. I'll send a patch in a moment.

PR #305, opened 2019-01-23
l2cap: Fix channel creation

michal-narajowski:
Previously the channel was added to the channels list after sending
COC Accept event to the application. The application called
ble_l2cap_coc_recv_ready which was trying to find channel on a list,
but failed. With this patch the channel is added on a list before sending
Accept event, and if the application returns an error, the channel is removed
from the list. Moreover if we fail to send Connection Success, the application
should now not receive a confirmation that we are connected.


nimble/host/src/ble_l2cap_sig.c
```diff
@@ -700,7 +701,7 @@ ble_l2cap_sig_coc_req_rx(uint16_t conn_handle, struct ble_l2cap_sig_hdr *hdr,
         /* Make sure we do not send disconnect event when removing channel */
         chan->cb = NULL;
 
-        ble_l2cap_chan_free(chan);
```
rymanluk: missing ble_hs_lock()

PR #341, opened 2019-02-11
Additional Fix For Handling conn_handle=0

RehamTarekAhmed:
A conn_handle of 0 should be handled as a valid connection handle while BLE_HS_CONN_HANDLE_NONE should be the proper assignment for the cases where the connection is no longer valid. This fix includes some spotted areas where a check for a valid connection handle is not properly handled.

### DIFF
--- a/nimble/host/mesh/src/proxy.c
+++ b/nimble/host/mesh/src/proxy.c
@@ -808,8 +808,9 @@ void bt_mesh_proxy_gatt_disconnect(void)
 	for (i = 0; i < ARRAY_SIZE(clients); i++) {
 		struct bt_mesh_proxy_client *client = &clients[i];
 
-		if (client->conn_handle && (client->filter_type == WHITELIST ||
-					    client->filter_type == BLACKLIST)) {
+		if ((client->conn_handle != BLE_HS_CONN_HANDLE_NONE) &&
+		    (client->filter_type == WHITELIST ||
+		     client->filter_type == BLACKLIST)) {
 			client->filter_type = NONE;
 			rc = ble_gap_terminate(client->conn_handle,
 			                       BLE_ERR_REM_USER_CONN_TERM);
--- a/nimble/host/mesh/src/prov.c
+++ b/nimble/host/mesh/src/prov.c
@@ -1013,7 +1013,7 @@ done:
 static inline bool is_pb_gatt(void)
 {
 #if MYNEWT_VAL(BLE_MESH_PB_GATT)
-	return !!link.conn_handle;
+	return (link.conn_handle != BLE_HS_CONN_HANDLE_NONE);
 #else
 	return false;
 #endif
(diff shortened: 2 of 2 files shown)


nimble/host/mesh/src/proxy.c
```diff
@@ -808,8 +808,8 @@ void bt_mesh_proxy_gatt_disconnect(void)
 	for (i = 0; i < ARRAY_SIZE(clients); i++) {
 		struct bt_mesh_proxy_client *client = &clients[i];
 
-		if (client->conn_handle && (client->filter_type == WHITELIST ||
-					    client->filter_type == BLACKLIST)) {
+		if ((client->conn_handle != BLE_HS_CONN_HANDLE_NONE) && (client->filter_type == WHITELIST ||
```
michal-narajowski: The line is too long here, otherwise looks good

RehamTarekAhmed: Fixed. Thanks.

nimble/host/mesh/src/proxy.c
```diff
@@ -808,7 +808,8 @@ void bt_mesh_proxy_gatt_disconnect(void)
 	for (i = 0; i < ARRAY_SIZE(clients); i++) {
 		struct bt_mesh_proxy_client *client = &clients[i];
 
-		if ((client->conn_handle != BLE_HS_CONN_HANDLE_NONE) && (client->filter_type == WHITELIST || 
+		if ((client->conn_handle != BLE_HS_CONN_HANDLE_NONE) &&
```
michal-narajowski: Please align each line to the beginning of the condition in the previous line, like this:

```
		if ((client->conn_handle != BLE_HS_CONN_HANDLE_NONE) &&
		    (client->filter_type == WHITELIST ||
		     client->filter_type == BLACKLIST)) {
			client->filter_type = NONE;
```
And then squash the style fix commits.
Thanks!

nimble/host/mesh/src/proxy.c
```diff
@@ -808,12 +808,13 @@ void bt_mesh_proxy_gatt_disconnect(void)
...
-			client->filter_type = NONE;
-			rc = ble_gap_terminate(client->conn_handle,
+		if ((client->conn_handle != BLE_HS_CONN_HANDLE_NONE) &&
+		    (client->filter_type == WHITELIST ||
+		     client->filter_type == BLACKLIST)) {
+			    client->filter_type = NONE;
+			    rc = ble_gap_terminate(client->conn_handle,
 			                       BLE_ERR_REM_USER_CONN_TERM);
-			assert(rc == 0);
+			    assert(rc == 0);
```
michal-narajowski: You shouldn't change the indentation of the code here. Just the if condition.

RehamTarekAhmed:
Should be resolved by now. Thanks.

michal-narajowski:
Ok, looks good to me. Thank you for the fix!

PR #343, opened 2019-02-15
nimble/mesh: Do not redefine ARRAY_SIZE

andrzej-kaczmarek:
os/util.h has it now, but it may still be needed for ports.


andrzej-kaczmarek:
this should fix build error from https://github.com/apache/mynewt-core/pull/1646

PR #344, opened 2019-02-19
bletest: update bletest HCI commands sending functions

EvalZero:


EvalZero:
why there is a compile error when using the CI tools and which way is right way that we submit the PR
@andrzej-kaczmarek

rymanluk:
@EvalZero do we still need it? If so could you please rebase and resolve conflicts?

andrzej-kaczmarek:
bletest was removed from tree, closing this one

PR #345, opened 2019-02-22
nimble/ll: Couple of fixes

rymanluk:


rymanluk:
Build failure is due to linux_blemesh app which is not related to those changes.
Fix will be delivered here: https://github.com/apache/mynewt-nimble/pull/346

Merging this PR then

PR #348, opened 2019-02-25
nimble/host: Fix ble_sm_tx called without a lock

### LINKED ISSUE #347: SM: When SM is disabled, `ble_sm_tx` is called without a host lock
Configuration:
```
BLE_SM_LEGACY: 0
BLE_SM_SC: 0
```
When pairing is attempted with a device with SM disabled, it leads to a this code path being executed, which calls `ble_sm_tx` without holding the host lock:
https://github.com/apache/mynewt-nimble/blob/master/nimble/host/src/ble_sm.c#L2743

The debug check inside `ble_sm_tx` asserts because the host lock is not held:
https://github.com/apache/mynewt-nimble/blob/master/nimble/host/src/ble_sm_cmd.c#L56

simonratner:
Fixes #347.

### DIFF
--- a/nimble/host/src/ble_sm.c
+++ b/nimble/host/src/ble_sm.c
@@ -2727,6 +2727,7 @@ ble_sm_rx(struct ble_l2cap_chan *chan)
     struct ble_sm_pair_fail *cmd;
     struct os_mbuf *txom;
     uint16_t handle;
+    int rc;
 
     handle = ble_l2cap_get_conn_handle(chan);
     if (!handle) {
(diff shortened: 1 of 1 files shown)


sjanc:
looks good to me, but due to some CI issues you'll have to rebase against latest master

simonratner:
Looks like one test job still failed: https://travis-ci.org/apache/mynewt-nimble/builds/498097122
Let me know what i can do to fix it, or if it is ok to proceed.

andrzej-kaczmarek:
it was just a random travis failure :)
I restarted build manually and it builds fine

PR #349, opened 2019-02-25
porting: Add NPL equivalent for OS_ASSERT_CRITICAL

andrzej-kaczmarek:


andrzej-kaczmarek:
included in https://github.com/apache/mynewt-nimble/pull/346

PR #361, opened 2019-03-04
clang: Fix "unused function" warnings

ccollins476ad:
Clang warns about some unused `static inline` functions.  Conditionally compile out these functions.


sjanc:
This also affects some GCC versions so I allowed myself to merge it :)

PR #365, opened 2019-03-07
controller: fix missing critical sections for ble_ll_xcvr_rfclk_x calls

haukepetersen:
While trying to rebase RIOT to the latest nimBLE version, I found that some critical sections are missing in the controller code, leading to failed assertions in `ble_ll_xcvr.c`, namely `OS_ASSERT_CRITICAL()` in lines 50 and 60.

This PR should fix those?!

### DIFF
--- a/nimble/controller/src/ble_ll_dtm.c
+++ b/nimble/controller/src/ble_ll_dtm.c
@@ -384,7 +384,9 @@ ble_ll_dtm_rx_start(void)
 
 #ifdef BLE_XCVR_RFCLK
     if (ble_ll_xcvr_rfclk_state() == BLE_RFCLK_STATE_OFF) {
+        OS_ENTER_CRITICAL(sr);
         ble_ll_xcvr_rfclk_start_now(os_cputime_get32());
+        OS_EXIT_CRITICAL(sr);
     }
 #endif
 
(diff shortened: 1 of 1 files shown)


nimble/controller/src/ble_ll_xcvr.c
```diff
@@ -102,7 +102,10 @@ void
 ble_ll_xcvr_rfclk_timer_exp(void *arg)
 {
     if (g_ble_ll_data.ll_rfclk_state == BLE_RFCLK_STATE_OFF) {
+        os_sr_t sr;
```
rymanluk: 'os_sr_t sr' shall be on top of the function according to coding guide

This is timer callback function which in nimble is called when interrupts disabled, could you do the same for RIOT?

andrzej-kaczmarek: this is called the same way in RIOT but since `OS_ASSERT_CRITICAL` does not check explicitly that interrupts are disabled (see https://github.com/apache/mynewt-nimble/pull/366) it will fail because in this case they were disabled by `__HAL_DISABLE_INTERRUPTS` (https://github.com/apache/mynewt-nimble/blob/master/porting/nimble/src/hal_timer.c#L28). I think a better solution would be to change those macros in `hal_timer.c` to use `ble_npl_hw_enter_critical` and `ble_npl_hw_exit_critical` so they "obey" hack in NPL OS implementation.

haukepetersen: +1

will propose an alternative to this PR taking the above comments into account. Together with #366 that should hopefully solve this for good.

haukepetersen: proposed a fix in https://github.com/apache/mynewt-nimble/pull/369

haukepetersen:
Removed the changes to `controller/src/ble_ll_xcvr.c` from this PR, as they should be addressed by #369.

The remaining changes should still be valid, though.

haukepetersen:
Any more thoughts on the remaining changes in this PR?

PR #366, opened 2019-03-07
npl/riot: fix ble_npl_hw_is_in_critical()

### LINKED ISSUE #364: npl: ble_npl_hw_is_in_critical() broken for RIOT
Just wanted to bump the Nimble version to use with RIOT and found nimBLE broken for the latest commit https://github.com/apache/mynewt-nimble/commit/70e1d20d214920623623f22180d4f9df19a22349. 

Turns out, that the changes introduced in https://github.com/apache/mynewt-nimble/pull/346 are leading to a failing `OS_ASSERT_CRITICAL()` in `nimble/controller/src/ble_ll_xcvr.c:60`. 

Main cause is that `ble_npl_hw_exit_critical()` is faulty for RIOT. It calls `irq_is_in()`, checking if the code is executed inside of an ISR instead of checking if interrupts are disabled (denoting a critical section in this context).

Now the problem is, that RIOT (up until now) does not offer any `isr_is_enabled()` or similar function, hence I can not simply PR a fix for this. I did try to poke the RIOT community about adding this function (see https://github.com/RIOT-OS/RIOT/pull/11117), but this might take a while to sort out...

So is there anything we could do to circumvent this issue so I can integrate the latest nimBLE version? Options could be to manually track enter/exit of crit sections using a state variable or maybe disabling `OS_ASSERT_CRITICAL()` for the RIOT port for now?

haukepetersen:
fixes #364 (for now)

This is a RIOT specific workaround to make `ble_npl_hw_is_in_critical()` working properly. As there is no `irq_is_enabled()` function in RIOT (yet), this is achieved by keeping the critical section state manually. The run-time overhead should be tolerable, and hopefully we can fix this once we get the missing API call into RIOT.

### DIFF
--- a/porting/npl/riot/include/nimble/nimble_npl_os.h
+++ b/porting/npl/riot/include/nimble/nimble_npl_os.h
@@ -38,6 +38,8 @@ extern "C" {
 typedef uint32_t ble_npl_time_t;
 typedef int32_t ble_npl_stime_t;
 
+extern volatile int ble_npl_in_critical;
+
 struct ble_npl_event {
     event_callback_t e;
     void *arg;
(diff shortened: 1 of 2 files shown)


haukepetersen:
rebased.

PR #367, opened 2019-03-07
controller: remove RIOT specific initialization hack

haukepetersen:
We finally added some additional functionality to RIOT's event queue implementation to allow for thread context independent event queue implementation. With this, there is no need anymore to treat the event queue initialization in the controller differently then for other ports.

Only caveat: the code now depends on the fact, that the first call to `ble_npl_eventq_get()` must be done from the thread's context, that is supposed to host the targeted event queue. But this should be always the case anyway, right?


andrzej-kaczmarek:
yes, every call to get element from this queue will be done from `ll_task` only

PR #380, opened 2019-03-22
[WP] nimble/host: Improve usage of MSYS

rymanluk:
Nimble allocates one mbuf for the packet headers and then do concat it
with data buffer. Till now header buffer was taken from MSYS mempool
which leads to waste big chunks od data, now Nimble maintain its own
mempool for this purpose


nimble/host/syscfg.yml
```diff
@@ -440,6 +440,16 @@ syscfg.defs:
...
+    BLE_HS_OWN_MBUF_FOR_HEADERS:
+        description: >
+             In order to not waste big msys blocks for the ATT + L2CAP and HCI headers
+             with this option on there will be dedicated mempool for this.
+        value : 0
+    BLE_HS_OWN_MBUF_FOR_HEADERS_CNT:
+        description: >
+            Number of blocks for the headers. If this will be used up, MSYS block
+            will be used.
+        value: 'MYNEWT_VAL_MSYS_1_BLOCK_COUNT'
```
andrzej-kaczmarek: you can combine both into single option (>0 means use separate pool)
I'd call it e.g. `BLE_HS_L2CAP_HDR_POOL_COUNT

nimble/host/src/ble_hs_mbuf.c
```diff
@@ -84,6 +98,50 @@ ble_hs_mbuf_l2cap_pkt(void)
...
+    struct os_mbuf *om;
+    int rc;
+
+    om = os_mbuf_get_pkthdr(&ble_hs_hdr_mbuf_pool, leading_space);
+    if (om == NULL) {
+        return NULL;
+    }
+
+    if (om->om_omp->omp_databuf_len < leading_space) {
+        rc = os_mbuf_free_chain(om);
```
wes3: Given that this is a dedicated pool should we presume that we always have the proper space and simply assert() here if the asked for leading space is not enough?

nimble/host/src/ble_hs_mbuf.c
```diff
@@ -19,6 +19,20 @@
...
+struct os_mbuf_pool ble_hs_hdr_mbuf_pool;
+struct os_mempool ble_hs_hdr_pool;
+
+#define BLE_HS_MEMPOOL_BLOCK_SIZE  (BLE_ATT_PREP_WRITE_CMD_BASE_SZ +\
+                                    BLE_HCI_DATA_HDR_SZ + BLE_L2CAP_HDR_SZ)
+static os_membuf_t ble_hs_hdr_mem[
+    OS_MEMPOOL_SIZE(MYNEWT_VAL(BLE_HS_OWN_MBUF_FOR_HEADERS_CNT),
+                    BLE_HS_MEMPOOL_BLOCK_SIZE)
+];
+
```
wes3: Not that I have given this a ton of thought but I wonder if another way to do this is a have a very small mbuf pool added to the msys pools for nimble? This would make for less code but could still waste bib mbufs if no small ones are available. I realize this last comment is probably lame but too bad we cannot force the applications to reserve some space so we can add these headers to the existing mbufs as it would save other overhead.

nimble/host/src/ble_hs_mbuf.c
rymanluk: agree

nimble/host/syscfg.yml
rymanluk: agree

nimble/host/src/ble_hs_mbuf.c
rymanluk: I was thinking to have msys_init_2 with smaller block size but then I found out that there are many places in the system where `os_msys_get_pkthdr(0,0)` is used which means, our smaller mempool whould be used very quickly.

That is why for the moment let's make sure this is only for NimBLE in this way. Does it makes sense?

ccollins476ad:
Sorry if I am misreading, but when we allocate an mbuf, don't we reserve sufficient leading space for the headers?  E.g.,
```
struct os_mbuf *
ble_hs_mbuf_att_pkt(void)
{
    /* Prepare write request and response are the larget ATT commands which
     * contain attribute data.
     */
     return ble_hs_mbuf_gen_pkt(BLE_HCI_DATA_HDR_SZ +
                                BLE_L2CAP_HDR_SZ +
                                BLE_ATT_PREP_WRITE_CMD_BASE_SZ);
}
```

Mbufs allocated with this function should be able to hold headers and data.

rymanluk:
@ccollins476ad thanks for pointing it out. Indeed this function should always allocates from msys as it needs to reserve space for all headers and data.

andrzej-kaczmarek:
@ccollins476ad you're right, but application is not required to use `ble_hs_mbuf_att_pkt` function to allocate mbuf for transfer. in particular, transport agnostic packages like OIC will just allocate buffer from msys and they won't have space reserved for a short header so as I understand this patch will help optimize such transfers so we do not waste msys buffers.

rymanluk:
@andrzej-kaczmarek, @ccollins476ad  actually looking into the code, even `ble_hs_mbuf_att_pkt` is used, the leading space is not used but instead new mbuf is created for the header. I might have fix for that.
But @andrzej-kaczmarek has still valid point  - it is not required to use this functions.

nimble/host/src/ble_hs_mbuf.c
```diff
@@ -84,15 +100,52 @@ ble_hs_mbuf_l2cap_pkt(void)
...
+#if MYNEWT_VAL(BLE_HS_L2CAP_HDR_POOL_COUNT)
+/**
+ * Allocates an mbuf for use by the nimble host.
+ */
+static struct os_mbuf *
+ble_hs_mbuf_gen_hdr_pkt(uint16_t leading_space)
+{
+    struct os_mbuf *om;
+
+    om = os_mbuf_get_pkthdr(&ble_hs_hdr_mbuf_pool, leading_space);
```
ccollins476ad: This allocates an mbuf packet with a user header of size `leading_space` bytes.  This isn't what we want, is it?  The user header is never filled with normal data.  It is just some extra space after the packet header that can be used to "tag" an mbuf.

ccollins476ad:
@andrzej-kaczmarek , @rymanluk - I see what you mean... most code won't call `ble_hs_mbuf_att_pkt()` when allocating mbufs.  But don't we have the same problem with `ble_hs_mbuf_l2cap_hdr_pkt()`?

rymanluk:
@andrzej-kaczmarek @ccollins476ad @wes3  Now code will used msys_x if needed. Related change in core is here: https://github.com/apache/mynewt-core/pull/1739

rymanluk:
fyi - fixing unit tests and there is something wrong with this PR. not ready for review.

rymanluk:
@ccollins476ad  @wes3 @andrzej-kaczmarek  Now it is ready for review.

nimble/host/src/ble_hs_mbuf.c
```diff
@@ -84,15 +85,33 @@ ble_hs_mbuf_l2cap_pkt(void)
...
+ble_hs_mbuf_hdr_pkt(void)
+{
+    struct os_mbuf *om;
+    uint16_t leading_space = BLE_HS_ATT_PKT_HDR_LEADINGSPACE;
+
+
+    /* This is just for header which is going to be added later on, that is
+    * why size is equal to leading_space
+    */
+    om = os_msys_get_pkthdr(leading_space, leading_space);
```
ccollins476ad: I must be missing something, because I still don't understand why we want a user-header length equal to `leading_space`.  Why not specify 0 here, and then perform `om->om_data += leading_space`?

Specifying `leading_space` as the user-header length causes `om->om_pkthdr_len` to be equal to `sizeof (struct os_mbuf_pkthdr) + `leading_space`.  Does the code ever decrease `om_pkthdr_len`?  If not, I think this space can never be used.

rymanluk: you are right, I misread a code. So there is no API to actually add leading space to mbuf contains packet header other then adding it manually.

rymanluk:
added patch which adds btshell command to test `ble_gattc_notify/indicate_custom`

PR #386, opened 2019-03-26
nimble/transport: Make HCI transport select as a choice

andrzej-kaczmarek:


andrzej-kaczmarek:
updated to mark old settings as deprecated

PR #387, opened 2019-03-26
nimble/host: Fix handling no memory for sending ACL data fragmets

rymanluk:
When host has no memory for sending ACL fragment it should not free the
mbuf but instead put packet in the queue.

It is very important espacially in case when there were already some
fragments sent and packet continuation is required


rymanluk:
yeah, mbuf exhaustion problem is other thing. Still thinking about it. Thanks for review.

PR #389, opened 2019-03-26
linux: fix for btvirt controller emulator.

turon:
Resolve lock-up issue when using simulated hci interfaces from bluez (btvirt).

### DIFF
--- a/nimble/host/src/ble_hs_startup.c
+++ b/nimble/host/src/ble_hs_startup.c
@@ -301,6 +301,8 @@ ble_hs_startup_set_evmask_tx(void)
         return rc;
     }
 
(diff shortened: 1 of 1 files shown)


sjanc:
Hi,

This looks like a bug in btvirt which seems to reject set event mask page 2 command is emulated device is not dual mode. Should be fixed there instead of hacking this in nimble host

sjanc:
I've sent a fix for btvirt
https://marc.info/?l=linux-bluetooth&m=155368084018712&w=2

turon:
Thanks Szymon!

sjanc:
btvirt fix is now merged into BlueZ master, I guess this PR can be closed?

PR #390, opened 2019-03-26
nimble/host: Register a 1-elem mbuf pool for ACL; prevents possible deadlock

ccollins476ad:
Register a one-element mbuf pool dedicated to holding outgoing ACL data packets.  This dedicated pool prevents a deadlock caused by mbuf exhaustion.  Without this pool, all msys mbufs could be permanently allocated, preventing us from fragmenting outgoing packets and sending them (and ultimately freeing them).

When the host fragments an outgoing data packet, it first tries to allocate the fragment using the new pool.  If this fails (because the single mbuf has already been sent to the controller), the host falls back to using msys.

### DIFF
--- a/nimble/host/src/ble_hs_hci.c
+++ b/nimble/host/src/ble_hs_hci.c
@@ -440,6 +440,24 @@ ble_hs_hci_frag_alloc(uint16_t frag_size, void *arg)
     return NULL;
 }
 
+/**
+ * Retrieves the total capacity of the ACL fragment pool (always 1).
+ */
+int
+ble_hs_hci_frag_num_mbufs(void)
+{
+    return ble_hs_hci_frag_mempool.mp_num_blocks;
+}
+
+/**
+ * Retrieves the the count of free buffers in the ACL fragment pool.
+ */
+int
+ble_hs_hci_frag_num_mbufs_free(void)
+{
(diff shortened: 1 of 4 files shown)


nimble/host/src/ble_hs_hci.c
```diff
@@ -574,4 +611,15 @@ ble_hs_hci_init(void)
...
+    rc = mem_init_mbuf_pool(ble_hs_hci_frag_data,
+                            &ble_hs_hci_frag_mempool,
+                            &ble_hs_hci_frag_mbuf_pool,
+                            1,
+                            BLE_HS_HCI_FRAG_MEMBLOCK_SIZE,
+                            "ble_hs_hci_frag");
+    BLE_HS_DBG_ASSERT_EVAL(rc == 0);
+
+    rc = os_msys_register(&ble_hs_hci_frag_mbuf_pool);
+    BLE_HS_DBG_ASSERT_EVAL(rc == 0);
```
andrzej-kaczmarek: this dedicated pool should not be registered with msys since anyone will be able to grab our dedicated buffer

ccollins476ad: Oops, you're right.  Thanks for catching that.

ccollins476ad: I have removed those last two lines.

PR #392, opened 2019-03-27
host/svc: add IPSS service for IPSP support

haukepetersen:
This PR adds support for the IPSS (Inernet protocol support service) to NimBLE. This service is defined in the IPSP (Internet Protocol Support Profile) spec. It has not characteristics and is simply there to notify others on a node's capability to run IP traffic (over l2cap coc).

I tested this PR using RIOT using this branch: https://github.com/haukepetersen/RIOT/tree/add_nimble_ipsp and building with `USEMODULE=nimble_svc_ipss`. Only problem is, that one has to manually set the RIOT package to this branch...

This setup does so far only configure this trivial service for a device, it does not yet enable the actual IP-over-BLE capabilities. These i will PR separately to RIOT.

### DIFF
--- a/nimble/host/services/ipss/include/services/ipss/ble_svc_ipss.h
+++ b/nimble/host/services/ipss/include/services/ipss/ble_svc_ipss.h
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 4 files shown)


haukepetersen:
@rymanluk thanks for merging! I do have a test application, it is however tightly coupled to RIOT and its GNRC network stack. The PR will be out somewhere in Mai. I'll let you know.

PR #394, opened 2019-03-28
BLE: configurable stack size

JustineKH:
Changing the BLE stack size from a #define to a syscfg.
We want this to be configurable.


andrzej-kaczmarek:
why this needs to be configurable? setting stack size to anything lower than current non-extadv setting is known to overflow stack in certain scenarios, setting to higher value does not make sense tbh.

PR #399, opened 2019-03-30
host: Always wake tx queues on rx of num-cmp-pkts

ccollins476ad:
Prior to PR: The code assumed that a transmission could stall for only one reason: the controller cannot accommodate any more ACL data packets.  When the host received a num-complered-packets event, it would only try to unstall each connection if all the controller's data buffers were previously used.

It turns out there is a second way that a host connection can stall: mbuf exhaustion.  If the host is unable to fragment an outgoing L2CAP packet into ACL data packets due to a lack of mbufs, the transmission stalls.

This PR causes the host to wake up its connections every time it receives a num-completed-packets event.  It does this even if the controller still had capacity for more ACL data packets.  If a host connection is stalled due to mbuf exhaustion, the controller will only send the next num-completed-packets event after it has freed the mbuf from the designated ACL fragment pool (ble_hs_hci_frag_mbuf_pool). In other words, we are guaranteed to receive a num-completed-packets event whenever a transmission has stalled, regardless of the cause.

### DIFF
--- a/nimble/host/src/ble_hs_hci_evt.c
+++ b/nimble/host/src/ble_hs_hci_evt.c
@@ -226,7 +226,6 @@ ble_hs_hci_evt_num_completed_pkts(uint8_t event_code, uint8_t *data, int len)
     uint16_t num_pkts;
     uint16_t handle;
     uint8_t num_handles;
-    int tx_outstanding;
     int off;
     int i;
 
@@ -241,13 +240,6 @@ ble_hs_hci_evt_num_completed_pkts(uint8_t event_code, uint8_t *data, int len)
         return BLE_HS_ECONTROLLER;
     }
     off++;
-
-    /* If we were previously blocked due to controller buffer exhaustion, and
-     * this event indicates that some buffers have been freed, then the host
-     * needs to resume transmitting.  `tx_outstanding` keeps track of whether
-     * this is the case.
-     */
-    tx_outstanding = 0;
 
     for (i = 0; i < num_handles; i++) {
         handle = get_le16(data + off);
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_hs_hci_evt.c
```diff
@@ -264,19 +256,14 @@ ble_hs_hci_evt_num_completed_pkts(uint8_t event_code, uint8_t *data, int len)
                     conn->bhc_outstanding_pkts -= num_pkts;
                 }
 
-                if (ble_hs_hci_avail_pkts == 0) {
```
rymanluk: If there is no traffic on ACL, Nimble controller will send Number of Completed packets every X sec. I think it is 2 sec by default with num_pkts set to 0. I believe we don't need to call `ble_hs_wakeup_tx()` in such a case.

Maybe  instead of this change we could just remove `if (ble_hs_hci_avail_pkts == 0)` and it should be sufficient.

ccollins476ad: I have not looked at the spec, but that is not how the nimble controller works.  The nimble controller only sends a 0-count num-completed-packets event if it has outgoing packets enqueued that it has not sent yet.  If the controller is idle, it does not send any num-completed-packets events (the code is a bit tricky; `ble_ll_conn_num_comp_pkts_event_send()`).

Like I said - I am not familiar with the controller sections of the spec.  Maybe the controller is allowed to send a num-completed-packets event every two seconds, even when idle.  I don't see that as a huge issue though.  Looping through some empty queues every two seconds in the host task should not present a performance problem.

wes3: Here is the section from the spec:
While the Controller has HCI data packets in its buffer, it must keep sending the Number Of Completed Packets event to the Host at least periodically, until it finally reports that all the pending ACL Data Packets have been transmitted or flushed. The rate with which this event is sent is manufacturer specific.

As Chris says, the controller will not send 0 in its num completed packet event unless there is data enqueued. The idea behind not doing this periodically is that we did not want to allocate an event buffer for it if not needed.

I do not see why the controller would be prohibited from doing this but why should it? It is just extra work that does not need to be done (imo).

rymanluk: @ccollins476ad @wes3   you are right, and change looks good.

PR #407, opened 2019-04-03
nimble/host: Replace chr_def_handle with chr_val_handle

dhrishi:
At some places for APIs ble_gatt_dsc_fn() and ble_gatts_chr_updated(), chr_val_handle was used and at other places chr_def_handle was used. Changed it uniformly with the right one to avoid confusion


dhrishi:
If it looks okay, can this be merged?

PR #411, opened 2019-04-10
nimble/gap: Add missing ble_gap_event_listener_call() call

mnasr93:
calling gap event listener when receiving notification/indication


mnasr93:
you are welcome

PR #419, opened 2019-04-16
nimble/gap: Allow reconfiguration of the same adv instance.

mfarag13:
1 - Allow reconfiguration for extended adv instance as long as it's not enabled.
If it's enabled and and host try to reconfigure it. Controller will send error code command disallowed (0x0c) status to host.
2 - Use correct return types of Host Errors.

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -2510,7 +2510,7 @@ ble_gap_ext_adv_configure(uint8_t instance,
     int rc;
 
     if (instance >= BLE_ADV_INSTANCES) {
-        return EINVAL;
+        return BLE_HS_EINVAL;
     }
 
     rc = ble_gap_ext_adv_params_validate(params);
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_gap.c
```diff
@@ -2520,12 +2524,6 @@ ble_gap_ext_adv_configure(uint8_t instance,
 
     ble_hs_lock();
 
-    /* TODO should we allow to reconfigure existing instance? */
-    if (ble_gap_slave[instance].configured) {
```
sjanc: I don't think we should allow to reconfigure active instance, should this be checked here?

nimble/host/src/ble_gap.c
```diff
@@ -1291,6 +1291,10 @@ ble_gap_rx_adv_set_terminated(struct hci_le_adv_set_terminated *evt)
 
     ble_gap_adv_finished(evt->adv_handle, reason, conn_handle,
                          evt->completed_events);
+
+    /* Remove ble_gap_slave instance */
+    memset(&ble_gap_slave[evt->adv_handle], 0, 
+           sizeof(struct ble_gap_slave_state));
```
sjanc: this is not valid,  set terminate means that advertising was stopped, not that instance was removed (ie you can re-start advertising without configuring instance)

nimble/host/src/ble_gap.c
mfarag13: I think we can allow reconfiguration as long as we didn't enable extended advertising.

nimble/host/src/ble_gap.c
mfarag13: removed this.

PR #425, opened 2019-04-19
nimble/host: Allow NULL in Find attribute by UUID

michal-narajowski:
Specifying UUID param as NULL will return any type of attribute.
This can be useful to get a list of attributes in the database.
For example it can be used in bttester application. It is an
internal API, so generally it should not be used by an application
but it is a testing application.

### DIFF
--- a/nimble/host/src/ble_att_svr.c
+++ b/nimble/host/src/ble_att_svr.c
@@ -171,7 +171,8 @@ ble_att_svr_find_by_handle(uint16_t handle_id)
 /**
  * Find a host attribute by UUID.
  *
- * @param uuid                  The ble_uuid_t to search for
+ * @param uuid                  The ble_uuid_t to search for; null means
+ *                                  find any type of attribute.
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_att_svr.c
```diff
@@ -171,7 +171,8 @@ ble_att_svr_find_by_handle(uint16_t handle_id)
 /**
  * Find a host attribute by UUID.
  *
- * @param uuid                  The ble_uuid_t to search for
+ * @param uuid                  The ble_uuid_t to search for; null means
+ *                                  find any attribute.
```
rymanluk: looks like not any, but fist from attribute list or next after prev

michal-narajowski: I meant any type of attribute. The behavior you describe is not related to the `uuid` param. It is described under the `prev` param.

michal-narajowski:
@rymanluk I updated the comment and the commit message so it is a little more clear.

PR #429, opened 2019-04-29
drivers/nrf51: make it RIOT compatible

aabadie:
This PR is adapting the nrf51 driver part to make it RIOT compatible:
- apply the same preprocessor macro conditionals use with nrf52 driver
- fix an issue with a discarded volatile qualifier


aabadie:
That was fast, thanks :)

PR #431, opened 2019-05-02
nimble/gap: Adding LE Clear Advertising Sets Command

mfarag13:
Adding ble_gap_ext_adv_clear() API that is used to send LE Clear Advertising Sets command.

### DIFF
--- a/nimble/host/include/host/ble_gap.h
+++ b/nimble/host/include/host/ble_gap.h
@@ -965,6 +965,13 @@ int ble_gap_ext_adv_stop(uint8_t instance);
 int ble_gap_ext_adv_set_data(uint8_t instance, struct os_mbuf *data);
 int ble_gap_ext_adv_rsp_set_data(uint8_t instance, struct os_mbuf *data);
 int ble_gap_ext_adv_remove(uint8_t instance);
+/**
+ * Clear All existing extended Adv Sets.
+ * @return              0 on success,
+ *                      BLE_HS_EBUSY if advertising is in progress,
+ *                      other error code on failure.
+ */
+int ble_gap_ext_adv_clear(void);
 
 #if MYNEWT_VAL(BLE_PERIODIC_ADV)
 /* Periodic Advertising */
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -3007,6 +3007,47 @@ ble_gap_ext_adv_remove(uint8_t instance)
     return 0;
 }
 
+int
+ble_gap_ext_adv_clear(void)
+{
+    int rc;
+    uint8_t instance;
+    uint16_t opcode;
+
+    ble_hs_lock();
+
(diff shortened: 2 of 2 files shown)


nimble/host/include/host/ble_gap.h
```diff
@@ -965,6 +965,7 @@ int ble_gap_ext_adv_stop(uint8_t instance);
 int ble_gap_ext_adv_set_data(uint8_t instance, struct os_mbuf *data);
 int ble_gap_ext_adv_rsp_set_data(uint8_t instance, struct os_mbuf *data);
 int ble_gap_ext_adv_remove(uint8_t instance);
+int ble_gap_ext_adv_clear(void);
```
rymanluk: Could you please add doxygen for this new API? I know that there is a bad example above with missing doxygen and for that I have just created an issue on Nimble, so don't worry about that.

nimble/host/src/ble_gap.c
```diff
@@ -3007,6 +3007,44 @@ ble_gap_ext_adv_remove(uint8_t instance)
...
+    uint16_t opcode;
+
+    ble_hs_lock();
+
+    for(instance = 0; instance < BLE_ADV_INSTANCES; instance++) {
+        /* If there is an active instance or periodic adv instance,
+         * Don't send the command
+         * */
+        if ((ble_gap_slave[instance].op == BLE_GAP_OP_S_ADV)
+#if MYNEWT_VAL(BLE_PERIODIC_ADV)
```
rymanluk: I would just add separate `if ( ` for periodic check as it will be cleaner and not against wrapping long lines coding style (https://github.com/apache/mynewt-core/blob/master/CODING_STANDARDS.md)

nimble/host/src/ble_gap.c
```diff
@@ -3007,6 +3007,44 @@ ble_gap_ext_adv_remove(uint8_t instance)
...
+int
+ble_gap_ext_adv_clear(void)
+{
+    int rc;
+    uint8_t instance;
+    uint16_t opcode;
+
+    ble_hs_lock();
+
+    for(instance = 0; instance < BLE_ADV_INSTANCES; instance++) {
```
rymanluk: minor coding style `if ( `

nimble/host/include/host/ble_gap.h
mfarag13: Added.

nimble/host/src/ble_gap.c
mfarag13: Modified.

nimble/host/src/ble_gap.c
mfarag13: Modified.

nimble/host/include/host/ble_gap.h
rymanluk: thanks, we just miss `@return ` with error code explanation.

mfarag13: Added.

PR #435, opened 2019-05-08
ble_hs: `ble_hs_id_set_rnd` should check if all bits are 0s or 1s

prasad-alatkar:
`ble_hs_id_set_rnd` should make sure that all bits are not 0s or 1s.
Useful for host only implementation.

### DIFF
--- a/nimble/host/src/ble_hs_id.c
+++ b/nimble/host/src/ble_hs_id.c
@@ -58,11 +58,14 @@ ble_hs_id_set_rnd(const uint8_t *rnd_addr)
 {
     uint8_t addr_type_byte;
     int rc;
+    uint8_t all_zeros[BLE_DEV_ADDR_LEN] = {0}, all_ones[BLE_DEV_ADDR_LEN] = { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
 
     ble_hs_lock();
 
+    /* Make sure all bits are neither one nor zero */
     addr_type_byte = rnd_addr[5] & 0xc0;
-    if (addr_type_byte != 0x00 && addr_type_byte != 0xc0) {
+    if ((addr_type_byte != 0x00 && addr_type_byte != 0xc0) ||
+        !memcmp(rnd_addr, all_zeros, BLE_DEV_ADDR_LEN) || !memcmp(rnd_addr, all_ones, BLE_DEV_ADDR_LEN)) {
         rc = BLE_HS_EINVAL;
         goto done;
     }
@@ -283,3 +286,13 @@ ble_hs_id_reset(void)
     memset(ble_hs_id_pub, 0, sizeof ble_hs_id_pub);
     memset(ble_hs_id_rnd, 0, sizeof ble_hs_id_pub);
 }
+
+/**
+ * Clears random address. This function is necessary when the host wants to
+ * clear random address.
+ */
+ void
+ ble_hs_id_rnd_reset(void)
+ {
+    memset(ble_hs_id_rnd, 0, sizeof ble_hs_id_rnd);
+ }
--- a/nimble/host/src/ble_hs_id_priv.h
+++ b/nimble/host/src/ble_hs_id_priv.h
@@ -31,6 +31,7 @@ int ble_hs_id_addr(uint8_t id_addr_type, const uint8_t **out_id_addr,
                    int *out_is_nrpa);
 int ble_hs_id_use_addr(uint8_t addr_type);
 void ble_hs_id_reset(void);
+void ble_hs_id_rnd_reset(void);
 
 #ifdef __cplusplus
 }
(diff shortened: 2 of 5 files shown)


nimble/host/src/ble_hs_id.c
```diff
@@ -58,9 +58,21 @@ ble_hs_id_set_rnd(const uint8_t *rnd_addr)
...
 
     ble_hs_lock();
 
+    /* Make sure all bits are neither one nor zero */
+     sum = 0;
+     for (rc = 0; rc < BLE_DEV_ADDR_LEN; ++rc) {
+         sum += rnd_addr[rc];
+     }
+
+     if ((sum == 0) || (sum == (6*255))) {
```
andrzej-kaczmarek: that's counter-intuitive tbh, why not just check if any byte is other that 0x00 or 0xff in loop above?

prasad-alatkar: Well I just tried to keep the implementation similar to `ble_ll_is_valid_random_addr`.

prasad-alatkar:
Hi @andrzej-kaczmarek, Can we use `ble_hs_id_reset` in API `ble_hs_test_util_init` instead of `ble_hs_test_util_set_static_rnd_addr((uint8_t[6]){ 0, 0, 0, 0, 0, 0 })` ? Because with the change in this PR , `ble_hs_id_set_rnd` will return `BLE_HS_EINVAL` if we try to set "rnd_addr" as "all zeros".

prasad-alatkar:
Hi @andrzej-kaczmarek , made few changes to original PR. However it is still failing the check. I am not sure why this assertion error should come. Can you please take a look? `nimble_host_test.elf: nimble/host/src/ble_hs.c:602: ble_hs_sched_reset: Assertion `ble_hs_reset_reason == 0' failed.
Error: Test failure(s):`

ccollins476ad:
@prasad-alatkar, the test is failing for a complicated reason.  The nimble host unit tests are very difficult to work with because they require you to schedule every HCI event that the host expected to receive.

You removed this line in the test code.
```
     /* Clear random address. */
-    ble_hs_test_util_set_static_rnd_addr((uint8_t[6]){ 0, 0, 0, 0, 0, 0 });
```

This change looks innocent (and it would be innocent if the nimble test code weren't so nasty!), but it has an unexpected side-effect.  In addition to clearing the random static address, the call to `ble_hs_test_util_set_static_rnd_addr()` also clears all scheduled HCI "acks".  One unit test leaves some scheduled acks in the queue, and the subsequent test was relying on this call to `ble_hs_test_util_set_static_rnd_addr()` to clear them.

These unit tests really need an overhaul.  There has been some discussion about this and we plan to do it, but it is a big job that hasn't been started.

In the meantime, I think this is the correct fix:
```
diff --git a/nimble/host/test/src/ble_hs_test_util.c b/nimble/host/test/src/ble_hs_test_util.c
index 2230423d..d79671f8 100644
--- a/nimble/host/test/src/ble_hs_test_util.c
+++ b/nimble/host/test/src/ble_hs_test_util.c
@@ -2044,6 +2044,7 @@ ble_hs_test_util_init(void)
     TEST_ASSERT_FATAL(rc == 0);
 
     ble_hs_test_util_hci_out_clear();
+    ble_hs_test_util_hci_acks_clear();
 
     /* Clear random address. */
     ble_hs_id_rnd_reset();
diff --git a/nimble/host/test/src/ble_hs_test_util_hci.c b/nimble/host/test/src/ble_hs_test_util_hci.c
index b9b36d10..3d2eceaa 100644
--- a/nimble/host/test/src/ble_hs_test_util_hci.c
+++ b/nimble/host/test/src/ble_hs_test_util_hci.c
@@ -107,6 +107,12 @@ ble_hs_test_util_hci_out_clear(void)
     ble_hs_test_util_hci_out_queue_sz = 0;
 }
 
+void
+ble_hs_test_util_hci_acks_clear(void)
+{
+    ble_hs_test_util_hci_num_acks = 0;
+}
+
 /*****************************************************************************
  * $build                                                                    *
  *****************************************************************************/
diff --git a/nimble/host/test/src/ble_hs_test_util_hci.h b/nimble/host/test/src/ble_hs_test_util_hci.h
index 500f8ed6..b941a0c2 100644
--- a/nimble/host/test/src/ble_hs_test_util_hci.h
+++ b/nimble/host/test/src/ble_hs_test_util_hci.h
@@ -46,6 +46,7 @@ void *ble_hs_test_util_hci_out_first(void);
 void *ble_hs_test_util_hci_out_last(void);
 void ble_hs_test_util_hci_out_enqueue(void *cmd);
 void ble_hs_test_util_hci_out_clear(void);
+void ble_hs_test_util_hci_acks_clear(void);
 
 /* $build */
 void ble_hs_test_util_hci_build_cmd_complete(uint8_t *dst, int len,
```

You can add this to your PR if you like, or I can submit a separate PR with the fix.

I did not look into why a test leaves scheduled acks on the queue.  That is another investigation for another day!

prasad-alatkar:
Thank you @ccollins476ad for detailed explanation. I will include the changes suggested by you !!

prasad-alatkar:
Hi @andrzej-kaczmarek , this PR is ready, Can you please merge this ?

prasad-alatkar:
Hi @andrzej-kaczmarek Can you merge this PR?

PR #439, opened 2019-05-14
mesh/adv.c: fix advertising duration

haukepetersen:
The default configuration for link layer (re)transmission in the `Configuration Server Model` for the `Network Transmit State` and the `Relay Retransmit State` is **3 transmissions** with an interval of **20ms**. But after doing some closer analysis, I discovered that the implementation actually does 5-6 transmissions.

Looking at `nimble/host/mesh/src/adv.c` line 108ff, it seems to me, that the calculation of the duration value is flawed -> adding in `MESH_SCAN_WINDOW_MS` should not be needed, right?!

With this fix, I see exactly the 3 transmissions (advertising events) that one would expect from the Configuration Server's Transmission State.

Does this make sense or did I miss something here?

### DIFF
--- a/nimble/host/mesh/src/adv.c
+++ b/nimble/host/mesh/src/adv.c
@@ -105,9 +105,14 @@ static inline void adv_send(struct os_mbuf *buf)
 
 	adv_int = max(adv_int_min,
 		      BT_MESH_TRANSMIT_INT(BT_MESH_ADV(buf)->xmit));
+#if MYNEWT_VAL(BLE_CONTROLLER)
+	duration = ((BT_MESH_TRANSMIT_COUNT(BT_MESH_ADV(buf)->xmit) + 1) *
+				(adv_int + 10));
+#else
 	duration = (MESH_SCAN_WINDOW_MS +
 		    ((BT_MESH_TRANSMIT_COUNT(BT_MESH_ADV(buf)->xmit) + 1) *
 		     (adv_int + 10)));
+#endif
 
 	BT_DBG("type %u om_len %u: %s", BT_MESH_ADV(buf)->type,
 	       buf->om_len, bt_hex(buf->om_data, buf->om_len));


rymanluk:
@haukepetersen original patch adding this is   6e2c7d190 which solves some problems with advertising scheduler when scanning is ongoing for zephyr controller.

I think for Nimble controller this is not an issue, because controller schedules advertising over scanning. But might be for others. Maybe we should have MYNEWT_VAL here? Need to think about it. 

For sure, whatever we do we need good commit message saying why we are doing this.

haukepetersen:
Ok, so it seems that this is really only a Zephyr specific fix and we might want to remove it from the code. If I see it correctly, `adv.c` is already (at least partly) specific to nimble, is it not? So IMHO it would make sense to simply drop that window offset here. Using a good commit message :-)


> For sure, whatever we do we need good commit message saying why we are doing this.

+1. My PR was rather meant to underline my issue then to have it merged right away :-)

rymanluk:
@haukepetersen How many other controllers did you test? If you haven't seen the issue with others in such scenario then I'm fine with dropping it as you did in you path.

haukepetersen:
> How many other controllers did you test?

Only the NimBLE (nrf) controller on `nrf52dk` boards. So no idea how other controller take this... So I guess some further expertise for other controllers is in dire need here.

haukepetersen:
So for now, I made the fix to apply when using the Nimble controller only. Do you think this would work for now?

haukepetersen:
but in general, I think this code path is still not very nice: 
- depending on the random jitter values between the advertisement, we still don't always send the exact amount of advertisement messages that are configured
- having the mesh thread block for the advertisement duration will lead to overflowing buffers in case the stack is under stress. It would be way nicer, if we would be able to send multiple advertisements in parallel at some point

Maybe one step could be to  have some GAP API call like `ble_gap_send_adv_num(.., num_events)`, where we could specify the number of advertising events until stopped. Has something like this been thought of at some point?

nimble/host/mesh/src/adv.c
```diff
@@ -105,9 +105,14 @@ static inline void adv_send(struct os_mbuf *buf)
 
 	adv_int = max(adv_int_min,
 		      BT_MESH_TRANSMIT_INT(BT_MESH_ADV(buf)->xmit));
+#ifdef MYNEWT_VAL(BLE_CONTROLLER)
```
rymanluk: I believe it should be #if MYNEWT_VAL(BLE_CONTROLLER) as we want to check for value 1

haukepetersen: yes, my bad. will fix

haukepetersen:
fixed broken #if condition, and squashed the changes right away (hope that was ok...).

haukepetersen:
@rymanluk Would you mind to take another look? All changes should be valid now. Thx.

rymanluk:
@haukepetersen thanks, lets have it for now, I guess at some point we will remove the alternative path.

PR #444, opened 2019-05-20
Adding System ID characteristic to DIS.

mlaz:
This PR adds the System ID characteristic to DIS. 
Also fixes some indentation inconsistencies.

### DIFF
--- a/nimble/host/services/dis/syscfg.yml
+++ b/nimble/host/services/dis/syscfg.yml
@@ -29,8 +29,8 @@ syscfg.defs:
     BLE_SVC_DIS_MODEL_NUMBER_READ_PERM:
         description: >
             Defines permissions for reading "Model Number" characteristics.
-            Can be set to MYNEWT_VAL_BLE_SVC_DIS_DEFAULT_READ_PERM or use
-            the possible values defined in BLE_SVC_DIS_DEFAULT_READ_PERM.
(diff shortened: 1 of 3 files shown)


nimble/host/services/dis/syscfg.yml
```diff
@@ -92,6 +92,17 @@ syscfg.defs:
             Defines a default value for "Manufacturer name" if not set with
             'ble_svc_dis_manufacturer_name_set'.
         value: NULL
+    BLE_SVC_DIS_SYSTEM_ID_READ_PERM:
+        description: >
+            Defines permissions for reading "System ID" characteristics.
+            Can be set to MYNEWT_VAL_BLE_SVC_DIS_DEFAULT_READ_PERM or use
+            the possible values defined in BLE_SVC_DIS_DEFAULT_READ_PERM.
+        value: MYNEWT_VAL_BLE_SVC_DIS_DEFAULT_READ_PERM
```
ccollins476ad: For references to other settings, it is slightly better to use
```
MYNEWT_VAL(BLE_SVC_DIS_DEFAULT_READ_PERM)
```
If the referenced setting doesn't exist, newt will produce an error message.  This is a little better than the alternative, because the newt error is easier to understand than the compiler error.

mlaz: Just changed it.

PR #457, opened 2019-06-03
nimble/mesh: Fix missing SAR timeout in Proxy Server

michal-narajowski:
Mesh Profile 1.0 Section 6.6:
"The timeout for the SAR transfer is 20 seconds. When the timeout expires,
the Proxy Server shall disconnect."

This allows passing MESH/SR/PROX/BV-05-C.

### DIFF
--- a/nimble/host/mesh/src/proxy.c
+++ b/nimble/host/mesh/src/proxy.c
@@ -30,6 +30,12 @@
 #define PDU_TYPE(data)     (data[0] & BIT_MASK(6))
 #define PDU_SAR(data)      (data[0] >> 6)
 
+/* Mesh Profile 1.0 Section 6.6:
+ * "The timeout for the SAR transfer is 20 seconds. When the timeout
+ *  expires, the Proxy Server shall disconnect."
+ */
+#define PROXY_SAR_TIMEOUT  K_SECONDS(20)
+
 #define SAR_COMPLETE       0x00
 #define SAR_FIRST          0x01
 #define SAR_CONT           0x02
@@ -117,6 +123,7 @@ static struct bt_mesh_proxy_client {
 #if (MYNEWT_VAL(BLE_MESH_GATT_PROXY))
 	struct ble_npl_callout send_beacons;
 #endif
+	struct k_delayed_work sar_timer;
 	struct os_mbuf    *buf;
 } clients[MYNEWT_VAL(BLE_MAX_CONNECTIONS)] = {
 	[0 ... (MYNEWT_VAL(BLE_MAX_CONNECTIONS) - 1)] = { 0 },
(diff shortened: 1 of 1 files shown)


nimble/host/mesh/src/proxy.c
```diff
@@ -401,6 +404,24 @@ static void proxy_send_beacons(struct ble_npl_event *work)
...
+static void proxy_sar_timeout(struct ble_npl_event *work)
+{
+	struct bt_mesh_proxy_client *client;
+	int rc;
+
+	client = ble_npl_event_get_arg(work);
+	assert(client != NULL);
+
+	if ((client->conn_handle != BLE_HS_CONN_HANDLE_NONE) &&
+	    (client->filter_type == WHITELIST ||
```
rymanluk: why filter_type is important here?

rymanluk:
@jhedberg

nimble/host/mesh/src/proxy.c
```diff
@@ -401,6 +404,21 @@ static void proxy_send_beacons(struct ble_npl_event *work)
...
 
+static void proxy_sar_timeout(struct ble_npl_event *work)
+{
+	struct bt_mesh_proxy_client *client;
+	int rc;
+
+	client = ble_npl_event_get_arg(work);
+	assert(client != NULL);
+
+	if ((client->conn_handle != BLE_HS_CONN_HANDLE_NONE)) {
```
jhedberg: Wouldn't it be better to cancel the delayed work in the disconnected callback? Otherwise you risk the client object getting reused with another valid handle by the time the timeout expires.

michal-narajowski:
@jhedberg Yes, MESH/SR/PROX/BV-05-C failed for us, because we didn't disconnect the link. In Zephyr, it was done by sending gap_disconnect on a specific WID, but that approach didn't work in Mynewt.

jhedberg:
@michal-narajowski what's "WID"? By gap_disconnect are you referring to the BTP protocol? I.e. was this worked around by requesting a disconnect explicitly? Meaning, even though the test might have been passing for Zephyr this patch should still be backported?

nimble/host/mesh/src/proxy.c
```diff
@@ -1415,6 +1435,8 @@ int bt_mesh_proxy_init(void)
 	for (i = 0; i < MYNEWT_VAL(BLE_MAX_CONNECTIONS); ++i) {
 #if (MYNEWT_VAL(BLE_MESH_GATT_PROXY))
 		k_work_init(&clients[i].send_beacons, proxy_send_beacons);
+		k_delayed_work_init(&clients[i].sar_timer, proxy_sar_timeout);
+		k_delayed_work_add_arg(&clients[i].sar_timer, &clients[i]);
```
jhedberg: You're adding the the sar_timer struct member unconditionally, but here you're only initialising it if BLE_MESH_GATT_PROXY is set? Isn't SAR also used for PB-GATT PDUs?

michal-narajowski:
WID = Window ID. Each window that PTS displays has an ID. Yes, the disconnect was explicit and I think this patch should be backported to Zephyr.

nimble/host/mesh/src/proxy.c
michal-narajowski: Thank you, fixed.

nimble/host/mesh/src/proxy.c
michal-narajowski: Thank you, fixed.

PR #462, opened 2019-06-05
nimble/host: Fix compilation error

dhrishi:
Fixed the below error:

    error: format '%x' expects a matching 'unsigned int' argument [-Werror=format=]
             BLE_HS_LOG(DEBUG, "LE Advertising Set Terminated len=%u "


dhrishi:
@rymanluk Done! Updated the commit message as well as the description here

PR #467, opened 2019-06-07
nimble/host: Add support to stop/de-init ports

dhrishi:
This PR adds the support to stop and deinit NimBLE ports. It is a common usecase, where BLE needs to be turned on whenever required and turned off when the job is done. 

This PR adds new APIs `nimble_port_stop` and `nimble_port_deinit` for applications to use. Additionally, `nimble_port_freertos_deinit` API is added for the FreeRTOS port.

I have made the changes and tested them for FreeRTOS and they seem to work fine. If this PR looks okay, can someone from the NimBLE team help with making changes for other ports as well?


porting/nimble/src/nimble_port.c
```diff
@@ -55,17 +58,75 @@ nimble_port_init(void)
...
+    /* Initiate a host stop procedure. */
+    rc = ble_hs_stop(&stop_listener, ble_hs_stop_cb,
+            NULL);
+    if (rc != 0) {
+        ble_npl_sem_deinit(&ble_hs_stop_sem);
+        return rc;
+    }
+
+    /* Wait till the host stop procedure is complete */
+    ble_npl_sem_pend(&ble_hs_stop_sem, BLE_NPL_TIME_FOREVER);
```
ccollins476ad: This function (`nimble_port_stop()`) can't be called from the same task that the host runs in, is that correct?  If so, that is probably worth mentioning in a comment.

dhrishi: That's right. I will add a comment mentioning the same.

ccollins476ad:
It has been a while since I've done any work on nimble, so I'm not qualified to review this.  But I am surprised that this change is necessary.  I am pretty sure I have tested stopping and starting nimble in Mynewt without issue (just using `ble_hs_stop()` and `ble_hs_start()`).  Does that approach leak resources or something?

dhrishi:
ble_hs_stop() stops all the stuff related to the host but nothing about the port (queue, sem, mutex, task) is freed.

And yes, some things in the host too are not deallocated. They are eventually done when you reinitialize.

Feel free to correct me in case I am missing anything.

porting/nimble/src/nimble_port.c
dhrishi: @ccollins476ad Done!

nimble/host/src/ble_hs_hci.c
```diff
@@ -643,3 +643,11 @@ ble_hs_hci_init(void)
...
     BLE_HS_DBG_ASSERT_EVAL(rc == 0);
 }
+
+void
+ble_hs_hci_deinit(void)
+{
+    ble_npl_mutex_deinit(&ble_hs_hci_mutex);
+
+    ble_npl_sem_deinit(&ble_hs_hci_sem);
+}
```
ccollins476ad: Should this function (`ble_hs_hci_deinit`) also include:
```
ble_npl_eventq_deinit(&ble_hci_sock_state.evq);
```

?

porting/nimble/src/nimble_port.c
```diff
@@ -55,17 +58,79 @@ nimble_port_init(void)
...
 nimble_port_run(void)
 {
     struct ble_npl_event *ev;
+    int arg;
 
     while (1) {
         ev = ble_npl_eventq_get(&g_eventq_dflt, BLE_NPL_TIME_FOREVER);
         ble_npl_event_run(ev);
+        arg = (int)ble_npl_event_get_arg(ev);
+        if (arg == NIMBLE_PORT_DEINIT_EV_ARG) {
```
ccollins476ad: My feeling is that it would be better to compare:
```
ev == &ble_hs_ev_stop
```

here rather than using an argument to identify this event.  Or at least I think it would be better to use a less likely value than `-1`.

nimble/host/src/ble_hs_hci.c
dhrishi: If I am not wrong, then it is used for socket transport. If you see, `ble_hci_sock_init` in which the eventq is inited is called from the application and not from the host stack. So, ideally its corresponding deinit function should also be a part of a separate API like `ble_hci_sock_deinit` which should be called by the application (only for socket transport). Also, `ble_hs_hci_deinit` is generic and required for all transports. 

I had verified with my PR that there were no leaks between start and stop (we do not use socket transport)

porting/nimble/src/nimble_port.c
dhrishi: That's right. I totally missed that this could be used. I will update the PR

dhrishi:
@ccollins476ad  Thanks for the review.
 
> 1. Mynewt implementations of the "deinit" functions.
We are working on FreeRTOS port and have done the changes for the same (and dummy one).
I am not much aware about other ports (MyNewt, riot, Linux) and that's why in the PR description I asked someone from the NimBLE team to assist with these APIs. Please let me know if someone can help in this regard.

> 2. A deinit function for the mesh code.
Yes. The current PR does not handle deinit for mesh code. The reasons being:
1. It is not a common usecase to disable/re-enable BLE mesh as compared to standard BLE.
2. This is not our requirement as of now.

I might add this support sometime in the future.

porting/nimble/src/nimble_port.c
dhrishi: Done!

PR #468, opened 2019-06-07
apps/btshell: Update conn-datalen params help to be more informative

michal-narajowski:


apps/btshell/src/cmd.c
```diff
@@ -2203,9 +2203,11 @@ cmd_conn_datalen(int argc, char **argv)
 
 #if MYNEWT_VAL(SHELL_CMD_HELP)
 static const struct shell_param conn_datalen_params[] = {
-    {"conn", "conn_datalen handle, usage: =<UINT16>"},
-    {"octets", "usage: =<UINT16>"},
-    {"time", "usage: =<UINT16>"},
+    {"conn", "Connection handle, usage: =<UINT16>"},
+    {"octets", "Max payload size to include in LL Data PDU, "
+               "range=<0x001B-0x00FB>, usage: =<UINT16>"},
```
andrzej-kaczmarek: decimal numbers would be so much nicer here

apps/btshell/src/cmd.c
```diff
@@ -2203,9 +2203,11 @@ cmd_conn_datalen(int argc, char **argv)
...
+    {"time", "Max number of microseconds the controller should use to tx "
+             "single LL packet, range=<0x0148-0x4290>, usage: =<UINT16>"},
```
andrzej-kaczmarek: and here

apps/btshell/src/cmd.c
michal-narajowski: Agreed, fixed.

apps/btshell/src/cmd.c
michal-narajowski: Agreed, fixed.

PR #477, opened 2019-06-12
npl/riot: reduce default MSYS_1_BLOCK_SIZE to minimum

haukepetersen:
The default configuration of `MYNEWT_VAL_MSYS_1_BLOCK_SIZE` for the RIOT port uses a pretty large value, preventing NimBLE to run with this configuration on `nrf51`-based boards. Also, I can't quite explain, why the value of `292` is actually chosen, it seems quite random?!

So with this PR I want to reduce the default `MSYS_1_BLOCK_SIZE` to the minimum possible for the typical GATT server/client and (non-extended) advertising scenarios. The idea is, to have the minimum value as default, and set it to something larger from inside the RIOT configuration if needed (e.g. for IP-over-BLE).

 As I see it, this should be `70` bytes:
- 24 bytes `mbuf` overhead (`sizeof(struct os_mbuf) + sizeof(struct os_mbuf_pkthdr)`)
- 46 bytes BLE packet size (4 access address, 2 ll-header, 37 PDU, 3 CRC) -> BT4.0, without data packet length extension

I gave this setting a quick try and the RIOT-based GATT server example works as expected. BUT I can't really tell this is a coincidence or if my calculations/assumptions are actually correct. So maybe someone could shed some light on the following questions:
- would `70` bytes work?
- how is `MYNEWT_VAL_BLE_ATT_PREFERRED_MTU` related and does it have any impact? It seems not be used anywhere in the code...
- could we also restrict `MYNEWT_VAL_MSYS_1_BLOCK_COUNT` do some lower default value?


haukepetersen:
Hej everyone, is there anybody that could clarify the items above? I would like to go forward on this to finally enable `nrf51` support for NimBLE in RIOT...

wes3:
@haukepetersen  I can try to answer some of these but not all. First off, I do not think folks will have any issue changing the RIOT default values for the default block size. As to your calculation of 70 there are some things you do not need for the nrf51. It has been a very long time since I looked at the nrf51 but a couple of things I am pretty sure of: you do not need to account for the access address or CRC in your calculation. They are handled by HW if I remember correctly. There is also a structure used by the controller and that is called struct ble_mbuf_hdr. Not sure if you are using the nimble controller but if you are you will need to account for that.

You could also restrict the default number of msys blocks; that will definitely help.

Finally, if recollection serves, the 292 number comes from a maximum size PDU of 251 bytes using data length extension. I thought there was some documentation somewhere stating how that number was calculated. It certainly was not "random" :-) but without knowing exactly how it got calculated I could see how one would think that.

Anyway, hope the above helps.

haukepetersen:
@wes3 thanks for you feedback!

Regarding my calculation: you are right about the access address, forgot about the nRFs easyDMA in-memory packet format... So taking this and the `struct ble_mbuf_hdr` into account, it should be:
- 24 bytes mbuf overhead (sizeof(struct os_mbuf) + sizeof(struct os_mbuf_pkthdr))
- 12 bytes `sizeof(struct ble_mbuf_hdr)`
- 39 bytes packet payload (w/o access address and w/o crc)

Sums up to `75 bytes` minimal viable buffer size, right?

> You could also restrict the default number of msys blocks; that will definitely help.

Of course. But the question here as well, what would be the minimal viable value here? 

> Finally, if recollection serves, the 292 number comes from a maximum size PDU of 251 bytes using data length extension. I thought there was some documentation somewhere stating how that number was calculated. It certainly was not "random" :-) but without knowing exactly how it got calculated I could see how one would think that.

The 'random' was of course meant very seriously :-)

 But I would be highly interested in the exact calculation that leads to this value, as between the max MTU of `251` and `292` there are still `41` bytes only partly accounted for.

Since I work with NImBLE, the most trouble I had was related to buffer management, as NimBLE seems to be very sensitive about running out of them... So far, I simply generously over provision them, but this is of course only a temporary solution.

My goal is to come up with some kind of elaborate documentation, that describes precisely how the buffers are used and how to adapt and configure them for ones specific use-case. I thought clarifying the above would be a good starting point :-)

haukepetersen:
Just gave this some tests using RIOTs `examples/nimble_gatt` example application, using the Nordic `nRF connect` Android app as GATT client. I deployed it on two different targets, `nrf52dk` and `microbit (nrf51)`.

The bare minimum numbers, for which the GATT server was still working were
- `MYNEWT_VAL_MSYS_1_BLOCK_SIZE := 77`
- `MYNEWT_VAL_MSYS_1_BLOCK_COUNT := 4`

For any block-size smaller than `77 bytes` I can't connect to the GATT server, so there must be something still missing in the numbers above?!

For any block count <4 the phone can not connect to the device. I tried to figure out why 4 buffers are needed, but I am unsure. So in the most busy case, I get:
- 1 mbuf allocated by the controller to receive the next packet
- 1 mbuf allocated by the controller for keeping an empty packet to reply on the next connection event in case there is no GATT payload to be send
- 1 mbuf holding data currently processed by the GATT server
- 1 mbuf allocated by the GATT server for sending a reply?

wes3:
@haukepetersen I can understand why you might want to try to come up with an absolute minimum number of buffers used by nimble but I think it might end up being a bit tricky. Unfortunately I do not know the host well enough to comment on what you have mentioned above so others would have to chime in.

Regarding the controller: the controller does not require a mbuf to send an empty pdu; there is an empty pdu structure it uses for this. Of course, the controller needs a mbuf to receive something from the peer controller so if there are none available no data will flow in that direction. Note that the connection will still continue in this case; the controller simply does not ack the transmission and the transmission should keep being repeated by the peer controller until a mbuf is available for reception.

I will see if I can dig up how the 292 was calculated.

andrzej-kaczmarek:
Proper minimum value for `MSYS_1_BLOCK_SIZE` is 80, assuming you do not use extended advertising. This allows to store `CONNECT_IND` PDU in a single mbuf, i.e. without creating a chain.

To store complete `CONNECT_IND` PDU we need 76 bytes:
- 16 bytes for `os_mbuf` header
- 8 bytes for `os_mbuf_pkthdr`
- 16 bytes for `ble_mbuf_hdr`
- 36 bytes for LL PDU with `CONNECT_IND` payload (2 + 34)
However, to move packets from RX ISR to LL task we use `ble_ll_rxpdu_alloc` which adds 4 bytes at the beginning of 1st mbuf for extra header used when moving ACL data to HCI transport. This gives 80 bytes in total. The reason why 77 works for you is because block size is aligned to word size, so it's 80 anyway.

Note that even with lower value we can pass PDUs from RX ISR to LL task properly by using mbuf chains, however there are few places in code which assume that complete PDU is available in first mbuf as this simply makes code faster (you can read data from flat array) and this is exactly the case when parsing `CONNECT_IND` - with chain in mbuf we'll just read some random data for certain parameters and connection won't work.

As for minimum number of blocks, I do not know what is the absolute minimum required but anyway I do not recommend to set this to such value since it can simply make things fail randomly. For example, you do not really know when received ACL data will be processed and freed by host (assuming RAM transport is used for HCI) and while they are queued either in controller or in host you may not be able to receive new packets. Of course we'll NAK incoming packets so eventually they will be received (unless supervision timeout hits) but performance won't be great. I do not also know what is a reasonably good low value though.

haukepetersen:
Slowly I start to see some light in the end of that tunnel :-)

The 80 bytes minimum size for `MSYS_1_BLOCK_SIZE` makes sense and is now clear to me. I will adapt the value accordingly.

However, I think the error handling in the NimBLE code could be vastly improved - having to actually now about these specifics from a user perspective is quite a lot to ask. And as for my experiences: when badly configuring buffers you end up with a silent fail, which is quite frustrating...

> however there are few places in code which assume that complete PDU is available in first mbuf

that is exactly one of those assumptions that is IMHO not handled very user friendly in case of bad configuration :-)

> As for minimum number of blocks, I do not know what is the absolute minimum required [...]

That does not sound good. For sure there can't the one general number that is valid for all the different possible transports etc, but for the specific use-case of running GATT over RAM transport using the NimBLE controller, I would think that there is value that can be stated. Not talking about degraded performance here (which is a tradeof a user has to chose), but a safe value that will at least not let the stack crash.

> [...] but anyway I do not recommend to set this to such value since it can simply make things fail randomly.

This comes pretty much to the core of the problems I was having: NimBLE has a tendency for random failures when it runs out of buffers. To me, this is unacceptable! I would rather expect NimBLE to (more or less silently) drop packets, or at least fail using assertions when in devel mode.

For now I can live with simply setting this value to something I found reasonable in testing, but it does not feel right with without being able to justify exactly why this value is working and the value-1 is not...

In general, I would very much like to see some improvements regarding the memory management: 
- better documentation about how to configure the buffers for selected usecases (transports, protocols used, ...)
- better error handling inside the host/controller when running out of buffers
I will gladly help to do this as far as I can, but there are apparently a lot things that exceed my knowledge :-)

simonratner:
>Proper minimum value for MSYS_1_BLOCK_SIZE is 80, assuming you do not use extended advertising. 

There are some old discussion about this which may help:
* https://www.mail-archive.com/dev@mynewt.apache.org/msg00240.html
* https://www.mail-archive.com/dev@mynewt.incubator.apache.org/msg02408.html

That discussion suggests minimum of 84 bytes without, and 88 bytes with extended advertising enabled. Has some struct been reduced in size recently (since v1.2)?

simonratner:
By the way, I would suggest setting the default to at least a value that can support extended advertising, even if that is disabled by default. Otherwise, enabling that feature will suddenly break things in a subtle way that most people will not understand.

haukepetersen:
This discussion is quite fruitful for me, and I hope I don't start to annoy anyone :-)

To go forward, I'd like to suggest the following:
- get this PR merged with the current values (80 bytes msys bufsize, 4 msys buffers) -> so is it ok if I squash?
- move/continue this discussion in a separate issue -> I will gladly open one an summarize what we discussed so far.

Only I am offline on holiday for the next 3 weeks, so you won't hear back from me until August...

haukepetersen:
@simonratner my browser did not refresh in properly, so I only saw your comment just now. Could you maybe elaborate, where those 'new' 4 byte (for non extended advertising) come from? Adding the additional 4 byte to cater for extended advertising as safety margin does indeed not hurt -> adapted the PR accordingly.

haukepetersen:
Oh, I am too tired this morning, everything I needed is actually in the email archive posted above...

To summarize (one again), for the current master, the numbers seem to be (just tested on a `nrf51` cortex-m0):
- `sizeof(struct ble_mbuf_hdr)` -> 16 bytes / 20 bytes (w and w/o extended adv)
- `sizeof(struct os_mbuf)` -> 16 bytes
- `sizeof(struct os_mbuf_pkthdr)` -> 8 bytes
- HCI ACL data header: 4 bytes
- PDU: 39 bytes

-> sums up to 83 bytes / 87 bytes -> 88 bytes still seem valid with the current code base.

haukepetersen:
so from what I learned, I further disabled the controller's data length extension by default for the RIOT port. As the `msys` buffer sizes are in the default config now fairly minimal, it seems to contradicting to allow for large link layer packtes at the same time. 

Also it fits better into the overall integration strategy for RIOT: per default, NimBLE provides a minimal configuration, and the RIOT build system takes care of (re)enabling all the features on demand.

andrzej-kaczmarek:
Note: to the best of my knowledge things work in NimBLE as I describe below, but there is very slight chance that some of numbers here are incorrect so please correct me with a concrete proof in such case.

@haukepetersen 
We definitely should improve sanity checks to make sure configuration makes sense, especially if there are settings like `MSYS_1_BLOCK_SIZE` where we know that certain size is simply required so we do not try to read past allocated buffers. You can create an issue for this so we can continue discussion and perhaps someone can pick this up in spare time (perhaps some volunteer would like to gain a bit of knowledge on LL internals?).

As for random failures and better error handling, I do not think there is any major issue in NimBLE in this regard. I mean, if we run out of buffers we don't just crash (if we do, this is an issue and not deliberate action) but try to do best we can that is allowed by spec, e.g. reply with an error or drop a packet (to be more precise: on RX we do not simply drop a packet, but we NAK it properly so it will be retransmitted). While it may seem like "random failures", this is not because of inherent randomness in NimBLE but due to Bluetooth traffic that is pretty much different every time you connect. Also, failures are usually logged to stats which can give a better understanding of what happens, but this is iirc not available in ports - feel free to port it :-)

And now `MSYS_1_BLOCK_COUNT`:
- 1 block is absolutely required for LL to start since it's allocated by scansm as a placeholder for SCAN_REQ PDU; by quickly looking through code I am pretty sure it would be possible to remove this requirement now since it should be possible to create SCAN_REQ on the fly (it was not possible when this was implemented few years ago)
- 1 block is required to receive any LL packet and send reply if needed, because RX PDU block is reused for TX PDU in such case

This means with only 2 blocks it is possible to create and maintain connection, and even receive some data. This is also enough for (G)ATT server to reply on requests in general, however reading attribute from ATT database requires an extra block for attribute value so some (most?) of those requests will fail with only 2 block available. With 3 blocks 80 bytes each you can have working connection with usable GATT (it works, I checked using `btshell`).

So why 3 does not work for you? I do not know, could be that phone you try to connect with does some procedures which requires an extra block and it's just not possible to handle all control and data packets before e.g. timeout occurs. Each device behaves differently and that's basically why using an "absolute minimum" value is not really a good solution. Perhaps something like "recommended value" would be more appropriate here and this would be default value since they are adjusted in case we find current value is not good for some reason.

One more example to give you better idea on why a single "absolute minimum" value won't work with all devices. While checking the above I found an issue in code handling `ATT Read By Group Type Request` where we can indeed silently drop packet inside ATT server due to oom. To be clear here, we do check this pretty consistently in code but there is single execution path where ATT error is not set properly which causes us to drop error response. This has an interesting consequence because it is actually what makes (G)ATT connection work with my Android phone using 3 blocks only. What happens after connection is that phone sends control PDU to update connection parameters and data packet with mentioned request - both consume remaining 2 blocks. We silently (and incorrectly) fail to reply on ATT request and apparently Android for whatever reason will retry after few seconds when LL procedure is completed so we have 2 blocks to handle request and it will work just fine. However, once I fix this issue (which is a simple one-liner) (G)ATT won't work because we will reply with an error on ATT request and phone will not retry so you basically can't discover services. So assuming I will fix the issue, what is an "absolute minimum" value here? I do not know, because I do not know how many other PDUs an Android phone can send us at once. As ambiguous as it sounds, "sane" is the proper value here.

As for data length extensions, small blocks do not mean you can't use this feature. You can assuming there are enough small blocks to hold long data PDU - we do not need a single block to hold complete data PDU and we'll do just fine with chained mbufs. In fact, having a lot of smaller blocks instead of few large blocks may be better on nRF51 since it allows to use them more efficiently. Using large blocks, whenever possible, means we do not need to chain them which improves performance a bit but you usually waste RAM by using large block for small PDU.

That said, since this applies to RIOT OS only I have no problem approving any values that are equal or above 3x80 (as this apparantly is what we can call "absolute mimimum" for legacy advertising + GATT over RAM transport scenario) - I am certain that what you select is tested and works for your purposes in RIOT OS so just let me know once you settled on final values in this PR.

@simonratner 
PDU handling has changed. Previously scanner was using a raw data pointer to read advertising PDU payload thus it was required that complete PDU fits in single block (similar as for `CONNECT_IND` I described previously). Currently scanner has acccess to complete mbuf so it just uses raw pointer to access few header bytes (which is guaranteed if block size is large enough to fit `CONNECT_IND`) and payload is copied using mbuf API so it can be fragmented in mbufs chain.

andrzej-kaczmarek:
update:
1. https://github.com/apache/mynewt-nimble/pull/509 is merged which means controller does no longer allocate spare block at startup (and does not need it later as well)
2. regarding why 3 blocks did not work - 4th block was most likely required for advertising data (I did not set them when testing)

so with latest NimBLE `MSYS_1_BLOCK_COUNT: 3` should work (advertiser, no ext adv, no data length extension), but may still randomly fail due to oom.

haukepetersen:
sorry for the late reply - the long holiday was very nice :-)

First of all, thanks again for the detailed explanations, this is very helpful. 

Also thanks for the improvements in #509. I can verify, that the `3x88` configuration is now working with my phone and both nrf51 and nrf52 based boards. I can however also observe the ATT issue pointed out above. Once again, the Bluetooth standard is allowing behavior that is not very nice for (very) contrained devices...

To settle this PR, I decided I'd like to go with 5 blocks of 88byes each as default configuration (according changes are force-pushed to this PR). This seems to be a good compromise allowing a little bit of variance concerning GATT/ATT behavior while still keeping memory consumption low enough for nrf51 targets.

andrzej-kaczmarek:
travis failed due to configuration issue, will merge anyway so there is no need to rebase it once more

haukepetersen:
awesome, thanks for sticking with me :-)

PR #466, opened 2019-06-05
nimble/ll: Fix aux_data ref counter

rymanluk:
This PR simplifies usage of aux_data ref counter.

PR #478, opened 2019-06-12
nimble/ll: Refactor duplicates filtering

andrzej-kaczmarek:
This patch refactors duplicates filtering to make it more robust by
auto-expiring entries, removes limitation of 255 entries in cache and
properly handles extended advertising reports for multiple sets and
anonymous devices.

Instead of a static array we use mempool of dup_entries which are linked
to create a queue where most recently found device is placed on top.
This allows to simply remove least recently found devices from queue
when list capacity is reached and thus keep filtering new devices.

Legacy advertising devices are simply matched by address thus each
device has single entry in the queue.
Extended advertising devices are matched by address and SID thus each
device can have multiple entries in queue (for each SID).

### DIFF
--- a/nimble/controller/src/ble_ll_scan.c
+++ b/nimble/controller/src/ble_ll_scan.c
@@ -54,9 +54,6 @@
  */
 
 /* Dont allow more than 255 of these entries */
-#if MYNEWT_VAL(BLE_LL_NUM_SCAN_DUP_ADVS) > 255
-    #error "Cannot have more than 255 duplicate entries!"
-#endif
 #if MYNEWT_VAL(BLE_LL_NUM_SCAN_RSP_ADVS) > 255
     #error "Cannot have more than 255 scan response entries!"
 #endif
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
this PR works better with https://github.com/apache/mynewt-nimble/pull/466 so probably that one should be merged first

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2847,6 +2709,187 @@ check_periodic_sync(const struct os_mbuf *om, struct ble_mbuf_hdr *rxhdr,
...
+
+static inline struct ble_ll_scan_dup_entry *
+ble_ll_scan_dup_new(void)
+{
+    struct ble_ll_scan_dup_entry *e;
+
+    e = os_memblock_get(&g_scan_dup_pool);
+    if (!e) {
+        e = TAILQ_LAST(&g_scan_dup_list, ble_ll_scan_dup_list);
+        TAILQ_REMOVE(&g_scan_dup_list, e, link);
```
sjanc: I'd memset entry here

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2847,6 +2709,187 @@ check_periodic_sync(const struct os_mbuf *om, struct ble_mbuf_hdr *rxhdr,
...
+    if (!aux_data) {
+        BLE_LL_ASSERT(0);
+        return 0;
+    }
+
+    is_anon = !(aux_data->flags & BLE_LL_AUX_HAS_ADDRA);
+    type = BLE_LL_SCAN_ENTRY_TYPE(addr_type, 1, is_anon, aux_data->adi);
+
+    TAILQ_FOREACH(e, &g_scan_dup_list, link) {
+        if ((e->type == type) &&
```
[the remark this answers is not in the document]
andrzej-kaczmarek: we do, SID is stored as a part of `type` when entry is for extended advertising (this is an entry type, not an address type). I'll put some comments in final version of this PR.

andrzej-kaczmarek:
this is now rebased on top of https://github.com/apache/mynewt-nimble/pull/466

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2840,6 +2702,178 @@ check_periodic_sync(const struct os_mbuf *om, struct ble_mbuf_hdr *rxhdr,
...
+static inline void
+ble_ll_scan_dup_move_to_head(struct ble_ll_scan_dup_entry *e)
+{
+    if (e != TAILQ_FIRST(&g_scan_dup_list)) {
+        TAILQ_REMOVE(&g_scan_dup_list, e, link);
+        TAILQ_INSERT_HEAD(&g_scan_dup_list, e, link);
+    }
+}
+
+
```
rymanluk: double empty line

andrzej-kaczmarek:
rebased, added proper support for ext without aux and some cleanup

PR #480, opened 2019-06-13
nimble/sm: Fix rejecting stronger bond

michal-narajowski:
If there exists an authenticated bond and the slave sends
a security request with No MITM we should not try to pair again
as the stored bond is stronger than the requested one.


michal-narajowski:
Fixed a typo in the description.

PR #482, opened 2019-06-13
Use common config option for periodic syncs count

sjanc:
This avoids accidental inconsistency in combined build.


sjanc:
@ahshokry @bassammohy FYI

PR #487, opened 2019-06-14
nimble/ll: Update DID when enabling or disabling periodic advertising

sjanc:
The Advertising DID is not required to change when a SyncInfo field is
added to or removed from an advertising set. However, if it does not
change, then scanners may fail to synchronize to periodic advertising
because entries in the Advertising DID cache (see Section 4.3.3) mean
they ignore the advertisements containing the SyncInfo field. Therefore,
advertisers should update the Advertising DID when a periodic advertising
train is enabled. Alternatively, the Host should enable periodic
advertising before enabling advertising.

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -2382,6 +2382,17 @@ ble_ll_adv_sm_start_periodic(struct ble_ll_adv_sm *advsm)
     uint32_t usecs;
     uint32_t ticks;
 
+    /*
+     * The Advertising DID is not required to change when a SyncInfo field is
+     * added to or removed from an advertising set. However, if it does not
+     * change, then scanners may fail to synchronize to periodic advertising
+     * because entries in the Advertising DID cache (see Section 4.3.3) mean
+     * they ignore the advertisements containing the SyncInfo field. Therefore,
+     * advertisers should update the Advertising DID when a periodic advertising
+     * train is enabled.
+     */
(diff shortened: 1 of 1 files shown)


nimble/controller/src/ble_ll_adv.c
```diff
@@ -2425,6 +2436,17 @@ ble_ll_adv_sm_stop_periodic(struct ble_ll_adv_sm *advsm)
...
+    /*
+     * The Advertising DID is not required to change when a SyncInfo field is
+     * added to or removed from an advertising set. However, if it does not
+     * change, then scanners may fail to synchronize to periodic advertising
+     * because entries in the Advertising DID cache (see Section 4.3.3) mean
+     * they ignore the advertisements containing the SyncInfo field. Therefore,
+     * advertisers should update the Advertising DID when a periodic advertising
+     * train is enabled.
+     */
+    ble_ll_adv_update_did(advsm);
```
andrzej-kaczmarek: do we need this here? it does not affect scanners' ability to sync and even comment above states that we should do this only when periodic is being enabled :-)

the only reason for doing this here I can think of is that we may want to notify host that syncinfo is no longer available on this instance, but not sure how what it would be useful for

sjanc: I'll update comment in disable section, but I think it is useful, cause remote scanner might be notified about no sync available before issuing Create Sync (which could not be a case if it was filtering duplicates)

PR #488, opened 2019-06-14
nimble/mesh: Fix provisioning errors handling

szymon-czapracki:
Provisioning errors won't cause device to close link.
Upon error, device will send Provisioning Failed PDU, and any further
PDU's will be considered as unexpected. (Mesh Profile section 5.4.4)

Also, a timer starts every time device sends a correct PDU.
This affects MESH/NODE/PROV/BV-10-C test case.

Mismatch in Public Key type will cause device to send
Invalid Format error, and treat any further PDU's as unexpected.

This affects MESH/NODE/PROV/BI-03-C test case.

### DIFF
--- a/nimble/host/mesh/src/prov.c
+++ b/nimble/host/mesh/src/prov.c
@@ -104,6 +104,7 @@ enum {
 	SEND_CONFIRM,          /* Waiting to send Confirm value */
 	WAIT_NUMBER,           /* Waiting for number input from user */
 	WAIT_STRING,           /* Waiting for string input from user */
+	LINK_INVALID,          /* Error occured during provisioning */
 
 	NUM_FLAGS,
 };
@@ -166,6 +167,7 @@ struct prov_rx {
 #define RETRANSMIT_TIMEOUT   K_MSEC(500)
 #define BUF_TIMEOUT          K_MSEC(400)
 #define TRANSACTION_TIMEOUT  K_SECONDS(30)
+#define PROTOCOL_TIMEOUT     K_SECONDS(60)
 
 #if (MYNEWT_VAL(BLE_MESH_PB_GATT))
 #define PROV_BUF_HEADROOM 5
@@ -177,7 +179,7 @@ static struct os_mbuf *rx_buf;
 #define PROV_BUF(len) NET_BUF_SIMPLE(PROV_BUF_HEADROOM + len)
 
 static struct prov_link link;
-
+static struct k_delayed_work prot_timer;
 static const struct bt_mesh_prov *prov;
 
 static void close_link(u8_t err, u8_t reason);
@@ -228,6 +230,8 @@ static void prov_clear_tx(void)
 static void reset_link(void)
 {
 	atomic_clear_bit(link.flags, LINK_ACTIVE);
+	atomic_clear_bit(link.flags, LINK_INVALID);
+	k_delayed_work_cancel(&prot_timer);
 
 	prov_clear_tx();
 
@@ -462,6 +466,8 @@ static int prov_send_gatt(struct os_mbuf *msg)
 
 static inline int prov_send(struct os_mbuf *buf)
 {
+    k_delayed_work_submit(&prot_timer, PROTOCOL_TIMEOUT);
+
 #if (MYNEWT_VAL(BLE_MESH_PB_GATT))
 	if (link.conn_handle != BLE_HS_CONN_HANDLE_NONE) {
 		return prov_send_gatt(buf);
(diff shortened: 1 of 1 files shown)


sjanc:
@jhedberg FYI

jhedberg:
Is this causing some qualification test case failures?

sjanc:
@jhedberg yes
@szymon-czapracki  please update commit message with affected TS name

nimble/host/mesh/src/prov.c
```diff
@@ -177,7 +179,7 @@ static struct os_mbuf *rx_buf;
 #define PROV_BUF(len) NET_BUF_SIMPLE(PROV_BUF_HEADROOM + len)
 
 static struct prov_link link;
-
+static struct os_callout protocol_timer;
```
andrzej-kaczmarek: you cannot use `os_callout` in nimble code - use `ble_npl_callout` instead

nimble/host/mesh/src/prov.c
```diff
@@ -166,6 +167,7 @@ struct prov_rx {
 #define RETRANSMIT_TIMEOUT   K_MSEC(500)
 #define BUF_TIMEOUT          K_MSEC(400)
 #define TRANSACTION_TIMEOUT  K_SECONDS(30)
+#define PROTOCOL_TIMER       os_cputime_usecs_to_ticks(60000000)
```
andrzej-kaczmarek: `ble_npl_time_ms_to_ticks32`
and not really sure how this could even work since cputime time is 1/32768s while os tick (which callouts use) is 1/128s...

nimble/host/mesh/src/prov.c
szymon-czapracki: Done

nimble/host/mesh/src/prov.c
szymon-czapracki: Done

nimble/host/mesh/src/prov.c
szymon-czapracki: Done

nimble/host/mesh/src/prov.c
szymon-czapracki: Done

nimble/host/mesh/src/prov.c
```diff
@@ -727,6 +727,14 @@ static void prov_start(const u8_t *data)
 		return;
 	}
 
+	/* Public Key value is hardcoded in prov_invite() */
+	if (data[1] != 0x00) {
```
rymanluk: so maybe we should have define for this used here and in prov_invite()

szymon-czapracki: Done

szymon-czapracki: Done

jhedberg:
Looks good to me as well, I'll proceed with backporting these to Zephyr.

nimble/host/mesh/src/prov.c
```diff
@@ -1195,6 +1221,12 @@ static void prov_retransmit(struct ble_npl_event *work)
 	}
 }
 
+static void prot_tmo_link_close(struct ble_npl_event *ev)
+{
+    close_link(PROV_ERR_UNEXP_ERR, CLOSE_REASON_FAILED);
+    reset_link();
```
jhedberg: `reset_link()` is a PB-ADV specific internal API, so calling it from this function doesn't seem correct. Also, passing a non-zero first parameter to `close_link()` doesn't seem right since that would cause a second (duplicate) failure message to be sent. It would also be in violation of the spec text " In the case of protocol timeout, the device shall not send a Provisioning Failed PDU.".

jhedberg: Actually it seems Zephyr and nimble are out of sync with each other regarding the `reset_link()` function's purpose. There's `reset_link()` and `reset_state()` in Zephyr. I'm renaming the former to `reset_adv_link()`.

szymon-czapracki: Is it better to use bearer_ctl_send(LINK_CLOSE, NULL, 0) here? 
And after that clear the LINK_INVALID bit with: atomic_clear_bit(link.flags, LINK_INVALID)?

jhedberg: That's essentially what I've done (in my yet to be pushed backporting work). The link_close() function went away completely since the timeout was the only remaining user of it. Note that the link close message has a reason parameter and you should use CLOSE_REASON_TIMEOUT (0x02) for it.

jhedberg: Also, I've put the clear_bit(LINK_INVALID) inside the prov_send_fail_msg() function since it was always cleared right after it anyway.

szymon-czapracki: So you've put clear_bit(LINK_INVALID) just after bearer_ctl_send() and another in prov_send_fail_msg() right? Also, I think that CLOSE_REASON_TIMEOUT has 0x01 is that correct?

jhedberg: Sorry, I meant set_bit(LINK_INVALID) - only place I found was needed was at the end of prof_send_fail_msg(). And yes, you're correct about the timeout error code.

PR #495, opened 2019-06-18
nimble/mesh: Generate new public key on each provisioning session

michal-narajowski:
Mesh Profile Spec v1.0.1 | Section 5.4.2.3:
`If the public key was not available using an OOB technology,
then the public keys are exchanged between the Provisioner
and the unprovisioned device. For each exchange, a new key pair shall be
generated by the Provisioner and the unprovisioned device.`

This allows passing MESH/NODE/PROV/BV-12-C.

### DIFF
--- a/nimble/host/mesh/src/prov.c
+++ b/nimble/host/mesh/src/prov.c
@@ -230,9 +230,17 @@ static void prov_clear_tx(void)
 	free_segments();
 }
 
+static void pub_key_ready(const u8_t *pkey);
+
 static void reset_link(void)
 {
+	static struct bt_pub_key_cb pub_key_cb = {
+		.func = pub_key_ready,
+	};
+	int err;
+
 	atomic_clear_bit(link.flags, LINK_ACTIVE);
+	atomic_clear_bit(link.flags, LOCAL_PUB_KEY);
 	atomic_clear_bit(link.flags, LINK_INVALID);
 	k_delayed_work_cancel(&prot_timer);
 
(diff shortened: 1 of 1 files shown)


rymanluk:
@jhedberg, mentioned testcase is enabled not only for Mesh v1.0.1 but also for Erratum 10395 which is Mandatory for Mesh v1.0

michal-narajowski:
Fixed a conflict.

michal-narajowski:
@jhedberg FYI

nimble/host/mesh/src/prov.c
```diff
@@ -249,8 +257,10 @@ static void reset_link(void)
 #endif
 	link.rx.prev_id = XACT_NVAL;
 
-	if (bt_pub_key_get()) {
-		atomic_set_bit(link.flags, LOCAL_PUB_KEY);
+	err = bt_pub_key_gen(&pub_key_cb);
```
jhedberg: At least in Zephyr `bt_pub_key_gen()` stores the callback persistently in a linked list. As long as we call this from `init()` there's no problem (it's only called once) but when you call it from `reset()` you'll end up with a corrupted linked list. I'm currently looking at how/if we should change our `bt_pub_key_gen()` implementation to clear the list when generation is complete (i.e. the callbacks wouldn't be persistently stored)

michal-narajowski:
This PR has been superseded by https://github.com/apache/mynewt-nimble/pull/506

PR #496, opened 2019-06-18
nimble/gap: Fix for checking sc only mode

rymanluk:
When host is configure for SC only mode, make sure to reject pairing
when remote wants Legacy pairing.

This fixes GAP/SEC/SEM/BV-23-C and GAP/SEC/SEM/BV-24-C

### DIFF
--- a/nimble/host/src/ble_sm.c
+++ b/nimble/host/src/ble_sm.c
@@ -1723,6 +1723,21 @@ err:
     }
 }
 
+static bool
+ble_sm_verify_auth_requirements(uint8_t authreq)
+{
+    /* For now we check only SC only mode. I.e.: when remote indicates
+     * to not support SC pairing, let us make sure legacy pairing is supported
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_sm.c
```diff
@@ -1723,6 +1723,20 @@ ble_sm_pair_exec(struct ble_sm_proc *proc, struct ble_sm_result *res,
     }
 }
 
+static bool
+ble_sm_verify_auth_requirements(uint8_t authreq)
+{
+    /* For now we check only SC only mode. I.e.: when remote request
+     * for legacy pairing, make sure legacy pairing is supported.
```
andrzej-kaczmarek: I'd reword this since remote cannot *request* legacy pairing, it just indicates whether it support LE SC - if it does not support LE SC, we need to support legacy pairing and this is what we check here.

PR #499, opened 2019-06-25
nimble/mesh: Reject invalid remote public key before sending ours

michal-narajowski:
Mesh Profile Spec v1.0.1 | Section 5.4.2.3:
`The Provisioner and the device shall check whether the public key
provided by the peer device or obtained OOB is valid (see Section 5.4.3.1).

When the Provisioner receives an invalid public key, then provisioning fails,
and the Provisioner shall act as described in Section 5.4.4. When the device
receives an invalid public key, then provisioning fails, and
the device shall act as described in Section 5.4.4.`

Described also in Erratum 10395 which is Mandatory for Mesh v1.0.

Fixes MESH/NODE/PROV/BI-13-C.

This test verifies that we do not send our public key if remote's
public key is invalid.

### DIFF
--- a/nimble/host/mesh/src/prov.c
+++ b/nimble/host/mesh/src/prov.c
@@ -898,17 +898,7 @@ static void send_pub_key(void)
 		goto done;
 	}
 
-	prov_buf_init(buf, PROV_PUB_KEY);
-
-	/* Swap X and Y halves independently to big-endian */
-	sys_memcpy_swap(net_buf_simple_add(buf, 32), key, 32);
-	sys_memcpy_swap(net_buf_simple_add(buf, 32), &key[32], 32);
-
-	memcpy(&link.conf_inputs[81], &buf->om_data[1], 64);
-
-	BT_DBG("Local Public Key: %s", bt_hex(&buf->om_data[1], 64));
-
-	prov_send(buf);
+	/* bt_dh_key_gen() will verify that the remote's public key is valid. */
 
 	/* Copy remote key in little-endian for bt_dh_key_gen().
 	 * X and Y halves are swapped independently.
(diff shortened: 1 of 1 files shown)


michal-narajowski:
@jhedberg FYI

michal-narajowski:
I force-pushed a simplified version of this patch.

nimble/host/mesh/src/prov.c
```diff
@@ -898,24 +898,14 @@ static void send_pub_key(void)
...
+	/* bt_dh_key_gen() will verify that the remote's public key is valid. */
 
 	/* Copy remote key in little-endian for bt_dh_key_gen().
 	 * X and Y halves are swapped independently.
 	 */
 	net_buf_simple_init(buf, 0);
-	sys_memcpy_swap(buf->om_data, &link.conf_inputs[17], 32);
-	sys_memcpy_swap(&buf->om_data[32], &link.conf_inputs[49], 32);
+	sys_memcpy_swap(net_buf_simple_add(buf, 32), &link.conf_inputs[17], 32);
+	sys_memcpy_swap(net_buf_simple_add(buf, 32), &link.conf_inputs[49], 32);
```
jhedberg: Why are you using net_buf_simple_add here? From what I see bt_dh_key_gen() doesn't care about the net_buf_simple internal bookkeeping, rather just gets a pointer to the data. It's only later for the response that the net_buf_simple state needs to be correct, and starting by calling prov_buf_init() should take care of that. In fact, prov_buf_init() does not take care of resetting buf->len so you're ending up with a corrupt (overflowing) buffer as far as I see.

michal-narajowski: I think net_buf_simple_init() does take care of buf->om_len in Mynewt. Just wanted to change that for consistency. I will go back to using buf->om_data directly.

jhedberg: But you're not calling net_buf_simple_init() before prov_buf_init() (after having added the above data to the buffer). Anyway, this code is quite different already in Zephyr so I'll let you sort this out.

michal-narajowski: We are calling net_buf_simple_init() in prov_buf_init(), that's why I used it.

jhedberg: Ok, makes sense. The Zephyr prov_buf_init() doesn't do that (it doesn't need to). Either way it's unnecessary code cycles to modify the buffer internal state as far as I see, since all you need is a place to temporarily store the 64 bytes.

jhedberg:
Btw, just so I'm fully understanding this bug: the issue is not that we reject invalid remote public keys, but that we (before this PR) send our own key in response before the rejection (which I assume the PTS sees as acceptance of the key), right?

nimble/host/mesh/src/prov.c
michal-narajowski: Agreed, I deleted that change.

michal-narajowski:
@jhedberg That is correct, I updated the description so it is more clear.

michal-narajowski:
This PR has been superseded by https://github.com/apache/mynewt-nimble/pull/506.

PR #500, opened 2019-06-25
Add the following sanity checks for Notify/Indicate

mfarag13:
- Check for the connection handle.
- Check for a correct entry for the attribute handle.
- Check for enable state for the provided attribute handle.

With the added checks for the following APIs:
- ble_gattc_notify()
- ble_gattc_notify_custom()
- ble_gattc_indicate()
- ble_gattc_indicate_custom()

Host will send the following errors in gap event BLE_GAP_EVENT_NOTIFY_TX
 - BLE_HS_ENOTCONN : If there is no connection with the provided conn_handle.
 - BLE_HS_ENOENT : If there is no entry for the provided att_handle.
 - BLE_HS_EDISABLED: If the cccd is disabled for the provided att_handle.

### DIFF
--- a/nimble/host/src/ble_gatt_priv.h
+++ b/nimble/host/src/ble_gatt_priv.h
@@ -186,6 +186,13 @@ int ble_gatts_register_svcs(const struct ble_gatt_svc_def *svcs,
 int ble_gatts_clt_cfg_access(uint16_t conn_handle, uint16_t attr_handle,
                              uint8_t op, uint16_t offset, struct os_mbuf **om,
                              void *arg);
+int ble_gatts_check_cccd_enabled(struct ble_hs_conn *conn,
+                                 uint16_t chr_val_handle,
+                                 uint8_t flag);
+int ble_gattc_notify_now(uint16_t conn_handle, uint16_t chr_val_handle,
+                         struct os_mbuf *txom, uint8_t now);
+int ble_gattc_indicate_now(uint16_t conn_handle, uint16_t chr_val_handle,
+                           struct os_mbuf *txom, uint8_t now);
 
 /*** @misc. */
 int ble_gatts_conn_can_alloc(void);
--- a/nimble/host/src/ble_gattc.c
+++ b/nimble/host/src/ble_gattc.c
@@ -4147,15 +4147,41 @@ int
 ble_gattc_notify_custom(uint16_t conn_handle, uint16_t chr_val_handle,
                         struct os_mbuf *txom)
 {
+    return ble_gattc_notify_now(conn_handle, chr_val_handle, txom, 0);
+}
+
+int
+ble_gattc_notify(uint16_t conn_handle, uint16_t chr_val_handle)
+{
+    return ble_gattc_notify_custom(conn_handle, chr_val_handle, NULL);
+}
+
+int
+ble_gattc_notify_now(uint16_t conn_handle, uint16_t chr_val_handle,
+                     struct os_mbuf *txom, uint8_t now)
+{
 #if !MYNEWT_VAL(BLE_GATT_NOTIFY)
     return BLE_HS_ENOTSUP;
 #endif
 
+    struct ble_hs_conn *conn;
     int rc;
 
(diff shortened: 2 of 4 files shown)


ccollins476ad:
I'll let others weigh in, but I don't think the host should check for the presence of a CCCD before sending notifications / indications.

In my mind, there is a distinction between subscribing to notifications vs. sending them. A CCCD is required for subscribing, not for sending.  I don't think the spec prohibits sending notifications for a characteristic without a CCCD.

Also, I didn't totally understand what `now` means in the new function names.  To me, "now" means immediately, but the `now` parameter doesn't affect when the notification gets sent.

mfarag13:
@ccollins476ad I agree with you, subscribing to notification/indication is not related to sending them. but It's more logical to send notification/indication to the only subscribed connections and this is what we do when using ble_gatts_tx_notifications_one_chr(). 

now is to check whether to send it directly without checking if cccd is enabled or not
if now = 1, send directly without checking
if now = 0, check first if cccd is enabled or not, if it's not enabled don't send anything.

andrzej-kaczmarek:
`ble_gatts_tx_notifications_one_chr` is indirectly used by `ble_gatts_chr_updated` which can be used to indicate to stack that value of certain characteristic was updated and it's clearly stated in description that as a result notification/indication will be send to subscribed clients only.
`ble_gattc_notify`/`indicate` on the other hand is just a simple API to send a notification or notification and it's up to application to decide what and if it should be sent.

so I agree with @ccollins476ad.

mfarag13:
@andrzej-kaczmarek based on the changes required in this [PR](https://github.com/apache/mynewt-nimble/pull/420). that is related to ANS Service.
We need a method to send to one connection only which has enabled CCCD. Now we don't have This method
We only can notify to all subscribed connection or we can notify without check.

So, we can Modify this PR to have the missing variant.
(Notify with checking if CCCD is enabled or not)

andrzej-kaczmarek:
you can just track cccd state on each connection (you will need to track enabled categories per-connection anyway)

mfarag13:
> 
> 
> you can just track cccd state on each connection (you will need to track enabled categories per-connection anyway)

no just enabled categories, we need to track cccd state per connection, and services are not part of host stack (It's considered as an application) and cccd state is not exposed to application.
Anyway, Is it acceptable to add new APIs that will check for the state of CCCD per connection?

andrzej-kaczmarek:
I think we can consider new API for querying cccd state, but this is not necessary for your case - there is `BLE_GAP_EVENT_SUBSCRIBE` which anyone can track to get updates on cccd state on any connection and characteristic.

ccollins476ad:
Adding new functions (rather than changing the semantics of existing ones) is definitely more appealing.  That said, nimble has been around for a few years and its API is stable, so I think we should be parsimonious with additions.

I may have different experiences using nimble than you, but I don't think I would find this sanity check particularly useful.  I don't think I have ever accidentally sent a notification from the wrong characteristic.

I would be more open to a compile-time setting which adds this sanity check to the existing API (disabled by default).

mfarag13:
@andrzej-kaczmarek for this service it can be handled through gap listener. But we will leave this up to Application writer to keep track for the subscription. which maybe the case for other services/profiles.

@ccollins476ad why not Adding new API that perform the cccd check before sending notifications/indications and leave the current one as it's.
So The Application writer can have the two options without an need to keep track for the subscription.

The two options can be used in one Applications, why restricting it to two options only.
- Sending to one connection without CCCD check.
- Sending to all connections that have CCCD enabled.

what if we need to send to one connection that have CCCD enabled which is the case in ANS.

ccollins476ad:
@mfarag13 Sorry, I totally misunderstood this PR.  I thought it just checked that the characteristic supported notifications / indications.  I didn't realize it checks the receipient's subscription status.

This seems like a useful feature to me.  You can disregard my earlier complaint.

mfarag13:
@ccollins476ad so how should I proceed? Adding new API or leave the PR as It is?

mfarag13:
Hi @ccollins476ad Any update here?

PR #503, opened 2019-06-27
Queuing of Partial Reading in ble_gattc_read_long

mfarag13:
In This PR, 
ble_gattc_read_long() is modified to queue the partial reading and report it only once when all readings received.
As if user requested to read a long characteristic, he knows that this characteristic is longer than ATT_MTU,
And he can process the partial reading rather he will queue it tell receive the whole reading.
and with This PR, Host will queue the reading and report it only once after finishing the read long procedure.


mfarag13:
@sjanc check this modification for ble_gattc_read_long which we have discussed earlier on slack.

nimble/host/include/host/ble_gatt.h
```diff
@@ -320,7 +320,7 @@ int ble_gattc_read_by_uuid(uint16_t conn_handle, uint16_t start_handle,
  * @param cb_arg                The optional argument to pass to the callback
  *                                  function.
  *
- * @return                      0 on success; nonzero on failure.
+ * @return                      BLE_HS_EDONE on success; nonzero on failure.
```
sjanc: why do you need to change success return value?

nimble/host/src/ble_gattc.c
```diff
@@ -3172,17 +3176,17 @@ ble_gattc_read_long_rx_read_rsp(struct ble_gattc_proc *proc, int status,
 
     data_len = OS_MBUF_PKTLEN(*om);
 
-    attr.handle = proc->read_long.handle;
-    attr.offset = proc->read_long.offset;
-    attr.om = *om;
-
-    /* Report partial payload to application. */
-    rc = ble_gattc_read_long_cb(proc, status, 0, &attr);
+    /* Queue Partial Reading. */
```
sjanc: Since this requires extra memory for caching I'd prefer if this would be configurable via syscfg

andrzej-kaczmarek: I'd just create new API for this, e.g. `ble_gattc_read_long_full` (or some smarter name) which can be a simple wrapper on current API

other than this patch lgtm

nimble/host/include/host/ble_gatt.h
mfarag13: because there is no partial reading , the full reading is returned when returning BLE_HS_EDONE

nimble/host/src/ble_gattc.c
mfarag13: So you suggest to add a syscfg to enable/disable full reading and if we enabled full reading we have ble_gattc_read_long_full instead of ble_gattc_read_long. Is this correct?

nimble/host/include/host/ble_gatt.h
ccollins476ad: IMO...

When possible, functions should return 0 on success.  `BLE_HS_EDONE` is a "necessary evil" for when there are multiple ways to succeed.  In this case, success can mean only one thing, so we should use 0.

nimble/host/src/ble_gattc.c
andrzej-kaczmarek: no, my proposal is not to make syscfg but keep old API as-is and add new one with new behavior

mfarag13: @andrzej-kaczmarek I have added the new API that perform the Full reading and update the PR accordingly.

nimble/host/include/host/ble_gatt.h
mfarag13: @ccollins476ad updated.

mfarag13:
Hi All,
Any Update on this PR?

PR #507, opened 2019-07-03
Use packed structures for HCI commands and events

sjanc:


nimble/controller/src/ble_ll_hci.c
```diff
@@ -1075,54 +999,37 @@ ble_ll_hci_le_cmd_proc(uint8_t *cmdbuf, uint16_t ocf, uint8_t *rsplen,
...
-        }
+        rc = ble_ll_conn_hci_le_set_phy(cmdbuf, len);
         break;
 #endif
 #if MYNEWT_VAL(BLE_LL_DIRECT_TEST_MODE)
-    case BLE_HCI_OCF_LE_ENH_RX_TEST:
-        if (len == BLE_HCI_LE_ENH_RX_TEST_LEN) {
-            rc = ble_ll_dtm_rx_test(cmdbuf, true);
-        }
+    case BLE_HCI_OCF_LE_RX_TEST_V2:
```
rymanluk: oh, you are changing also names in the same time according to v5.1 - that should be mentioned in commit message

PR #508, opened 2019-07-05
npl/riot: strip down the default configuration

haukepetersen:
The intention for the NimBLE RIOT port is to provide a rather simplistic default configuration, which can be extended from inside the RIOT build system. For example, the IP-over-BLE implementation (https://github.com/RIOT-OS/RIOT/pull/11578) re-configures the buffer sizes and count to more suitable values for that specific use-case.

This PR disables the link layers security and privacy features, and optimizes the internal buffer space needed by the l2cap layer.

Building RIOTs default `examples/nimble_gatt` application for the `nrf52dk`, this saves ~6.5k of ROM and ~1k of RAM.


porting/npl/riot/include/syscfg/syscfg.h
```diff
@@ -830,7 +830,8 @@
 #endif
 
 #ifndef MYNEWT_VAL_BLE_L2CAP_MAX_CHANS
-#define MYNEWT_VAL_BLE_L2CAP_MAX_CHANS (3*MYNEWT_VAL_BLE_MAX_CONNECTIONS)
+#define MYNEWT_VAL_BLE_L2CAP_MAX_CHANS ((2 + MYNEWT_VAL_BLE_SM_SC) * \
```
andrzej-kaczmarek: we always need 3 channels per connection (ATT, signalling and SM) - this is mandatory in spec

haukepetersen: But in the current configuration, SM is not compiled in, right? So I figured there is no need to allocate the corresponding L2CAP channel - despite from not being standards compliant. Or is there  anything in the code I missed that depends on all 3 channels being allocated even if the SM is missing?

andrzej-kaczmarek: We simply create all 3 channels in `ble_hs_conn_alloc` and failure to create at least one means connection is not allocated in the host. Even if both LE SC and LE legacy pairing are disabled, SM channel has to be allocated in order to be compliant with spec [1] that is so we can reply on signalling with an error as otherwise remote has no way to know that we do not support pairing.

[1] Core 5.1, Vol 3, Part A,  section 2.1
> At a minimum, the L2CAP Signaling channel (fixed channel 0x0001) or the
> L2CAP LE Signaling channel (fixed channel 0x0005) shall be supported. **If
> fixed channel 0x0005 is supported, then fixed channels 0x0004 and 0x0006
> shall be supported** (see Table 2.2).

andrzej-kaczmarek: And yes, SM is not compiled in when LE SC and LE legacy pairing are both disabled. There's just a simple handler for L2CAP channel to reply with an error on any received signalling.

haukepetersen: ok, I understand. Will fix.

haukepetersen:
removed the invalid l2cap channel optimization as discussed above. So this PR boils down to disabling the link layer security and privacy features.

andrzej-kaczmarek:
travis failed due to configuration issue, will merge anyway so there is no need to rebase it once more

PR #509, opened 2019-07-07
Few controller optimizations and cleanup

andrzej-kaczmarek:
See commits for details.


andrzej-kaczmarek:
@haukepetersen

andrzej-kaczmarek:
replaced `asm` with `__asm__` since RIOT uses `std=c99` and used local labels in inline assembly to fix build errors in optimized code

andrzej-kaczmarek:
forgot to update also for nRF51 :-)

PR #519, opened 2019-07-10
nimble: Add attribute caching protocols and service changed

maziiz:
GATT caching feature baseline:
Add ble_att_caching.c/h to store services, characteristics, and descriptors on the client-side. the basic concept is an optimization that allows the client to discover the Attribute information such as Attribute Handles used by the server once.


maziiz:
> thanks for this PR. For this moment idea was to have caching in the application layer as per example in btshell app. However, I think this is a good idea to add caching to NimBLE however I have couple of suggestions for such a caching
> 
> 1. imho we should not cache based on conn_handle but rather per device address so cache is valid also after reconnect at least for bonded devices.
> 2. Having above in mind maybe we should use ble_store for it if we would like to keep valid cache on the power cycle (?)
> 3. I think we should have separate package for caching as not everybody might want to use it. In the same time I would not introduce new GAP_EVENT for service changed but rather add the logic somewhere else e.g. app or new package which could be registered for the GAP events.
>    Having proposed event here would break old applications.
> 4. I would call new files ble_attr_cache.c/h or ble_gatt_db_cache.c/h
> 
> @andrzej-kaczmarek @sjanc @ccollins476ad any thoughts?

Hello @rymanluk,
Thank you for your comments, please consider:
1-Cache is already valid per different connection as all cached data is in the static array ble_att_caching_conns and when adding new connection check on address, but I noticed a change per different connections we need to update the conn_handle I updated the PR with it.
2-Using ble_store is a very good idea I was thinking of it and I will work on it for future PR.
3-what about making cache feature is a sys_cfg macro and lock all codes with it. and I don't see any breaking for old apps when adding a new GAP_EVENT all old apps will not check for that type of events
4-I already changed them 👍 .

PR #524, opened 2019-07-12
nimble/mesh: Increase transaction ID after sending unack message

michal-narajowski:
UNACK messages are not retransmitted so we should increase TID
after sending each message.

### DIFF
--- a/nimble/host/mesh/src/model_cli.c
+++ b/nimble/host/mesh/src/model_cli.c
@@ -52,8 +52,6 @@ static void gen_onoff_status(struct bt_mesh_model *model,
 
 	BT_DBG("state: %d", state);
 
-	transaction_id++;
-
 	k_sem_give(&cli->op_sync);
 }
 
@@ -83,8 +81,6 @@ static void gen_level_status(struct bt_mesh_model *model,
 	}
 
 	BT_DBG("level: %d", level);
-
-	transaction_id++;
 
 	k_sem_give(&cli->op_sync);
 }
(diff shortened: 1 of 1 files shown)


nimble/host/mesh/src/model_cli.c
```diff
@@ -176,6 +176,8 @@ int bt_mesh_gen_onoff_set(u16_t net_idx, u16_t addr, u16_t app_idx,
 	}
 
 	if (!state) {
+		/* Increase transaction ID after sending set_unack message */
+		transaction_id++;
```
rymanluk: Maybe we have increment it in `done:` label instead here and in status callback?

michal-narajowski: Done, thanks!

nimble/host/mesh/src/model_cli.c
```diff
@@ -181,6 +177,7 @@ int bt_mesh_gen_onoff_set(u16_t net_idx, u16_t addr, u16_t app_idx,
 
 	err = cli_wait(&gen_onoff_cli, &param, OP_GEN_ONOFF_STATUS);
 done:
+	transaction_id++;
```
rymanluk: do I understand it correctly that err will be non zero when timeout on ACK will happen? If so I'm not sure if we should increase transaction_id

michal-narajowski: Correct, fixed.

PR #525, opened 2019-07-12
nimble/transport: Add option to use custom HCI transport

andrzej-kaczmarek:
Since it's recommended for apps to explicitly include only meta-package
for HCI transport (so it can be easily changed in target for any BSP),
we should also have an option to allow HCI transport other than one of
those we have in tree. This is useful for development and 3rd party
solutions which may have own HCI transport.


andrzej-kaczmarek:
perhaps we could also just remove `$notnull` restriction, but I think it makes more sense to override target like this:
`BLE_HCI_TRANSPORT: custom`
instead of
`BLE_HCI_TRANSPORT:`

PR #527, opened 2019-07-15
nimble/mesh: Fix null pointer access on prov_init

michal-narajowski:
During bt_mesh_prov_init() a bt_mesh_proxy_get_buf() function
is called that returns a buffer that tries to initialize
unallocated buffer. With this patch, we just return NULL
here, and then the buffer is allocated after Proxy is initialized.


michal-narajowski:
This should fix issue https://github.com/apache/mynewt-nimble/pull/527.

PR #534, opened 2019-07-23
nimble/phy: Fix memblock overwrite

andrzej-kaczmarek:
Optimized ble_phy_rxpdu_copy (f19002e) can overwrite header of adjacent
block at specific payload lengths.

This happens if length of data that should be copied to last mbuf in
chain is less than 4 bytes of data. In such case the last but one mbuf
is already full when remaining data are appended to it beyond its data
buffer effectively overwriting adjacent buffer.

We should always check if there is enough space in current mbuf to copy
data. If not, we just advance to next mbuf in chain. Even in such case
there will be free mbuf to store remaining few bytes of data as this is
guaranteed by ble_ll_rxpdu_alloc.


andrzej-kaczmarek:
changed register constraints to fix build error on nrf51

PR #528, opened 2019-07-16
nimble/ll: Workaround and traces for the possible race scenario

rymanluk:
It looks like there is a race when closing connection with different
paths. It can lead to situation that Nimble tries to close connsm which is
already in IDLE, which leads to crash when removing connsm from
g_ble_ll_conn_active_list which does not contain connsm.

This patch adds workaround and plus assert and stats which should help
to understand the root cause

PR #536, opened 2019-07-23
Fixes for 1.2-rc2

sjanc:
Few fixes for upcoming 1.2-rc2


rymanluk:
That looks good. I would consider adding this as well: https://github.com/apache/mynewt-nimble/pull/528

PR #540, opened 2019-07-24
nimble_host: Add support to accept or reject pairing request through GAP event

prasad-alatkar:
- The PR is intended to give chance to application to accept or reject pairing request through GAP event.
- In case of implementing **Mode 1 security level 4**, the connection should be `authenticated le secure connection only`, but there is no way to make sure that the `Mode 1 security level 4` is achieved or not. This PR addresses this issue in particular but not limited to.


nimble/host/include/host/ble_gap.h
```diff
@@ -428,6 +429,17 @@ struct ble_gap_disc_desc {
...
+struct ble_gap_pairing_req {
+    uint16_t conn_handle;
+    /** Properties of the existing pairing request */
+    uint8_t io_cap;
+    uint8_t oob_data_flag;
+    uint8_t authreq;
+    uint8_t max_enc_key_size;
+    uint8_t init_key_dist;
+    uint8_t resp_key_dist;
+};
```
andrzej-kaczmarek: I don't think we should just expose to app anything that is in SMP request... what app could use key distribution settings for?

nimble/host/src/ble_gap.c
```diff
@@ -5309,6 +5309,29 @@ ble_gap_identity_event(uint16_t conn_handle)
...
+    memset(&event, 0, sizeof event);
+    event.type = BLE_GAP_EVENT_PAIRING_REQUEST;
+    event.pairing_req.io_cap = req->io_cap;
+    event.pairing_req.oob_data_flag = req->oob_data_flag;
+    event.pairing_req.authreq = req->authreq;
+    event.pairing_req.max_enc_key_size = req->max_enc_key_size;
+    event.pairing_req.init_key_dist = req->init_key_dist;
+    event.pairing_req.resp_key_dist = req->resp_key_dist;
+
+    rc = ble_gap_call_conn_event_cb(&event, req->conn_handle);
```
andrzej-kaczmarek: this basically blocks nimble until app responds and app may simply decide to ask user whether to pair or not which will block nimble for quite a long time - this is not acceptable.

nimble/host/include/host/ble_gap.h
prasad-alatkar: Totally agree to that.. will make the change !!

nimble/host/src/ble_gap.c
prasad-alatkar: Well IMO, in `ble_gap_event` if there is no mention of `BLE_GAP_EVENT_PAIRING_REQUEST`, the app will simply `return 0` which is similar to accepting the key, so we should not have any issue I guess. Please let me know if you find this wrong.

nimble/host/include/host/ble_gap.h
prasad-alatkar: Done !!

nimble/host/src/ble_gap.c
prasad-alatkar: Hi @andrzej-kaczmarek , thank you for your valuable review. Just to give you brief background about our requirement: 

- We want the device to be in `Secure Connection Only`, in which all its service requests, except the one in `Mode 1 Level 1`, should only be accessed in `Mode 1 Level 4`. The current NimBLE host implementation doesn't have any APIs or support for this particular requirement . 

- Now with this PR, we intend to give application choice to reject pairing on various parameters (in our particular case `Mode 1 level 2` and `Mode 1 level 3` requests will be rejected when `Secure Connection only` mode is required.).

- Coming back to 

> app may simply decide to ask user whether to pair or not which will block nimble for quite a long time

andrzej-kaczmarek: you want to solve one scenario here, however if you let application to accept just works pairing, the application does not need to decide whether to accept just works pairing on its own - it may display ui for user to accept it and this would block. for this reason it should be handled in similar way as other pairing methods that require user input. for this we probably would need to rework pairing API a bit, not yet sure.

prasad-alatkar: Thank you @andrzej-kaczmarek for your response, I understand your point.
 
We are on 'NimBLE 1.1.0' stable release, but on master branch I observed new API (commit: f8a8da8): `ble_sm_verify_auth_requirements`. I think this API can help us achieve the desired goal with few modifications. The solution I have in mind is: 
- Introduce new API which simply sets a local flag(`ble_sm_sc_only_requested`) in `ble_sm.c`, indicating the connection needs to be in `secure connection only mode`.  
- In `ble_sm_verify_auth_requirements` function check if the `ble_sm_sc_only_requested` is set to `true`, if `true` first check `sm_sc` bit is on for both `pairing requester` and `pairing responder`, then check if the authentication requirements are (i.e. not just works only)

Please let me know if the above method is acceptable to you, please feel free to suggest any alternative method. Just to give you a very brief background on why do we need `secure connection only` mode, we are working on providing NimBLE host based solution to **Amazon FreeRTOS** and they have specific requirement of `secure connection only mode`.

prasad-alatkar: Hi @andrzej-kaczmarek, Could you please suggest any alternative to solve this particular case ?

prasad-alatkar: Hi @andrzej-kaczmarek if the change is not acceptable, can you suggest any alternatives ? I can work in that direction and update this PR ?

andrzej-kaczmarek: there is already API which allows for asynchronous replies for pairing (`ble_sm_inject_io`) and it should be reused here. we had offline discussion about this and reasonable way to do this would be to add two extra actions to inject: `BLE_SM_IOACT_JUSTWORKS` and `BLE_SM_IOACT_REJECT`. the former would be used to accept just works and the latter to reject any pairing (we do not have option to reject other pairing methods now, but if just works can be rejected then all other methods should be also "rejectable"). to ensure compatiblility with existing app, pairing event should assume just works to be automatically accepted if handler returned 0 and wait for inject io only in case of e.g. non-zero return value.

prasad-alatkar: Hi @andrzej-kaczmarek , apologies for not actively working on this PR. If I understand correctly, the application should accept/reject the pairing request using API `ble_sm_inject_io`, I have made changes to this PR according to this assumptions. Please take a look.

prasad-alatkar: Hi @andrzej-kaczmarek , I have given another shot to this long pending PR. By introducing extra actions to inject (`BLE_SM_IOACT_JUSTWORKS` for accept & `BLE_SM_IOACT_REJECT` for reject), now application can ask for user input using `ble_sm_inject_io`, if pairing is accepted the PAIRING event handler should put `BLE_SM_IOACT_JUSTWORKS` in action & if rejected then put `BLE_SM_IOACT_REJECT`. However It may just seem a bit odd in case of UI has display + keyboard functionality and accepting pairing is to put `BLE_SM_IOACT_JUSTWORKS` in action of `ble_sm_io`. Please let me know your thoughts on this.

rymanluk: @prasad-alatkar I think this PR will not work. I can't see where pairing is paused when waiting for user inject action.
Please have a look on how  BLE_GAP_EVENT_PASSKEY_ACTION works, as what you are trying  to do is something similar.

Maybe as a starting point you could add new actions as a part of that event?

rymanluk: Hi @prasad-alatkar just wondering if you are planning to work on that.

prasad-alatkar: Yes @rymanluk, I haven't been able to allocate good amount of time for this PR, will work on this early next week.

rymanluk: Thanks @prasad-alatkar

prasad-alatkar:
@rymanluk Apologies for not updating the PR for long time. I have incorporated your suggestions. Please take a look and suggest if any improvements. I am expecting few CI test failures given the nature of PR, please provide inputs on modifying tests as well. Thanks

prasad-alatkar:
@andrzej-kaczmarek @rymanluk Could you guys please take a look ?

prasad-alatkar:
@rymanluk Can you please take a look at the PR ?

rymanluk:
@prasad-alatkar  looks like unit tests are failing. newt test @apache-mynewt-nimble/nimble/host/test  could you check that? Also could you please add a patch to btshell which would allow to test this patch? Thanks

prasad-alatkar:
> @prasad-alatkar looks like unit tests are failing. newt test @apache-mynewt-nimble/nimble/host/test could you check that? Also could you please add a patch to btshell which would allow to test this patch? Thanks

@rymanluk , as per my understanding the failure is seen in unit test : `ble_sm_test_case_peer_lgcy_fail_confirm`. However, I am a little confused with the error prints here: `nimble_host_test.elf: nimble/host/src/ble_hs.c:194: ble_hs_lock: Assertion `!ble_hs_locked_by_cur_task()' failed`. It seems that `ble_hs_lock` is getting called twice somewhere, but the PR changes do not add any extra host lock. Could you please help here to pass the CI tests ? I am not very familiar with BT tests here, could you please let me know if it is possible to collect debug logs through CI ?

PR #547, opened 2019-07-31
Mynewt: Define logs in packages' syscfg files

ccollins476ad:
This is a Mynewt-specific PR.

Defining logs in syscfg files provides a few benefits:

1. Ability to set log level per module at build time.  Setting a global log level of 0 (`LOG_LEVEL=0`) might produce an image that is too big.  Instead, we can enable debug logging only for the modules we are interested in.

2. Easier to visualize system log configuration (see below).

3. Log modules can be remapped at build time in case of conflicts.

To see the system-wide log configuration, use either of these commands:

```
newt target logcfg brief <target-name>
newt target logcfg show <target-name>
```

### DIFF
--- a/nimble/host/include/host/ble_hs_log.h
+++ b/nimble/host/include/host/ble_hs_log.h
@@ -21,6 +21,9 @@
 #define H_BLE_HS_LOG_
 
 #include "modlog/modlog.h"
+#if MYNEWT
+#include "os/mynewt.h"
+#endif
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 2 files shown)


ccollins476ad:
@sjanc Sounds good.  I wonder if it would be difficult to move logging to NPL and still allow newt to generate the Mynewt logging macros.  If you think this PR is not the right direction, I am happy to close it so that it can be re-thought.

ccollins476ad:
@sjanc Is there something else I can do with this?  Do you want me to close it?

PR #550, opened 2019-08-05
nimble/ll: Indicate proper transition from AUX_CONNECT_REQ

andrzej-kaczmarek:


andrzej-kaczmarek:
added explanation in commit message

PR #555, opened 2019-08-13
nimble/mesh: configurable segment retransmist attempts

hardik-proxy:
Current SEG_RETRANSMIT_ATTEMPTS value is defined as 4. And hence, while sending non-ack segmented messages, each segment is being re-transmitted 4 times. When it comes, to the timing I observed that a message with three segments takes at least 8-15 seconds between successive re-transmissions. For reliability, if the whole message is being transmitted multiple times, I believe segment re-transmissions becomes redundant (please correct me if I am wrong) and thus SEG_RETRANSMIT_ATTEMPTS can be set to lower value to reduce the time between successive re-transmissions.

As corrected by @rymanluk minimum suggested by the spec is 2.


nimble/host/mesh/syscfg.yml
```diff
@@ -191,6 +191,12 @@ syscfg.defs:
             as there are outgoing segments.
         value: 3
 
+    BLE_MESH_SEG_RETRANSMIT_ATTEMPTS:
+        description: >
+            Number of retransmit attempts (after the initial transmit) per segment
```
rymanluk: Please put here specification reference we discussed, so it is clear where restriction came from.

porting/examples/linux_blemesh/include/syscfg/syscfg.h
```diff
@@ -1021,6 +1021,10 @@
 #define MYNEWT_VAL_BLE_MESH_TX_SEG_MSG_COUNT (4)
 #endif
 
+#ifndef MYNEWT_VAL_BLE_MESH_TX_SEG_RETRANSMIT_ATTEMPTS
```
rymanluk: linux_blemesh does not build because define is incorrect here. There is no "TX" in the syscfg definition.

nimble/host/mesh/syscfg.yml
hardik-proxy: @rymanluk Sure I will put the specification reference. Also, the restriction doesn't work if I implement the way I have done. I have tried $notnull but I didn't get any error message. Could you please let me know the better way to implement the restrictions. Thank you!

porting/examples/linux_blemesh/include/syscfg/syscfg.h
hardik-proxy: Updated. Please review.

PR #559, opened 2019-08-21
nimble/host: Cleanup address type conversion

michal-narajowski:
We should consistently store sec/cccds using public/random address types.
To do this we introduce dedicated conversion functions for own and
peer address types. Even though they use the same values it makes
code easier to follow.

### DIFF
--- a/nimble/host/src/ble_gatts.c
+++ b/nimble/host/src/ble_gatts.c
@@ -718,6 +718,8 @@ ble_gatts_clt_cfg_access_locked(struct ble_hs_conn *conn, uint16_t attr_handle,
             /* Successful writes get persisted for bonded connections. */
             if (conn->bhc_sec_state.bonded) {
                 out_cccd->peer_addr = conn->bhc_peer_addr;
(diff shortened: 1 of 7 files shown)


nimble/host/src/ble_gatts.c
```diff
@@ -718,6 +718,8 @@ ble_gatts_clt_cfg_access_locked(struct ble_hs_conn *conn, uint16_t attr_handle,
             /* Successful writes get persisted for bonded connections. */
             if (conn->bhc_sec_state.bonded) {
                 out_cccd->peer_addr = conn->bhc_peer_addr;
+                out_cccd->peer_addr.type =
+                    ble_hs_misc_addr_type_to_id(conn->bhc_peer_addr.type);
```
andrzej-kaczmarek: `ble_hs_misc_addr_type_to_id` is for own addr type

michal-narajowski: @andrzej-kaczmarek thanks, fixed.

PR #562, opened 2019-08-22
nimble/sm: Add support for debug mode

rymanluk:
When MYNEWT_VAL BLE_SM_SC_DEBUG_MODE is set then Secure Manager enters into
Debug Mode. Public and private keys are fix as per Bluetooth 5.1
specification.

Note this is only available for Secure Connections.


andrzej-kaczmarek:
isn't this the same as `BLE_SM_SC_DEBUG_KEYS`?

rymanluk:
heh, so there is another place with debug keys..did not found it, but maybe we should have one place where debug keys are and not two.

andrzej-kaczmarek:
debug keys enabled with 'BLE_SM_SC_DEBUG_KEYS` are what spec says
the other debug keys have corresponding API to set any keys for debugging purposes

rymanluk:
don't need it as functionality is there already

PR #565, opened 2019-08-27
nimble/ll: Refactor PHY transitions

andrzej-kaczmarek:
This changes semantics of ble_phy_mode_set(). Instead of only setting
current phy_mode and TX-to-RX phy_mode, it will set an actual phy_modes
used on every TX and RX during packet sequence. This allows ble_phy to
configure RF properly for both transitions and also makes it resonsible
for changing phy_mode on each transition - ble_ll shall now only call
ble_phy_mode_set before 1st packet in an event, i.e. before calling
ble_phy_tx/rx_set_start_time.

### DIFF
--- a/nimble/controller/src/ble_ll_conn.c
+++ b/nimble/controller/src/ble_ll_conn.c
@@ -1110,10 +1110,6 @@ conn_tx_pdu:
     }
 #endif
 
-#if (BLE_LL_BT5_PHY_SUPPORTED == 1)
-    ble_phy_mode_set(connsm->phy_data.tx_phy_mode,connsm->phy_data.rx_phy_mode);
-#endif
-
     /* Set transmit end callback */
     ble_phy_set_txend_cb(txend_func, connsm);
     rc = ble_phy_tx(ble_ll_tx_mbuf_pducb, m, end_transition);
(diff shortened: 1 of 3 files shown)


nimble/drivers/nrf52/src/ble_phy.c
```diff
@@ -1017,6 +1016,10 @@ ble_phy_rx_end_isr(void)
 #endif
     }
 
+#if (BLE_LL_BT5_PHY_SUPPORTED == 1)
+        ble_phy_mode_apply(g_ble_phy_data.phy_tx_phy_mode);
```
rymanluk: indentation

nimble/controller/src/ble_ll_conn.c
```diff
@@ -1232,8 +1228,8 @@ ble_ll_conn_event_start_cb(struct ble_ll_sched_item *sch)
 #endif
 
 #if (BLE_LL_BT5_PHY_SUPPORTED == 1)
-        ble_phy_mode_set(connsm->phy_data.rx_phy_mode,
-                             connsm->phy_data.rx_phy_mode);
+        ble_phy_mode_set(connsm->phy_data.tx_phy_mode,
```
rymanluk: if I understand it correctly, this call should be done somewhere around line 1191 otherwise phy is not set before TX as a master because this `ble_phy_mode_set()` has been removed from `ble_ll_conn_tx_data_pdu()`
But maybe I miss something - did you test it?

andrzej-kaczmarek:
thanks @rymanluk - fixed (I tested only with bleprph so this is why it worked fine for me...)

PR #569, opened 2019-08-30
[DNM] nimble: Workaround for controllers sending incorrect events order

rymanluk:
Some controllers incorrectly sends LE Advertising Set Terminate Event before
LE Connection Complete Event when connection is established

This patch handles this bad behavior

NOTE: Not sure if we want it as I'm not sure how many controllers has this issue. I saw one but it is not on the market - however posting this PR here in case somebody will face it.


rymanluk:
well I hope there is a test for it  and I posted it here just in case anybody would face it - so the workaround is here.
What worries me here is that if this happens, Host and Controller are out of sync i.e Controller has the connection and host is not aware of it. Most probably link will be dropped 30 sec after first ATT request, but still this is not nice. I close this PR  but have it in mind :)

PR #572, opened 2019-09-05
nimble/gatt: Fix null pointer dereference

mfarag13:
Modify ble_gattc_proc_matches_conn_rx_entry to return 0 in case no matching entry found to avoid accessing Null Pointer.

### DIFF
--- a/nimble/host/src/ble_gattc.c
+++ b/nimble/host/src/ble_gattc.c
@@ -912,7 +912,7 @@ ble_gattc_proc_matches_conn_rx_entry(struct ble_gattc_proc *proc, void *arg)
     criteria->matching_rx_entry = ble_gattc_rx_entry_find(
         proc->op, criteria->rx_entries, criteria->num_rx_entries);
 
-    return 1;
+    return (criteria->matching_rx_entry != NULL);
 }
 
 static void


nimble/host/src/ble_gattc.c
```diff
@@ -912,7 +912,11 @@ ble_gattc_proc_matches_conn_rx_entry(struct ble_gattc_proc *proc, void *arg)
     criteria->matching_rx_entry = ble_gattc_rx_entry_find(
         proc->op, criteria->rx_entries, criteria->num_rx_entries);
 
-    return 1;
+    if (criteria->matching_rx_entry) {
+        return 1;
+    } else {
```
rymanluk: Else is not needed here. Otherwise it looks good to me

nimble/host/src/ble_gattc.c
```diff
@@ -912,7 +912,14 @@ ble_gattc_proc_matches_conn_rx_entry(struct ble_gattc_proc *proc, void *arg)
     criteria->matching_rx_entry = ble_gattc_rx_entry_find(
         proc->op, criteria->rx_entries, criteria->num_rx_entries);
 
-    return 1;
+    if (criteria->matching_rx_entry) {
+        /* Return 1 in case we find an entry only
```
rymanluk: this comment is not needed.
And actually return (criteria->matching_rx_entry != NULL) would be enough.

nimble/host/src/ble_gattc.c
```diff
@@ -912,7 +912,14 @@ ble_gattc_proc_matches_conn_rx_entry(struct ble_gattc_proc *proc, void *arg)
     criteria->matching_rx_entry = ble_gattc_rx_entry_find(
         proc->op, criteria->rx_entries, criteria->num_rx_entries);
 
-    return 1;
+    if (criteria->matching_rx_entry) {
```
michal-narajowski: I would just do `return criteria->matching_rx_entry;`

michal-narajowski:
@mfarag13 I updated the PR because we need to merge this fix but you are still the author.

PR #574, opened 2019-09-09
nimble/ll: Randomize initial advertising event

sjanc:
When starting advertising we can also randomize first event. This is
primary useful for mesh relay where multiple devices starts advertising
at the same time.

This is a bit of RFC :)

Also I don't think we need 5ms fixed delay anymore but I've left it as is for now...

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -2613,7 +2613,7 @@ ble_ll_adv_sm_start(struct ble_ll_adv_sm *advsm)
     if (xtal_state != BLE_RFCLK_STATE_SETTLED) {
         if (xtal_state == BLE_RFCLK_STATE_OFF) {
             advsm->adv_pdu_start_time += g_ble_ll_data.ll_xtal_ticks;
-            ble_ll_xcvr_rfclk_start_now(os_cputime_get32());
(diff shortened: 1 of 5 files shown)


sjanc:
@haukepetersen  FYI

nimble/controller/src/ble_ll_adv.c
```diff
@@ -2596,16 +2597,18 @@ ble_ll_adv_sm_start(struct ble_ll_adv_sm *advsm)
...
-     * to check if we are currently doing anything. Just makes this simple.
+     * advertising event some time in the future (5 msecs + adv rand). This will
+     * give time to start up any clocks or anything and also avoid a bunch of
+     * code to check if we are currently doing anything. Just makes this simple.
      *
      * Might also want to align this on a slot in the future.
      *
      * NOTE: adv_event_start_time gets set by the sched_adv_new
      */
+    start_delay_us = 5000 + (rand() % (BLE_LL_ADV_DELAY_MS_MAX * 1000));
```
rymanluk: maybe instead `rand()` we could use `ble_ll_rand_data_get()`?

andrzej-kaczmarek: we need random number, not random data...

haukepetersen:
Nice, looks good to me!

PR #578, opened 2019-09-18
nimble: Add pre-enable and post-stop callbacks

rymanluk:
@dhrishi do you think it might help somehow in the context of #467 ?


dhrishi:
Hi @rymanluk Sorry for missing this. Can you please let me know how these callbacks will help wrt stop/deinit ports? That PR cleans up the host and port specific things in a generic way.

rymanluk:
@sjanc @andrzej-kaczmarek I think we could still have it

PR #579, opened 2019-09-19
nimble/host: Fix ble_gap_adv_set_fields for zero length

dhrishi:
Replace assert with appropriate error so that the caller of
ble_gap_adv_set_fields() gets to know the error

### DIFF
--- a/nimble/host/src/ble_hs_adv.c
+++ b/nimble/host/src/ble_hs_adv.c
@@ -356,7 +356,7 @@ adv_set_fields(const struct ble_hs_adv_fields *adv_fields,
     }
 
     /*** 0x16 - Service data - 16-bit UUID. */
-    if (adv_fields->svc_data_uuid16 != NULL) {
+    if (adv_fields->svc_data_uuid16 != NULL && adv_fields->svc_data_uuid16_len) {
         rc = ble_hs_adv_set_flat_mbuf(BLE_HS_ADV_TYPE_SVC_DATA_UUID16,
                                       adv_fields->svc_data_uuid16_len,
                                       adv_fields->svc_data_uuid16, dst, dst_len,
@@ -402,7 +402,7 @@ adv_set_fields(const struct ble_hs_adv_fields *adv_fields,
     }
 
     /*** 0x20 - Service data - 32-bit UUID. */
-    if (adv_fields->svc_data_uuid32 != NULL) {
+    if (adv_fields->svc_data_uuid32 != NULL && adv_fields->svc_data_uuid32_len) {
         rc = ble_hs_adv_set_flat_mbuf(BLE_HS_ADV_TYPE_SVC_DATA_UUID32,
                                      adv_fields->svc_data_uuid32_len,
                                      adv_fields->svc_data_uuid32, dst, dst_len,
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_hs_adv.c
```diff
@@ -68,7 +68,9 @@ ble_hs_adv_set_flat_mbuf(uint8_t type, int data_len, const void *data,
 {
     int rc;
 
-    BLE_HS_DBG_ASSERT(data_len > 0);
+    if (data_len <= 0) {
+        return BLE_HS_EMSGSIZE;
+    }
```
andrzej-kaczmarek: handling error here makes sense here, but error code should be different (invalid parameters) and it should also be returned if `data` is `NULL`

however, I would keep a debug assert here since `data_len<=0` means application provided invalid parameters and an assert makes it easier to catch it during development. in theory you could set field with 0 length but then it also would make this check invalid. in practice field with length of 0 does not make sense.

dhrishi: handling error here makes sense here, but error code should be different (invalid parameters) and it should also be returned if data is NULL

> The check for NULL data is done even before the control reaches here (done in `adv_set_fields`). When it reaches here it is like non NULL data with zero length. So, I thought the appropriate error code should be returned to the application. Let  me know what you think about this so that I can make the changes (changing errror code) accordingly

andrzej-kaczmarek: ok, so now I see this is declared in `ble_hs_adv_priv.h` so it means some other code calls this with length=0 which means the error is in other place - we do not need to check length parameter here and debug assert is ok.

so the question in what exactly causes you to hit this assert?

dhrishi: @andrzej-kaczmarek Yes, this is not a public API. This is internally called from `ble_hs_adv_set_fields()`. Also, if the error is handled, it is correctly propagated to the application.

> so the question in what exactly causes you to hit this assert?

So, an user was setting a advertisement TLV with valid value and 0 length (by mistake) using `ble_hs_adv_set_fields()` and hence came to this.

andrzej-kaczmarek: so imo `ble_hs_adv_set_fields` should simply not call `ble_hs_adv_set_flat_mbuf` in such case (but I still do not know what exactly caused it, i.e. which contents of fields struct causes it)

dhrishi: @andrzej-kaczmarek Yes, the check can be done here (in additon to NULL buffer): https://github.com/apache/mynewt-nimble/blob/90168812576b99c9e4315690062bdf46548409a4/nimble/host/src/ble_hs_adv.c#L437

Also, similar length checks would be required for all the adv types as well. Let me know if the fix here makes more sense.

dhrishi: @andrzej-kaczmarek  Ok. Will add length checks for manufacturer data (>=2) and other types accordingly in `adv_set_fields`.

dhrishi: @andrzej-kaczmarek Done!

dhrishi:
@andrzej-kaczmarek Can you please review the changes made as per our discussion.

PR #583, opened 2019-09-24
NimBLE/host: Fix multiple `discard const qualifier` warnings

prasad-alatkar:
Minor changes in `porting/nimble/` and `ble_gattc.c` to remove `discard const
  qualifier` warnings.

### DIFF
--- a/nimble/host/src/ble_gattc.c
+++ b/nimble/host/src/ble_gattc.c
@@ -485,7 +485,7 @@ ble_gattc_dbg_assert_proc_not_inserted(struct ble_gattc_proc *proc)
  *****************************************************************************/
 
 static void
-ble_gattc_log_proc_init(char *name)
+ble_gattc_log_proc_init(const char *name)
 {
     BLE_HS_LOG(INFO, "GATT procedure initiated: %s", name);
 }
@@ -587,7 +587,7 @@ ble_gattc_log_read_mult(const uint16_t *handles, uint8_t num_handles)
 static void
 ble_gattc_log_write(uint16_t att_handle, uint16_t len, int expecting_rsp)
 {
-    char *name;
+    const char *name;
 
     if (expecting_rsp) {
         name = "write; ";


andrzej-kaczmarek:
can you tell us how do you build NimBLE and give some examples of these warnings?

prasad-alatkar:
These warnings are thrown with `make` build system, also when compiler option `Enable -Wwrite-strings warning flag` are enabled. Below are the actual warnings for reference:
```
esp-idf/components/bt/host/nimble/nimble/nimble/host/src/ble_gatts.c:1249:26: **warning:** passing argument 5 of 'os_mempool_init' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
                          "ble_gatts_clt_cfg_pool");
```
and 
```
esp-idf/components/bt/host/nimble/nimble/porting/nimble/include/os/os_mempool.h:173:12: **note:** expected 'char *' but argument is of type 'const char *'
 os_error_t os_mempool_init(struct os_mempool *mp, uint16_t blocks,
            ^
```

I understand we can always ignore these warnings by adding CFLAG ` -Wno-discarded-qualifiers flag`.

andrzej-kaczmarek:
these changes are valid, however the problem is that it makes our local copy of Mynewt files out of sync. we would need to make these changes in Mynewt first and then update here and this will probably not be trivial due to many dependencies in Mynewt code itself. for now we can merge only changes in NimBLE code.

prasad-alatkar:
Hi @andrzej-kaczmarek , I think I misread your reply. By 

> for now we can merge only changes in NimBLE code.
Do you mean changes in `porting/` won't be merged but ? If this is the case I can modify my PR accordingly.

andrzej-kaczmarek:
@prasad-alatkar what I mean is that I will merge this PR if you update it to modify only NimBLE code. changes in `porting/nimble` have to be done in mynewt-core first and then we can sync them here. as I said, changing mynewt-core won't probably be an easy task since this will likely cause a lot of build errors due to dependency and we probably should fix all similar usages at once, not just single instance, so this may be quite big task.

and I am not sure why you try to build NimBLE with extra warnings enabled? we simply cannot make it build with any flags one may use so it should be build with the same flags as in examples and Mynewt since this is "officially" supported configuration.

prasad-alatkar:
Sure @andrzej-kaczmarek . I will make the changes to PR, however feel free to close the PR if you find these changes unnecessary.

PR #585, opened 2019-09-25
tests/bledtm: Add test app for DTM TX

andrzej-kaczmarek:
This adds simple test app to run DTM TX with given parameters.
Parameters are set via syscfg so simply configure, build, flash, run
and forget.

### DIFF
--- a/tests/bledtm/syscfg.yml
+++ b/tests/bledtm/syscfg.yml
@@ -1,0 +1,33 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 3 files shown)


tests/bledtm/syscfg.yml
```diff
@@ -0,0 +1,33 @@
...
+syscfg.defs:
+    DTM_RF_CHANNEL:
+        description: RF channel used by transmitter
+        value: 0
+        range: 0..39
+    DTM_PAYLOAD_LENGTH:
+        description: Length of payload of the test reference packets
+        value: 100
+        range: 0..255
+    DTM_PAYLOAD_TYPE:
```
rymanluk: since name is bledtm I would add here DTM_ROLE: so we could also configure second board on RX.
On RX side we could have a timer which would stop DTM. Once timer is stopped we could print the result and start RX and time again

andrzej-kaczmarek: agree, but we can add this syscfg when we have RX role, for now it's redundant

rymanluk: We need it to verify TX role :)

rymanluk: To be clear, we need RX role, not only syscfg

andrzej-kaczmarek: sure, but we can verify using blehci as usual and add RX role later - I do not think that lack of RX support in this app should be considered a blocker, someone can add it later if needed.

sjanc:
I'd just add comment in readme file about intention of having tests/

PR #599, opened 2019-10-04
nimble/ll: Fix crash when scanning for AUX packets

sjanc:
When dropping scan for AUX packet we need to make sure to send HCI
report with data truncated flag set whe needed. Otherwise assert is
triggered in ble_ll_scan_aux_data_unref (as this leaves host in
undefined state).


nimble/controller/src/ble_ll_scan.c
```diff
@@ -182,6 +182,19 @@ static struct os_mempool ext_adv_pool;
...
+static void
+ble_ll_aux_scan_drop(struct ble_ll_aux_data *aux_data)
+{
+    struct ble_ll_scan_sm *scansm = &g_ble_ll_scan_sm;
+
+    STATS_INC(ble_ll_stats, aux_scan_drop);
+
+    BLE_LL_ASSERT(aux_data);
+
+    ble_npl_event_set_arg(&scansm->aux_scan_drop_ev, aux_data);
```
andrzej-kaczmarek: what if you need to drop aux before previous one was handled by ll_task?

rymanluk: Scheduling next chain happens in the LL so for sure dropping will happen after previous chain is handled since it goes thru LL as well.  Anyway there can be 2 possibilities
1) previous chain has been sent with success to the host
2) previous chain has been sent as truncated

In both cases we should be fine thanks to ref counter and flags which keeps information that truncated has been sent already to the host.

andrzej-kaczmarek: if you have aux scans for 2 separate instances already scheduled then you may need to drop aux from 2nd chain before event sent for 1st chain was handled in ll_task

rymanluk: ok, now I see what you are saying. Agree. Then this drop event probably should be in aux_data

sjanc:
updated with event in aux_data

PR #607, opened 2019-10-10
nimble/gatt: fix custom notify/indicate to work with proper connection handle

mnasr93:
suppose using service that has multiple clients (i.e. ans), and each client has different instance for service characteristics.
when server notifies this characteristic using ble_gattc_notify(), then server can not read the proper instance using connection handle in the service access callback, because conn_handle is overwritten in ble_gattc_notify_custom() API with BLE_HS_CONN_HANDLE_NONE

### DIFF
--- a/nimble/host/src/ble_gattc.c
+++ b/nimble/host/src/ble_gattc.c
@@ -4166,8 +4166,8 @@ ble_gattc_notify_custom(uint16_t conn_handle, uint16_t chr_val_handle,
             rc = BLE_HS_ENOMEM;
             goto done;
         }
-        rc = ble_att_svr_read_handle(BLE_HS_CONN_HANDLE_NONE,
-                                     chr_val_handle, 0, txom, NULL);
+        rc = ble_att_svr_read_handle(conn_handle |
+                BLE_ATT_SVR_CHECK_PERMS_MASK, chr_val_handle, 0, txom, NULL);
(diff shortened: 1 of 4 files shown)


nimble/host/src/ble_gattc.c
```diff
@@ -4166,8 +4166,8 @@ ble_gattc_notify_custom(uint16_t conn_handle, uint16_t chr_val_handle,
             rc = BLE_HS_ENOMEM;
             goto done;
         }
-        rc = ble_att_svr_read_handle(BLE_HS_CONN_HANDLE_NONE,
-                                     chr_val_handle, 0, txom, NULL);
+        rc = ble_att_svr_read_handle(conn_handle |
```
rymanluk: what will happen if we change here on from `BLE_HS_CONN_HANDLE_NONE` to `conn_handle` ?

I understand that there will be ` ble_att_svr_check_perms` triggered, but does it creates any issues?

mnasr93: yes it creates an issue, 
if the characteristic to be notified has only write and notify permissions,
then reading the characteristic value to be notified will fail

rymanluk: you are right, then we need to solve it somehow but for sure in some other way than masking conn_handle. Maybe some additional bool parameter to that function saying it is local_read or not?

PR #608, opened 2019-10-11
nimble/mesh: Fix provisioning rx_buf initialization

rymanluk:
This patch fixes regresion after
 5f56107d nimble/mesh: Sync provisioning code with Zephyr

When PB GATT is not used then rx_buf would be not initialized


rymanluk:
Travis fails due to unrelated reason - therefor merging this PR

PR #613, opened 2019-10-16
couple host scan fixes

sjanc:


nimble/host/src/ble_gap.c
```diff
@@ -4030,6 +4030,12 @@ ble_gap_disc(uint8_t own_addr_type, int32_t duration_ms,
     p.passive = disc_params->passive;
     p.window = disc_params->window;
 
+    if (duration_ms == BLE_HS_FOREVER) {
+        duration_ms = 0;
+    } else if (duration_ms == 0) {
```
rymanluk: behavior on duration_ms == 0 is not specified in the API, however indeed it worked like that in legacy scan.
Would be worth to add it there.

PR #630, opened 2019-10-30
nimble/host: Use os_mbuf_pack_chains

wes3:
if BLE_L2CAP_JOIN_RX_FRAGS is set to 1, use the new API to pack
mbuf chains instead of previous method. This new API will do more
copying (possibly) but will better utilize memory in that it
will use the entire data buffer of each mbuf in the chain.

NOTE: This PR requires PR #2075 from apache-mynewt-core to be used. That PR has been merged into core but just mentioning it here so folks know to use the proper version of mynewt-core.


andrzej-kaczmarek:
you need to copy changes from mynewt-core to nimble as otherwise ports won't build

PR #636, opened 2019-10-30
NimBLE/host: `ble_gap_unpair` should continue un-pairing after connection termination

prasad-alatkar:
1. Modifies `ble_hs_pvcy.c` to translate the `peer_addr_type` to valid HCI parameters (0x00, 0x01). Modify `ble_hs_pvcy_remove_entry` to not treat return code `BLE_ERR_UNK_CONN_ID` as an error. The reason for this change is in `ble_gap_unpair`, `ble_hs_pvcy_remove_entry` is called now as per the spec Vol 2, Part E, section 7.8.39 

> When a Controller cannot remove a device from the resolving list because it is
not found, it shall return the error code Unknown Connection Identifier (0x02).

In `ble_gap_unpair` this is treated as an error and un-pairing is discontinued.     

2. In `ble_gap_unpair`, if peer is connected we should not stop with just terminating
  connection, instead we should continue the un-pairing process after disconnection.

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -5185,7 +5185,6 @@ int
 ble_gap_unpair(const ble_addr_t *peer_addr)
 {
     struct ble_hs_conn *conn;
-    int rc;
 
     if (ble_addr_cmp(peer_addr, BLE_ADDR_ANY) == 0) {
         return BLE_HS_EINVAL;
@@ -5195,20 +5194,13 @@ ble_gap_unpair(const ble_addr_t *peer_addr)
 
     conn = ble_hs_conn_find_by_addr(peer_addr);
     if (conn != NULL) {
-        rc = ble_gap_terminate_with_conn(conn, BLE_ERR_REM_USER_CONN_TERM);
-        if ((rc != BLE_HS_EALREADY) && (rc != BLE_HS_ENOTCONN)) {
-            ble_hs_unlock();
-            return rc;
-        }
+        ble_gap_terminate_with_conn(conn, BLE_ERR_REM_USER_CONN_TERM);
     }
 
     ble_hs_unlock();
 
-    rc = ble_hs_pvcy_remove_entry(peer_addr->type,
-                                  peer_addr->val);
-    if (rc != 0) {
-        return rc;
-    }
+    ble_hs_pvcy_remove_entry(peer_addr->type,
+                             peer_addr->val);
 
     return ble_store_util_delete_peer(peer_addr);
 }
--- a/nimble/host/src/ble_hs_pvcy.c
+++ b/nimble/host/src/ble_hs_pvcy.c
@@ -65,7 +65,7 @@ ble_hs_pvcy_remove_entry(uint8_t addr_type, const uint8_t *addr)
     struct ble_hci_le_rmv_resolve_list_cp cmd;
 
     if (addr_type > BLE_ADDR_RANDOM) {
-        return BLE_ERR_INV_HCI_CMD_PARMS;
+        addr_type = addr_type % 2;
     }
 
     cmd.peer_addr_type = addr_type;


nimble/host/src/ble_hs_pvcy.c
```diff
@@ -64,6 +64,12 @@ ble_hs_pvcy_remove_entry(uint8_t addr_type, const uint8_t *addr)
 {
     struct ble_hci_le_rmv_resolve_list_cp cmd;
 
+    /* First check if privacy started flag is enabled, no point in going
+     * further if pvcy_started is not enabled */
+    if (!ble_hs_pvcy_started) {
```
rymanluk: ble_hs_prvcy_started  means only that rpa timeout has been set in the controller and Controller will keep generating local RPA.
If timeout is disabled we can still use resolving list.
It is usually a case when NImble use public address but still wants to use resolving list.

Therefore I believe that this change will actually break Nimble when it not uses RPA.

nimble/host/src/ble_gap.c
```diff
@@ -5192,7 +5192,7 @@ ble_gap_unpair(const ble_addr_t *peer_addr)
     conn = ble_hs_conn_find_by_addr(peer_addr);
     if (conn != NULL) {
         rc = ble_gap_terminate_with_conn(conn, BLE_ERR_REM_USER_CONN_TERM);
-        if ((rc != BLE_HS_EALREADY) && (rc != BLE_HS_ENOTCONN)) {
+        if ((rc != 0) &&  (rc != BLE_HS_EALREADY) && (rc != BLE_HS_ENOTCONN)) {
```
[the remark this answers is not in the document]
prasad-alatkar: Sure @rymanluk I will modify `ble_hs_pvcy_remove_entry` accordingly. I think apart from this, there is need to change `ble_gap_unpair` a little, where return code `BLE_ERR_UNK_CONN_ID` from `ble_hs_pvcy_remove_entry` should also not be considered as error.

nimble/host/src/ble_hs_pvcy.c
prasad-alatkar: Thank you for the clarification !!

nimble/host/src/ble_hs_pvcy.c
```diff
@@ -63,17 +63,23 @@ int
...
+        addr_type = addr_type % 2;
     }
 
     cmd.peer_addr_type = addr_type;
     memcpy(cmd.peer_id_addr, addr, BLE_DEV_ADDR_LEN);
 
-    return ble_hs_hci_cmd_tx(BLE_HCI_OP(BLE_HCI_OGF_LE,
-                                        BLE_HCI_OCF_LE_RMV_RESOLV_LIST),
-                             &cmd, sizeof(cmd), NULL, 0);
+    rc = ble_hs_hci_cmd_tx(BLE_HCI_OP(BLE_HCI_OGF_LE,
```
rymanluk: @prasad-alatkar  does it really can happen?

prasad-alatkar: @rymanluk actually in our case (NimBLE host only) it does happen. As per the spec,
  
>   When a Controller cannot remove a device from the resolving list because it is not found, it shall return the error code Unknown Connection Identifier (0x02).

Please let me know if I am missing anything.

rymanluk: You are right, so this is valid error massage and we should just allow to send it back to the app.

prasad-alatkar: Makes sense... !! Then I might need to handle `BLE_ERR_UNK_CONN_ID` in `ble_gap_unpair` here: [return_from_ble_hs_pvcy_remove_entry](https://github.com/apache/mynewt-nimble/pull/636/files#diff-7075c75d7ac538394ce8b3302c237cdaR5201),  what do you think ?

rymanluk: I would just allow it to go back as a return on ble_gap_unpair()

prasad-alatkar: But it will again block `ble_gap_unpair` to call ` ble_store_util_delete_peer`. I mean, what if the device was not present in the resolvable list and still we are trying to remove it ? From user's perspective `ble_gap_unpair` API should complete calling all the 3 APIs till `ble_store_util_delete_peer`.

rymanluk: @prasad-alatkar Ah ok, I missed `ble_store_util_delete_peer` which is called after `ble_hs_pvcy_remove_entry` and of course course it will fail for devices which does not use privacy. 

In such a case I would ignore `rc` from `ble_hs_pvcy_remove_entry` at all.

prasad-alatkar: Thank you @rymanluk , will update the PR accordingly !!

PR #639, opened 2019-11-06
nimble/ll: Fix extended scanner data fragmentation issues

andrzej-kaczmarek:
This fixes major issue with aux_data->flags which are updated from both ISR and LL at the same time so LL may use incorrect flags state to process data which leads to data stream which cannot be reassembled properly.

...and few small fixes for other things.

### DIFF
--- a/nimble/controller/src/ble_ll_scan.c
+++ b/nimble/controller/src/ble_ll_scan.c
@@ -436,7 +436,7 @@ ble_ll_scan_get_ext_adv_report(struct ext_adv_report *copy_from)
 }
 
 static void
-ble_ll_scan_send_truncated_if_chained(struct ble_ll_aux_data *aux_data)
+ble_ll_scan_send_truncated(struct ble_ll_aux_data *aux_data)
 {
     struct ble_hci_ev_le_subev_ext_adv_rpt *ev;
     struct ext_adv_report *report;
@@ -448,8 +448,9 @@ ble_ll_scan_send_truncated_if_chained(struct ble_ll_aux_data *aux_data)
 
     BLE_LL_ASSERT(aux_data);
 
-    if (!BLE_LL_AUX_CHECK_FLAG(aux_data, BLE_LL_AUX_CHAIN_BIT)) {
-        /* if not chained, there is nothing to do here */
+    /* No need to send if we did not send any report or sent truncated already */
+    if (!(aux_data->flags_ll & BLE_LL_AUX_FLAG_HCI_SENT_ANY) ||
+        (aux_data->flags_ll & BLE_LL_AUX_FLAG_HCI_SENT_TRUNCATED)) {
         return;
     }
 
@@ -465,8 +466,7 @@ ble_ll_scan_send_truncated_if_chained(struct ble_ll_aux_data *aux_data)
     report->data_len = 0;
 
     report->evt_type = aux_data->evt_type;
-    report->evt_type |= (BLE_HCI_ADV_DATA_STATUS_TRUNCATED);
-    BLE_LL_AUX_SET_FLAG(aux_data, BLE_LL_AUX_TRUNCATED_SENT);
+    report->evt_type |= BLE_HCI_ADV_DATA_STATUS_TRUNCATED;
 
     if (BLE_LL_AUX_CHECK_FLAG(aux_data, BLE_LL_AUX_HAS_ADDRA)) {
         memcpy(report->addr, aux_data->addr, 6);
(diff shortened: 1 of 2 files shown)


nimble/controller/src/ble_ll_scan.c
```diff
@@ -1205,11 +1204,9 @@ ble_ll_scan_aux_data_unref(struct ble_ll_aux_data *aux_data)
     ble_ll_trace_u32x2(BLE_LL_TRACE_ID_AUX_UNREF, (uint32_t) aux_data, aux_data->ref_cnt);
 
     if (aux_data->ref_cnt == 0) {
-        /* Below assert is to detect missing complete or truncated event in case of chaining */
-        BLE_LL_ASSERT(!((BLE_LL_AUX_CHECK_FLAG(aux_data, BLE_LL_AUX_CHAIN_BIT) &&
-                            BLE_LL_AUX_CHECK_FLAG(aux_data, BLE_LL_SENT_EVENT_TO_HOST)) &&
-                            !(BLE_LL_AUX_CHECK_FLAG(aux_data, BLE_LL_AUX_TRUNCATED_SENT) ||
-                            !BLE_LL_AUX_CHECK_FLAG(aux_data, BLE_LL_AUX_INCOMPLETE_BIT))));
+        BLE_LL_ASSERT(!(aux_data->flags_ll & BLE_LL_AUX_FLAG_SCAN_ERROR) ||
```
sjanc: I'm not sure if this is doing what it is suppose to do

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2647,15 +2648,18 @@ ble_ll_hci_send_ext_adv_report(uint8_t ptype, uint8_t *adva, uint8_t adva_type,
...
+
         if (aux_data &&
-                BLE_LL_AUX_CHECK_FLAG(aux_data, BLE_LL_SENT_EVENT_TO_HOST) &&
-                BLE_LL_AUX_CHECK_FLAG(aux_data, BLE_LL_AUX_CHAIN_BIT)) {
-            report = ev->reports;
+            (aux_data->flags_ll & BLE_LL_AUX_FLAG_HCI_REPORT_SENT) &&
+            !(aux_data->flags_ll & BLE_LL_AUX_FLAG_HCI_TRUNCATED_SENT)) {
+
+            aux_data->flags_ll |= BLE_LL_AUX_FLAG_SCAN_ERROR;
+            aux_data->flags_ll |= BLE_LL_AUX_FLAG_HCI_REPORT_SENT;
```
sjanc: report_sent should already be set here, shoudn't it?

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2707,42 +2706,35 @@ ble_ll_hci_send_ext_adv_report(uint8_t ptype, uint8_t *adva, uint8_t adva_type,
...
-        if (aux_data) {
-            if (!(aux_data->evt_type & BLE_HCI_ADV_SCAN_MASK) ||
-                (aux_data->evt_type & BLE_HCI_ADV_SCAN_RSP_MASK)) {
-                BLE_LL_AUX_SET_FLAG(aux_data, BLE_LL_SENT_EVENT_TO_HOST);
+            if (!(aux_data->flags_ll & BLE_LL_AUX_FLAG_HCI_REPORT_SENT)) {
+                ble_hci_trans_buf_free((uint8_t *)hci_ev);
+                goto  done;
             }
+
+            aux_data->flags_ll |= BLE_LL_AUX_FLAG_HCI_REPORT_SENT;
```
sjanc: same here,  report_sent is already set (checked in if above)

andrzej-kaczmarek:
@sjanc updated flags and checks should make more sense now :)

nimble/controller/src/ble_ll_scan.c
```diff
@@ -2685,19 +2702,12 @@ ble_ll_hci_send_ext_adv_report(uint8_t ptype, uint8_t *adva, uint8_t adva_type,
...
-        }
-
-        /* Assume data status is not "completed" */
-        rc = 1;
+        /*
+         * We need another event if either there are still some data left to
+         * send in current PDU or scan is not completed. The only exception is
+         * when this is a scannable event which is not a scan response.
+         */
+        need_event = ((offset < datalen) || (aux_data && !(aux_data->flags_ll & BLE_LL_AUX_FLAG_SCAN_COMPLETE))) && !is_scannable_aux;
```
rymanluk: For the moment I was thinking it will not work because in case of long scannable_aux  we might need next event, but it is not true as scannable aux has no advertising data - maybe we could point it here in the comment so it is clear.

btw. this line is too long.

nimble/controller/src/ble_ll_scan.c
rymanluk: it does :)

PR #646, opened 2019-11-13
nimble/ll: Replace BLE_NUM_COMP_PKT_RATE with an interval in ms

michal-narajowski:
Using an OS specific define in MYNEWT_VAL can be cumbersome
when porting to other systems. Replace this value with
universal interval in milliseconds.


sjanc:
@haukepetersen FYI

PR #648, opened 2019-11-14
porting: Add targets to generate syscfg.h file and others for porting examples

michal-narajowski:
This PR introduces porting targets. These targets can be used to generate syscfg.h and other
files that Newt generates on compilation and that are required to compile ported Nimble. It makes it easier to update syscfg.h, to track changes in these files and to track dependencies of a target. There are some additional commits that make this change possible.


michal-narajowski:
I just force-pushed an update that keeps syscfg values for RIOT port the way they before.

michal-narajowski:
@haukepetersen FYI

porting/targets/riot/syscfg.yml
```diff
@@ -0,0 +1,30 @@
...
+    BLE_HCI_EVT_HI_BUF_COUNT: 2
+    BLE_HW_WHITELIST_ENABLE: 0
+    BLE_LL_CFG_FEAT_DATA_LEN_EXT: 0
+    BLE_LL_CFG_FEAT_LE_CSA2: 1
+    BLE_LL_CFG_FEAT_LE_ENCRYPTION: 0
+    BLE_LL_CFG_FEAT_LL_PRIVACY: 0
+    BLE_MAX_PERIODIC_SYNCS: 0
+    MSYS_1_BLOCK_COUNT: 5
+    MSYS_1_BLOCK_SIZE: 88
+    XTAL_32768: 1
```
haukepetersen: newline missing?

haukepetersen:
I am not too familiar with the `newt` tool (yet): with this change its simply a matter of editing the `porting/targets/riot/syscfg.yml` file for RIOT specific config options, and then triggering the `porting/riot` target for generating the actual `syscfg.h` file, right? So no more manual editing of the `syscfg.h` file for synchronizing with newer NimBLE versions?! I love it :-)

porting/targets/riot/syscfg.yml
michal-narajowski: Thanks, fixed.

michal-narajowski:
@haukepetersen Yes, that is correct! Although you will also need to copy these files from `<proj-root>/bin/...` because that's where the `newt` tool generates them. 

This will make updating these files easier and also it will be easier to see what new configs are there.  There are still some superfluous values that are OS-specific but they come from dependencies that are automatically added by nimble packages so for now, I was not able to get rid of them.

porting/targets/riot/syscfg.yml
haukepetersen: Could you also add `BLE_LL_CONN_INIT_MAX_TX_BYTES:  BLE_LL_MAX_PKT_SIZE` to this list?! see https://github.com/apache/mynewt-nimble/issues/649

michal-narajowski: Sure! Added BLE_LL_CONN_INIT_MAX_TX_BYTES to syscfg.

haukepetersen: Thx

haukepetersen:
Just wanted to give this a try, but it seems my `newt` knowledge is insufficient. Lets say I want to generate the syscfg for RIOT, who do I tigger the `riot` target? I thought its something like
- create a new project (`newt new ...`)
- add target to the new project somehow - here I failed...

Any pointers?

michal-narajowski:
@haukepetersen To test it you would have to take the following steps:
- `newt new <project name>`
- `cd <project-name>`
- edit project.yml file and change the version of `apache-mynewt-core` repo to `0.0.0`
- `newt upgrade`
- now you should have all the repositories downloaded and set to `master`
- `newt build riot` (riot target is already in the nimble repo and this will compile it)
- Then you can access files in `bin/@apache-mynewt-nimble/targets/riot/generated/include`

haukepetersen:
works as expected - the version and upgrade steps were the missing link.

Thanks a lot for helping a lazy newbie, who is too lazy to properly read the `newt` documentation :-)

PR #652, opened 2019-11-15
nimble/ll: Add LLCP tracing via HCI events

andrzej-kaczmarek:
This adds option to trace LLCP PDUs via HCI events. Each LLCP PDU is
sent in a vendor-specific HCI event which can be decoded e.g. by btmon.
Identifier and format of HCI event os the same as used by controllers
from Intel so it may be necessary to fake NimBLE's manufacturer id by
setting 'BLE_LL_MFRG_ID: 2' so btmon can decode events properly.


andrzej-kaczmarek:
updated code to send txed llcp after it was actually sent, not when it was enqueued. this seems to work fine.

PR #655, opened 2019-11-15
nimble/ll: Fix for race which leads to assert in ble_ll_conn

rymanluk:


nimble/controller/src/ble_ll_conn.c
```diff
@@ -1808,13 +1810,33 @@ ble_ll_conn_end(struct ble_ll_conn_sm *connsm, uint8_t ble_err)
...
+     */
+    OS_ENTER_CRITICAL(sr);
+    if (g_ble_ll_conn_cur_sm == connsm) {
+        /* Disable the PHY */
+        ble_phy_disable();
+        /* Disable the wfr timer */
+        ble_ll_wfr_disable();
+        /* Link-layer is in standby state now */
+        ble_ll_state_set(BLE_LL_STATE_STANDBY);
+        /* Set current LL connection to NULL */
```
andrzej-kaczmarek: all these comments are redundant since all calls are self-explanatory

and seems like you just copied this from `ble_ll_conn_current_sm_over` so why not make it a common helper?

rymanluk: can do it for you

rymanluk: @andrzej-kaczmarek ping

PR #678, opened 2019-11-22
controller/adv: fix 'unused function' warning

haukepetersen:
When compiling without the LL_EXT_ADV feature and with disabled
assertions, the compiler throws an 'unused function' warning for
the internal function `ble_ll_adv_active_chanset_is_pri()`. This
commit fixes this by only compiling that functions if actually
needed.

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -230,13 +230,13 @@ ble_ll_adv_sm_find_configured(uint8_t instance)
     return NULL;
 }
 
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_EXT_ADV)
 static int
 ble_ll_adv_active_chanset_is_pri(struct ble_ll_adv_sm *advsm)
 {
     return (advsm->flags & BLE_LL_ADV_SM_FLAG_ACTIVE_CHANSET_MASK) == 0x10;
 }
 
-#if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_EXT_ADV)
 static int
 ble_ll_adv_active_chanset_is_sec(struct ble_ll_adv_sm *advsm)
 {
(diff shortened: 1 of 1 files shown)


nimble/controller/src/ble_ll_adv.c
```diff
@@ -199,11 +199,13 @@ struct ble_ll_adv_sm
 #define SYNC_DATA_LEN(_advsm) \
                 (_advsm->periodic_adv_data ? OS_MBUF_PKTLEN(advsm->periodic_adv_data) : 0)
 
+#if defined (BLE_LL_CFG_FEAT_LL_EXT_ADV) || !defined(NDEBUG)
```
andrzej-kaczmarek: should be `MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_EXT_ADV)` here

sjanc:
I have a bit of mixed feelings about this !define(NDEBUG), while in short term it is fine maybe in long term we should always use BLE_LL_ASSERT() and evaluate parameters but make it dead code for NDEBUG...

andrzej-kaczmarek:
we could also remove all those asserts for `is_pri` since they were primarily to debug things when adv code was refactored. I did not see any of them actually trigger an error in long long time.

haukepetersen:
I share the same feeling about `NDEBUG`, but it was the best I could think of right now...

+1 for simply removing the asserts. @sjanc do you agree as well? If yes, I will adapt this PR.

haukepetersen:
so how does this look?

sjanc:
yeap, seems fine

haukepetersen:
perfect, squashed just in case @andrzej-kaczmarek likes it as well :-)

andrzej-kaczmarek:
try to merge or let me know in case you can't do it, I'll merge then

haukepetersen:
I am still not allowed to merge, so please go ahead.

Is there maybe anyone else I should contact regarding the repository permissions?

rymanluk:
@haukepetersen I gave you instruction whom to contact on the slack

haukepetersen:
rebased. Still working on getting the commit permissions - was occupied elsewhere the last weeks...

PR #683, opened 2019-11-28
[RFC] nimble/ll: Allow to update resolving list

rymanluk:
So far, when host was trying to add keys for already existing entry,
controller rejected it.
Since it is not really specified how the Controller should handle it
and because we saw some host was doing it in order to update keys,
Controller will be nicer now and will allow to update keys

Note: This is RFC and I need to run certification test cases as I'm not sure if there is any covering that scenario.

### DIFF
--- a/nimble/controller/src/ble_ll_resolv.c
+++ b/nimble/controller/src/ble_ll_resolv.c
@@ -273,6 +273,37 @@ ble_ll_resolv_list_find(const uint8_t *addr, uint8_t addr_type)
 }
 
 /**
+ * Helper function to remove address from the resolving list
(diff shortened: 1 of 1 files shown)


rymanluk:
@sjanc good point and yes this is back to network and I should put it in the comment.
I also think this is what we want.
As you know this is already undefined behavior and error code we are sending now is not great as it does not really say nothing valuable to the host what is wrong.

I think the assumption that the only issue on the host side is that they forget to remove old key is good enough. So the controller will just drop old entry and create new one with a new keys. Host need to configure privacy mode

PR #684, opened 2019-11-30
Update ble_l2cap.c

Francois-Belanger:
Fix potential null pointer crash

### DIFF
--- a/nimble/host/src/ble_l2cap.c
+++ b/nimble/host/src/ble_l2cap.c
@@ -403,7 +403,9 @@ ble_l2cap_tx(struct ble_hs_conn *conn, struct ble_l2cap_chan *chan,
 
     case BLE_HS_EAGAIN:
         /* Controller could not accommodate full packet.  Enqueue remainder. */
-        STAILQ_INSERT_TAIL(&conn->bhc_tx_q, OS_MBUF_PKTHDR(txom), omp_next);
+        if (txom) {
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_l2cap.c
```diff
@@ -403,7 +403,9 @@ ble_l2cap_tx(struct ble_hs_conn *conn, struct ble_l2cap_chan *chan,
 
     case BLE_HS_EAGAIN:
         /* Controller could not accommodate full packet.  Enqueue remainder. */
-        STAILQ_INSERT_TAIL(&conn->bhc_tx_q, OS_MBUF_PKTHDR(txom), omp_next);
+        if (txom) {
+            STAILQ_INSERT_TAIL(&conn->bhc_tx_q, OS_MBUF_PKTHDR(txom), omp_next);
+        }
```
andrzej-kaczmarek: I did not look into code too much, but this does look like a workaround for an issue in in `ble_hs_hci_acl_tx` which should not return `BLE_HS_EAGAIN` and nullify `txom`. Basically with this fix we would return success to caller and drop packet.

rymanluk: @francois-berder any chances to get traces from that issue?

Francois-Belanger: Hi, will try to test within days.  Was on another project.

Francois-Belanger:
Just used the latest idf, bug seems fixed.  Unable to reproduce :-)

IDF v4.1-dev-2224-gbcd2358eb

PR #688, opened 2019-12-04
nimble/host: Add support for OOB Secure Connections

michal-narajowski:
Previously we supported only Legacy OOB pairing procedure.
Secure connections version uses a very different Authentication stage
which required some changes to the SM code.

SC OOB uses two 128-bit values as OOB data. OOB communication
is possible in one or both directions. Due to the difference in
OOB Data compared to the Legacy version, we had to introduce additional
IO Action that is used in ble_sm_inject_io() API. Also the CONFIRM
state is omitted as there is no confirmation data sent in-band. The
confirmation is done when application is passing OOB data to the host
using ble_sm_inject_io(). The application should call
ble_sm_sc_oob_generate_data() to generate local OOB data for later use.

### DIFF
--- a/apps/btshell/src/main.c
+++ b/apps/btshell/src/main.c
@@ -119,6 +119,9 @@ struct btshell_tx_data_s
 static struct btshell_tx_data_s btshell_tx_data;
 int btshell_full_disc_prev_chr_val;
 
+struct ble_sm_sc_oob_data oob_data_local;
+struct ble_sm_sc_oob_data oob_data_remote;
+
 #define XSTR(s) STR(s)
 #ifndef STR
 #define STR(s) #s
@@ -2102,6 +2105,16 @@ btshell_on_reset(int reason)
 static void
 btshell_on_sync(void)
 {
+#if MYNEWT_VAL(BLE_SM_SC)
+    int rc;
+
+    rc = ble_sm_sc_oob_generate_data(&oob_data_local);
+    if (rc) {
+        console_printf("Error: generating oob data; reason=%d\n", rc);
+        return;
+    }
+#endif
+
     console_printf("Host and controller synced\n");
 }
 
--- a/nimble/host/include/host/ble_sm.h
+++ b/nimble/host/include/host/ble_sm.h
@@ -84,7 +84,16 @@ extern "C" {
 #define BLE_SM_IOACT_INPUT                      2
 #define BLE_SM_IOACT_DISP                       3
 #define BLE_SM_IOACT_NUMCMP                     4
-#define BLE_SM_IOACT_MAX_PLUS_ONE               5
+#define BLE_SM_IOACT_OOB_SC                     5
+#define BLE_SM_IOACT_MAX_PLUS_ONE               6
+
+struct ble_sm_sc_oob_data {
+    /** Random Number. */
+    uint8_t r[16];
+
+    /** Confirm Value. */
+    uint8_t c[16];
+};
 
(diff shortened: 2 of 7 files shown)


michal-narajowski:
This feature was tested manually with two devices running btshell and automatically with bttester and auto-pts.

```
SM/MAS/SCOB/BV-01-C   PASS
SM/MAS/SCOB/BV-04-C   PASS
SM/MAS/SCOB/BI-01-C   PASS
SM/MAS/SCOB/BI-04-C   PASS
SM/SLA/SCOB/BV-02-C   PASS
SM/SLA/SCOB/BV-03-C   PASS
SM/SLA/SCOB/BI-02-C   PASS
SM/SLA/SCOB/BI-03-C   PASS
```

nimble/host/src/ble_sm_priv.h
```diff
@@ -269,6 +269,8 @@ struct ble_sm_proc {
     struct ble_sm_public_key pub_key_peer;
     uint8_t mackey[16];
     uint8_t dhkey[32];
+    const struct ble_sm_sc_oob_data *oobd_local;
+    const struct ble_sm_sc_oob_data *oobd_remote;
```
andrzej-kaczmarek: better use `oob_data_*` to be consistent with existing names

nimble/host/src/ble_sm_priv.h
```diff
@@ -300,8 +300,8 @@ int ble_sm_alg_c1(uint8_t *k, uint8_t *r,
                   uint8_t iat, uint8_t rat,
                   uint8_t *ia, uint8_t *ra,
                   uint8_t *out_enc_data);
-int ble_sm_alg_f4(uint8_t *u, uint8_t *v, uint8_t *x, uint8_t z,
-                  uint8_t *out_enc_data);
+int ble_sm_alg_f4(const uint8_t *u, const uint8_t *v, const uint8_t *x,
+                  uint8_t z, uint8_t *out_enc_data);
```
andrzej-kaczmarek: would be nice to change it at least for all ble_sm_alg

apps/btshell/src/main.c
```diff
@@ -2102,7 +2105,15 @@ btshell_on_reset(int reason)
 static void
 btshell_on_sync(void)
 {
+    int rc;
+
     console_printf("Host and controller synced\n");
+
+    rc = ble_sm_sc_oob_generate_data(&oobd_local);
```
andrzej-kaczmarek: `#if NIMBLE_BLE_SM`

michal-narajowski:
Thanks, @andrzej-kaczmarek. I fixed the issues you mentioned.

nimble/host/include/host/ble_sm.h
```diff
@@ -84,17 +84,32 @@ extern "C" {
...
 struct ble_sm_io {
     uint8_t action;
     union {
         uint32_t passkey;
         uint8_t  oob[16];
         uint8_t  numcmp_accept;
+        struct {
+	    struct ble_sm_sc_oob_data *oob_sc_local;
+	    struct ble_sm_sc_oob_data *oob_sc_remote;
+	};
```
andrzej-kaczmarek: something went wrong here :)

also I think it would be better to do like this:
```
struct {
    struct ble_sm_sc_oob_data *local;
    struct ble_sm_sc_oob_data *remote;
} oob_sc_data;
```

michal-narajowski:
Thanks again, fixed.

PR #696, opened 2019-12-09
apps/bttester: Fixes and changes related to PTS update to 7.5.1

michal-narajowski:


apps/bttester/syscfg.yml
```diff
@@ -33,11 +33,11 @@ syscfg.defs:
 
     BTTESTER_RTT_BUFFER_SIZE_UP:
         description: Bttester upstream buffer size
-        value: 512
+        value: MYNEWT_VAL_BTTESTER_BTP_DATA_SIZE_MAX
```
andrzej-kaczmarek: should be `MYNEWT_VAL(BTTESTER_BTP_DATA_SIZE_MAX)` so newt can evaluate it

michal-narajowski: Thanks, fixed.

PR #697, opened 2019-12-11
Added extra argument to sync and reset collback to simplify building c++ wrapper

redradist:


ccollins476ad:
I like the idea, but this change breaks backwards compatibility.  If we want this feature, I think we'll need to devise a way of getting it in while maintaining compatibility.

andrzej-kaczmarek:
would be good if you could explain what is the use case related to this change...

redradist:
> would be good if you could explain what is the use case related to this change...

Use-case is pretty simple:

`cpp
Wrapper::Wrapper() {
...
  ble_hs_cfg.reset_cb = &NimBleController::onResetCallback;
  ble_hs_cfg.reset_cb_arg = this;
...
}

void Wrapper::onResetCallback(int _reason, void *arg) {
  reinterpret_cast<Wrapper*>(arg)->onResetHandler(_reason);
}

void Wrapper::onResetHandler(int _reason) {
  ...
}
`

redradist:
@ccollins476ad 
> I like the idea, but this change breaks backwards compatibility. If we want this feature, I think we'll need to devise a way of getting it in while maintaining compatibility.

But it also is not good to have lots of functions like this:
```c
...
    ble_hs_reset_fn *reset_cb;

    ble_hs_reset_extra_fn *reset_extra_cb;
    void *reset_extra_cb_arg;

    ble_hs_sync_fn *sync_cb;

    ble_hs_sync_extra_fn *sync_extra_cb;
    void *sync_extra_cb_arg;
...
```

andrzej-kaczmarek:
so basically you can have static member in class and assign `this` there. it will be pretty much the same since `ble_hs_cfg` is global variable.

redradist:
@andrzej-kaczmarek 
> so basically you can have static member in class and assign `this` there. it will be pretty much the same since `ble_hs_cfg` is global variable.

Sorry but it is a bad solution ))
What's the problem ? You already have addition arguments for other callbacks:
```c
...
    /**
     * An optional callback that gets executed upon registration of each GATT
     * resource (service, characteristic, or descriptor).
     */
    ble_gatt_register_fn *gatts_register_cb;

    /**
     * An optional argument that gets passed to the GATT registration
     * callback.
     */
    void *gatts_register_arg;
...
```

And also those args could be used not only for this pointer, but in case of C Language for any other purposes ...
Also it is standard way for providing C API that could be optionally wrapped by C++ code

ccollins476ad:
> But it also is not good to have lots of functions like this:
> 
> ```c
> ...
>     ble_hs_reset_fn *reset_cb;
> 
>     ble_hs_reset_extra_fn *reset_extra_cb;
>     void *reset_extra_cb_arg;
> 
>     ble_hs_sync_fn *sync_cb;
> 
>     ble_hs_sync_extra_fn *sync_extra_cb;
>     void *sync_extra_cb_arg;
> ...
> ```

You may have a good point here (I didn't understand the point), but it is irrelevant.  We have existing users and we simply cannot break backwards compatibility without a transition period.

redradist:
@ccollins476ad 
> > But it also is not good to have lots of functions like this:
> > ```c
> > ...
> >     ble_hs_reset_fn *reset_cb;
> > 
> >     ble_hs_reset_extra_fn *reset_extra_cb;
> >     void *reset_extra_cb_arg;
> > 
> >     ble_hs_sync_fn *sync_cb;
> > 
> >     ble_hs_sync_extra_fn *sync_extra_cb;
> >     void *sync_extra_cb_arg;
> > ...
> > ```
> 

> You may have a good point here (I didn't understand the point), but it is irrelevant. We have existing users and we simply cannot break backwards compatibility without a transition period.

Would you like this temporary solution as I suggested ?
Also I could add comment that in future ble_hs_reset_fn and ble_hs_sync_fn will be extended with extra argument ??

andrzej-kaczmarek:
> Would you like this temporary solution as I suggested ?
> Also I could add comment that in future ble_hs_reset_fn and ble_hs_sync_fn will be extended with extra argument ??

What is so bad about workaround I proposed except it's not the-clean-way-to-do-this? I am not really sure why we should add some extra functions and basically make workarounds in our code...

redradist:
@andrzej-kaczmarek 
> > Would you like this temporary solution as I suggested ?
> > Also I could add comment that in future ble_hs_reset_fn and ble_hs_sync_fn will be extended with extra argument ??
> 
> What is so bad about workaround I proposed except it's not the-clean-way-to-do-this? I am not really sure why we should add some extra functions and basically make workarounds in our code...

And where you think it should be done ?
In fork ? Than all folks will fork from this repository, fix issues, add new feature and this repository could lose contributors (

andrzej-kaczmarek:
> And where you think it should be done ?

In your code. You did not say why this is a bad solution so I assume the only problem is that it's not a clean solution and honestly this is not a good reason to add some quick hacks in our code.

We can make changes in that code and will be happy to accept your contribution, but it has to be well thought out change and most of all it cannot break APIs that probably every single NimBLE app uses.

rymanluk:
@redradist 
Few cents from my side. 
I agree with @ccollins476ad  @andrzej-kaczmarek  that backward compatibility is important.
Solution with additional `_extra_` is no a good one as we would end up with those `_extra_` API, which is not that great. We need some other clean solution which is unknown now.
In the meantime I think that @andrzej-kaczmarek suggestion for you to keep `cb_args` in global `ble_hs_cfg` and use them in your callbacks in the porting layer is fair enough.
Waiting for your contribution. Thanks

PR #698, opened 2019-12-12
Fix MaxRxTime and MaxTxTime when LE Coded is supported

andrzej-kaczmarek:


andrzej-kaczmarek:
@rymanluk this actually does make sense because DLE is pre-5.0 feature so there was no coded PHY and max time was 2120. sending larger value to pre-5.0 device could basically break it so I am pretty sure this is the reason for this requirement.

PR #700, opened 2019-12-13
RFCLK refactor

andrzej-kaczmarek:
This is quite an extensive change which addresses recent problems found with ll_rfclk implementation. Basically, as we added new features it was quite difficult to synchronize rfclk between different pieces of code so there were a lot of small fixes addressing concrete problems. I think this is a good opportunity to refactor this code and make it more robust and easier to follow. With new code it should no longer be a problem to figure out "what and where should I do with rfclk in my code?" since the API is not pretty simple and straightforward and the decision to enable/disable clock is in almost all cases done automagically inside rfclk code.

Main change is 15a3573, however there are some additional changes, mostly cleanups and improvements in scanner which I did when trying to figure out rfclk scenarios. Each change is described in commit message so I will not duplicate this here.

Finally, I decided to change ll_rfclk to ll_rfmgmt since this is meant to manage RF state in general (in case we have more radio drivers in future) and not only XTAL as it does now on Nordic platforms.


nimble/controller/src/ble_ll_dtm.c
```diff
@@ -368,8 +368,7 @@ ble_ll_dtm_tx_create_ctx(uint8_t packet_payload, uint8_t len,
     sch->sched_cb = ble_ll_dtm_tx_sched_cb;
     sch->cb_arg = &g_ble_ll_dtm_ctx;
     sch->sched_type = BLE_LL_SCHED_TYPE_DTM;
-    sch->start_time =  os_cputime_get32() +
-                                       os_cputime_usecs_to_ticks(5000);
+    sch->start_time =  os_cputime_get32() + ble_ll_rfmgmt_enable_now() + 1;
```
sjanc: why not just make ble_ll_rfmgmt_enable_now() do all those and return exact start time? s

nimble/controller/src/ble_ll_rfmgmt.c
```diff
@@ -0,0 +1,324 @@
...
+    g_ble_ll_rfmgmt_data.timer_scheduled = false;
+    os_cputime_timer_stop(&g_ble_ll_rfmgmt_data.timer);
+
+    ble_npl_eventq_remove(&g_ble_ll_data.ll_evq,
+                          &g_ble_ll_rfmgmt_data.release_ev);
+
+    ble_ll_rfmgmt_disable();
+
+    g_ble_ll_rfmgmt_data.enable_scan = false;
+    g_ble_ll_rfmgmt_data.enable_sched = false;
```
sjanc: nitpick: I'd clear all members here, just to make future debugging less confusing

nimble/controller/src/ble_ll_rfmgmt.c
```diff
@@ -0,0 +1,324 @@
...
+ble_ll_rfmgmt_init(void)
+{
+    struct ble_ll_rfmgmt_data *rfmgmt = &g_ble_ll_rfmgmt_data;
+
+    rfmgmt->state = RFMGMT_STATE_OFF;
+
+    rfmgmt->ticks_to_enabled =
+            ble_ll_usecs_to_ticks_round_up(MYNEWT_VAL(BLE_XTAL_SETTLE_TIME));
+
+    rfmgmt->timer_scheduled = false;
```
sjanc: nitpick: this should be already cleared on init

PR #704, opened 2019-12-17
porting: Fix dereference of NULL pointer

yang325:
Assuming 'membuf' is NULL and 'blocks' is equal to 0, the 'block_ptr' in 114 will also be NULL.


andrzej-kaczmarek:
this code is copied from mynewt-core so it needs to be fixed there first and then can be updated here

yang325:
Got it.

yang325:
I found this issue is resolved in mynewt-core. Is there a mechanism to update code from mynewt-core ?

PR #706, opened 2019-12-18
LL network privacy and address resolution fixes

sjanc:


nimble/controller/src/ble_ll_resolv.c
```diff
@@ -744,6 +729,10 @@ ble_ll_resolv_init(void)
                          &g_ble_ll_data.ll_evq,
                          ble_ll_resolv_rpa_timer_cb,
                          NULL);
+
+    /* RPA timer is always running */
+    ble_npl_callout_reset(&g_ble_ll_resolv_data.rpa_timer,
```
rymanluk: maybe we should start to run it when we have first entry on resolving list and disable it when last one is removed ?

PR #707, opened 2019-12-18
nimble/controller: Use mynewt-core crypto driver for certain crypto operations on nrf5x

wes3:
The nimble controller performs certain encryption operations on the nrf5x mcu's by accessing the ECB peripheral directly. We would like to share this peripheral, when possible, with other packages that might want to use the peripheral. Apache Mynewt has crpyto packages that allow sharing the peripheral and these have been added to the controller and will now be used by the controller. Note that some code attempts to access the ECB peripheral inside ISR context (scanning and initiating when resolving private addresses). If the peripheral is being used when address resolution is required, the address resolution will fail (we will not attempt to use the peripheral). This should be a rare occurrence as the peripheral will hopefully not be used extensively nor for long periods of time.


nimble/controller/src/ble_ll_resolv.c
```diff
@@ -138,13 +138,13 @@ ble_ll_resolv_rpa_timer_cb(struct ble_npl_event *ev)
 
     rl = &g_ble_ll_resolv_list[0];
     for (i = 0; i < g_ble_ll_resolv_data.rl_cnt; ++i) {
-        OS_ENTER_CRITICAL(sr);
+        //OS_ENTER_CRITICAL(sr);
```
sjanc: is this not needed anymore?  RPA can be obtained from ISR too so if this happen in the middle of generation we can end up with invalid RPA. Am I missing something here?

andrzej-kaczmarek: and also one thing that I missed before is that if we do not use CRYPTO on nRF52, this critical section is still required. so we should leave it here.

wes3: Ah yes; if we are not using CRYPTO we do need that. So here is my take: this function is called at the LL task so with the new code we will attempt to grab a mutex which we should not do with interrupts disabled. This means I think the enter/exit critical should only be if CRYPTO is not enabled.

@sjanc Yes, address resolution is done within an ISR. If the crypto periperhal is being used at that time we will certainly end up with an invalid RPA. This should not happen often and if it does, we will just have to wait till the next time that particular event occurs and hope the peripheral is not busy. Hopefully that is OK.

andrzej-kaczmarek: so what I do not like is that we would need critical section wrapped in `#if MYNEWT_VAL(CRYPTO)` - in PHY that's perfectly fine, but not so much in LL. I did not look into code too much, but perhaps it's possible that code is modified so critical section is handled in `ble_hw_encrypt_block` instead?

wes3: I believe I have resolved the issues with the latest commit. Take a look and see what y'all think.

CDPA-SCA: Hi, It seems that this commit is not merged into the master branch. I wonder if you are still considering addressing this issue?

PR #716, opened 2020-01-03
Random fixes for issues found by static analysis 

sjanc:


nimble/drivers/nrf52/src/ble_phy.c
```diff
@@ -432,7 +432,7 @@ ble_phy_rxpdu_copy(uint8_t *dptr, struct os_mbuf *rxpdu)
     om = rxpdu;
     dst = om->om_data;
 
-    while (om) {
```
rymanluk: we have same thing for nrf51

PR #719, opened 2020-01-07
Sync Mesh changes and fixes with Zephyr

michal-narajowski:
This PR contains various changes and fixes as well as new features like Provisioner support and Mesh Model extensions.


rymanluk:
@michal-narajowski big set :)  What is the PTS pass rate after that?

michal-narajowski:
@rymanluk This PR doesn't introduce new regressions. There are still some failing test cases after updating PTS to version 7.5.1 and they should be looked into. We still need this PR to be merged so we can test these new features at the UPF.

PR #724, opened 2020-01-08
mesh: Fix GATT Proxy behavior to match Mesh Profile Spec 1.0.1

michal-narajowski:
According to Mesh Profile Spec 1.0.1, Section 4.2.11:
"If the Proxy feature is disabled, a GATT client device can connect
over GATT to that node for configuration and control. Messages from
the GATT bearer are not relayed to the advertising bearer."

Moreover some notes have been removed from the spec compared to
version 1.0:

Mesh Profile Spec 1.0, Section 4.2.11:

"Upon transition from GATT Proxy state 0x01 to GATT Proxy state 0x00
the GATT Bearer Server shall disconnect all GATT Bearer Clients."

"The Configuration Client should turn off the Proxy state as the last
step in the configuration process."

Mesh Profile Spec 1.0, Section 4.2.11.1:

"When the GATT Proxy state is set to 0x00, the Node Identity state
for all subnets shall be set to 0x00 and shall not be changed."

### DIFF
--- a/nimble/host/mesh/src/net.c
+++ b/nimble/host/mesh/src/net.c
@@ -1325,13 +1325,19 @@ void bt_mesh_net_recv(struct os_mbuf *data, s8_t rssi,
 	/* Save the state so the buffer can later be relayed */
 	net_buf_simple_save(buf, &state);
 
+	rx.local_match = (bt_mesh_fixed_group_match(rx.ctx.recv_dst) ||
+			  bt_mesh_elem_find(rx.ctx.recv_dst));
+
 	if ((MYNEWT_VAL(BLE_MESH_GATT_PROXY)) &&
 	    net_if == BT_MESH_NET_IF_PROXY) {
 		bt_mesh_proxy_addr_add(data, rx.ctx.addr);
-	}
 
-	rx.local_match = (bt_mesh_fixed_group_match(rx.ctx.recv_dst) ||
-			  bt_mesh_elem_find(rx.ctx.recv_dst));
+		if (bt_mesh_gatt_proxy_get() == BT_MESH_GATT_PROXY_DISABLED &&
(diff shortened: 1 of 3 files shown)


nimble/host/mesh/src/net.c
```diff
@@ -1325,13 +1325,19 @@ void bt_mesh_net_recv(struct os_mbuf *data, s8_t rssi,
...
+			  bt_mesh_elem_find(rx.ctx.recv_dst));
+
 	if ((MYNEWT_VAL(BLE_MESH_GATT_PROXY)) &&
 	    net_if == BT_MESH_NET_IF_PROXY) {
 		bt_mesh_proxy_addr_add(data, rx.ctx.addr);
-	}
 
-	rx.local_match = (bt_mesh_fixed_group_match(rx.ctx.recv_dst) ||
-			  bt_mesh_elem_find(rx.ctx.recv_dst));
+		if (bt_mesh_gatt_proxy_get() == BT_MESH_GATT_PROXY_DISABLED &&
```
michal-narajowski: "If the Proxy feature is disabled, a GATT client device can connect
over GATT to that node for configuration and control. Messages from
the GATT bearer are not relayed to the advertising bearer."

I'm not 100% sure if this means we should drop messages that are not directed at us or that we just shouldn't relay messages to ADV (see `bt_mesh_net_relay()` call below.

michal-narajowski:
@jhedberg

jhedberg:
@trond-snekvik we'll need to check if these fixes are applicable to Zephyr

LingaoM:
@michal-narajowski @rymanluk  CC https://github.com/zephyrproject-rtos/zephyr/pull/29217

michal-narajowski:
Thanks, @LingaoM

PR #726, opened 2020-01-09
[DO NOT MERGE] host: Don't allow unauth pairing if MITM protection required

ccollins476ad:
DO NOT MERGE: This is not the correct fix.  If I apply the same fix to `ble_sm_lgcy.c`, some unit tests fail.

I think this PR identifies a real bug, but I am not sure if the fix conforms to the letter of the Bluetooth spec.  Hopefully someone more knowledgeable can take a look.

Before this PR, the following scenario was possible:

1. We send pair request with mitm=1
2. Peer ignores our mitm flag and sends a pair response attempting just works.
3. We accept the pair response and proceed with just works pairing.

This commit changes step 3 such that we abort the pairing procedure if either side has mitm=1 but the pairing algorithm is unauthenticated.

### DIFF
--- a/nimble/host/src/ble_sm_sc.c
+++ b/nimble/host/src/ble_sm_sc.c
@@ -100,16 +100,18 @@ int
 ble_sm_sc_io_action(struct ble_sm_proc *proc, uint8_t *action)
 {
     struct ble_sm_pair_cmd *pair_req, *pair_rsp;
+    bool any_mitm;
 
     pair_req = (struct ble_sm_pair_cmd *) &proc->pair_req[1];
     pair_rsp = (struct ble_sm_pair_cmd *) &proc->pair_rsp[1];
 
+    any_mitm = (pair_req->authreq & BLE_SM_PAIR_AUTHREQ_MITM) ||
+               (pair_rsp->authreq & BLE_SM_PAIR_AUTHREQ_MITM);
+
     if (pair_req->oob_data_flag == BLE_SM_PAIR_OOB_YES ||
         pair_rsp->oob_data_flag == BLE_SM_PAIR_OOB_YES) {
         *action = BLE_SM_IOACT_OOB_SC;
-    } else if (!(pair_req->authreq & BLE_SM_PAIR_AUTHREQ_MITM) &&
-               !(pair_rsp->authreq & BLE_SM_PAIR_AUTHREQ_MITM)) {
-
+    } else if (!any_mitm) {
         *action = BLE_SM_IOACT_NONE;
     } else if (pair_req->io_cap >= BLE_SM_IO_CAP_RESERVED ||
                pair_rsp->io_cap >= BLE_SM_IO_CAP_RESERVED) {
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
This is partially correct.

We can set our mitm=1 if we either mandate MITM or if we allow to downgrade to mitm=0 and proceed with unauthenticated pairing - this is up to implementation and we do the latter. However, if remote device set mitm=1 we cannot assume that it will not accept pairing without MITM.

That said, we either keep current behavior or add some setting to control our MITM policy (either separate setting or just say mitm=2 means we require MITM). I'd say current behavior is ok since in case we have unauthenticated pairing subsequent access to GATT database which requires authenticated key will be rejected or application can simply disconnect link if it does not want unauthenticated link.

The relevant test case which verifies this behavior is GAP/SEC/AUT/BV-17-C. It has 2 alternatives depending on mentioned policy of IUT. We now pass this test case because we set IXIT_iut_mandates_mitm=0. After changing the behavior we would need to also adjust this settings in PTS.

rymanluk:
@ccollins476ad I agree with @andrzej-kaczmarek . Out of curiosity, what is your use case?

ccollins476ad:
The use case is this:

A device ships locked.  The user can only unlock the device by pairing with it.  Pairing requires a passkey that the user gets from the vendor web site after logging in and authenticating.

The application sets mitm=1.  I assumed this meant that JustWorks pairing would never occur.  The application unlocks the device as soon as the connection changes its state to "encrypted".

I don't see a means for the application to abort the pairing procedure when the peer indicates that it will be using an unauthenticated algorithm.  The application has to allow pairing to succeed, but then it can inspect the `authenticated` flag to determine if it should unlock the device, is this correct?  I hadn't noticed the `authenticated` member of `struct ble_gap_sec_state`!

ccollins476ad:
I think the fix I described in my last comment works, so we can close this PR.

I wonder if there is a user-friendliness issue here though (I have no idea what the solution is).  It might just be me, but I don't think it's obvious that the stack allows unauthenticated pairing when it is configured to use authentication.  In my mind if you configure the stack to use authentication, you want to reject unauthenticated attempts (otherwise why even allow authentication?).

That alone isn't too bad, but my real concern is that it is easy to overlook such a security vulnerability.  Most Bluetooth devices won't attempt JustWorks if the nimble device sets `mitm=1`, so this problem never gets detected.

Again, I don't know what the solution is.  Maybe just a big disclaimer in the documentation about it being the application's responsibilty to check the authenticated state if it cares about it?

andrzej-kaczmarek:
Correct, you can check `authenticated` flag to see if link is authenticated and unlock device. Also if at least one device set mitm=1, then pairing method is determined by IO capabilities so assuming you pair with a phone which should have `KeyboardDisplay`, setting `DisplayOnly` on your side should always trigger Passkey Entry pairing which results in an authenticated key. If remote device forced Just Works, you should be able to force MITM by initiating pairing once more, as per spec (Core 5.1, Vol 3, Part C, 10.3.2):

> If an authenticated pairing is required but only an unauthenticated pairing has occurred and the link is currently encrypted, the pairing procedure should be executed with the required authentication settings. If the pairing procedure fails, or an authenticated pairing cannot be performed with the IO capabilities of the local device and remote device, then the service request shall be aborted.

I do not think there is an issue here. IMO it's more user-friendly that Just Works is triggered if service does not require authenticated link. If it does, authenticated pairing should be triggered automatically. In your case it looks like an issue because as I understand you do not need to access any particular service and just want to trigger certain pairing method - that is not really an use-case for LE link in GAP spec since it defines behavior for either initiating local or remote service, i.e. accessing L2CAP channel or GATT service.

ccollins476ad:
Thanks for the clarification @andrzej-kaczmarek.

PR #727, opened 2020-01-10
nimble/ll: Fix setting anchor point and WFR for sync PDUs

sjanc:
Don't set anchor point in the middle of expected window for first
packet. Now we set anchor to calculated point in time and adjust
WFR with window widenning (and offset if synchronizing).


sjanc:
@haukepetersen   seems like riot port is broken with latest master, could you check this?

haukepetersen:
@sjanc did read my notifications only now: RIOT should be fixed with 37245d0c3f239c35c0c734a337c7b6301fcfd70b

PR #730, opened 2020-01-13
nimble/host: Fix setting connection flags after pairing

andrzej-kaczmarek:


dmastain:
@andrzej-kaczmarek , I pulled this branch and re-built my btshell application with your changes. The flags reported in the console are looking much better with this change, thanks!

Issue #712, opened 2020-01-01
A case in ble_ll_sched_adv_reschedule


JaydenH215:
If there is a adv event sched item:
1. sch->start_time < head sched item's start_time in queue.
2. sch->end_time > head sched item's start_time in queue.

After following code:
`                    sch->start_time = entry->end_time;
                    sch->end_time = sch->start_time + duration;`

The adv event may overlap the next_sch ?

wnnwoo:
Yes, it may overlap the next_sch

and the procedure will judge whether it is overlapped it in the next iteration in this while loop, the procedure will go and find the right position until the end of the Q.

sjanc:
Hi,

Since our documentation needs improvements, it would be great if you consider updating it with your notes and send PR :-)

JaydenH215:
> Yes, it may overlap the next_sch
> 
> and the procedure will judge whether it is overlapped it in the next iteration in this while loop, the procedure will go and find the right position until the end of the Q.
> 
> U can refer this doc.
> https://wnnwoo.github.io/2020/01/05/nimble-scheduler/

```
    if (entry == end_overlap) {
        rand_ticks = (orig_start + max_delay_ticks) - sch->start_time;
        if (rand_ticks > max_delay_ticks) {
            /* No place for advertisement. */
            rc = -1;
        } else {
            if (next_sch == NULL) {
                TAILQ_INSERT_TAIL(&g_ble_ll_sched_q, sch, link);
            } else {
                TAILQ_INSERT_BEFORE(next_sch, sch, link);
            }
        }
        break;
    }
```

But the latter condition should be true, it will 'break' and will not enter the while again.
I have read your blog and thank you for sharing. I am also learning nimble and hope to make friends with you. 

[https://jaydenh215.github.io/](https://jaydenh215.github.io/)

wnnwoo:
Dose my blog resole ur question?

sjanc:
reopen if this is stil an issue

PR #731, opened 2020-01-15
update ble_ll_sched_adv_reschedule

JaydenH215:
Hi, 

I have created a [issue](https://github.com/apache/mynewt-nimble/issues/712).In fact, I can't really understand how the author's scheduling mechanism works. If anyone could explain it, it will be great.

Q1: Issue mentioned above 
Q2: How do we ensure that the start_time of item inserted before end_overlap is less than max_delay.

Thanks

### DIFF
--- a/nimble/controller/src/ble_ll_sched.c
+++ b/nimble/controller/src/ble_ll_sched.c
@@ -914,6 +914,9 @@ ble_ll_sched_adv_reschedule(struct ble_ll_sched_item *sch, uint32_t *start,
     rand_ticks = max_delay_ticks;
     sch->end_time += max_delay_ticks;
 
+    /* Mark original schedule item 'start' */
+    orig_start = sch->start_time;
+
     start_overlap = NULL;
     end_overlap = NULL;
     before = NULL;
@@ -932,6 +935,8 @@ ble_ll_sched_adv_reschedule(struct ble_ll_sched_item *sch, uint32_t *start,
                 } else {
                     end_overlap = entry;
                 }
+                sch->start_time = entry->end_time;
+                sch->end_time = sch->start_time + max_delay_ticks;
             } else {
                 if ((int32_t)(sch->end_time - entry->start_time) <= 0) {
                     before = entry;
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
if you make such change, the commit title shall have short summary of change and elaborate message describing the problem and then fix.

nimble/controller/src/ble_ll_sched.c
```diff
@@ -961,20 +966,25 @@ ble_ll_sched_adv_reschedule(struct ble_ll_sched_item *sch, uint32_t *start,
...
              */
             before = NULL;
-            orig_start = sch->start_time;
+            sch->start_time = orig_start;
             entry = start_overlap;
             sch->end_time = sch->start_time + duration;
             while (1) {
                 next_sch = entry->link.tqe_next;
                 if ((int32_t)(sch->end_time - entry->start_time) <= 0) {
-                    rand_ticks = entry->start_time - sch->end_time;
```
andrzej-kaczmarek: when checking for overlapping events we had `sch->end_time = orig_end_time + max_delay_ticks` so for any `entry` in this loop we have `entry->start_time <= orig_end_time + max_delay_ticks`. this expression translates to `rand_ticks = entry->start_time - orig_end_time` (because we restored `sch->end_time` to original value) so by combining the two expressions together we have `rand_ticks <= max_delay_ticks` which is what we need.

nimble/controller/src/ble_ll_sched.c
```diff
@@ -961,20 +966,25 @@ ble_ll_sched_adv_reschedule(struct ble_ll_sched_item *sch, uint32_t *start,
...
-                    before = entry;
-                    TAILQ_INSERT_BEFORE(before, sch, link);
-                    break;
-                } else {
-                    sch->start_time = entry->end_time;
-                    sch->end_time = sch->start_time + duration;
```
andrzej-kaczmarek: here we can indeed move `sch->start_time` beyond `orig_start_time + max_delay_ticks` but this can only happen for last overlapping event which is handled in the condition below.

nimble/controller/src/ble_ll_sched.c
wnnwoo: understood.

nimble/controller/src/ble_ll_sched.c
wnnwoo: but there is issue in the handle for the below condition.
The sch may overlap next_sch if next_sch is not NULL **if (sch->end_time >next_sch->start_time)**
```
                    } else {
                        if (next_sch == NULL) {
                            TAILQ_INSERT_TAIL(&g_ble_ll_sched_q, sch, link);
                        } else {
                            TAILQ_INSERT_BEFORE(next_sch, sch, link);
                        }
                    }
```

JaydenH215: I think
`rand_ticks = (orig_start + max_delay_ticks) - sch->start_time;  ` 
equal to 
`rand_ticks = (orig_end + max_delay_ticks) - sch->end_time;  `

If we can insert sch into link, this means  `sch->end_time <= orig_end + max_delay_ticks`, so we can make sure sch will not overflow(max_delay). If `sch` overlaps `next_sch`, `end_overlap` should be `next_sch` (sch->end_time <= orig_end + max_delay < end_overlap->next_sch->start_time).

andrzej-kaczmarek: `sch->end_time > next_sch->start_time` is impossible, because `next_sch` is the next element after `entry` (see line 968). `entry` is the last element overlapping `sch` so `next_sch` cannot overlap `sch`.

wnnwoo: understood.

PR #733, opened 2020-01-15
nimble/ll: Fix connection create if CONNECT_IND was not txed

andrzej-kaczmarek:
Consider following scenario:
- we are receiving ADV_IND at the end of scan window
- scan timer fires in the middle of that PDU and we are handling
  scan proc to switch window
- in the meantime ADV_IND is received and we decide to transmit
  CONNECT_IND
- CONNECT_IND is setup for TX, but then we are back handling window
  switch which decides to disable PHY since it already checked that
  it should move to next window; this effectively stops TX and
  CONNECT_IND in the air is truncated (i.e. corrupted)
- ADV_IND is handed over to LL which assumes, according to flags, that
  CONNECT_IND was transmitted so it should start connection.

And here's the problem: we did not send proper CONNECT_IND so we should
not start connection, because peer does not really expects it.

The solution here is to delay connection on legacy adv until CONNECT_IND
was transmitted completly. Basically, when processing ADV_IND for which
we sent CONNECT_IND (according to flags set in isr) we setup conns as
usual, but we do not start connection. Instead, this will be done from
separate event which is sent from txpdu_cb which means PDU was sent
successfully. The same event is also used for AUX_CONNECT_RSP for
consistency. At the time we receive such event we know that connsm is
aready setup properly so we just need to stop scanning and create
connection.

### DIFF
--- a/nimble/controller/src/ble_ll_conn.c
+++ b/nimble/controller/src/ble_ll_conn.c
@@ -135,6 +135,8 @@ struct ble_ll_empty_pdu
 /* Global connection complete event. Used when initiating */
 uint8_t *g_ble_ll_conn_comp_ev;
 
+static struct ble_npl_event g_ble_ll_init_conn_created_ev;
+
 /* Global LL connection parameters */
 struct ble_ll_conn_global_params g_ble_ll_conn_params;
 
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
this does not fix the problem properly and introduces some side-effects, need to think more about it...

PR #735, opened 2020-01-21
nimble/ll: Fix HCI checks for periodic advertising

sjanc:
Core Specification 5.2 clarified HCI behavior for periodic advertising
with regards to reconfiguring advertising instance and checking
data length against periodic advertising interval.

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -3222,27 +3222,20 @@ ble_ll_adv_ext_set_param(const uint8_t *cmdbuf, uint8_t len,
     props = le16toh(cmd->props);
 
 #if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_PERIODIC_ADV)
-    /* For now just disallow to reconfigure instance that has periodic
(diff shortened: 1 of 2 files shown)


nimble/controller/src/ble_ll_adv.c
```diff
@@ -3605,7 +3598,35 @@ ble_ll_adv_clear_all(void)
...
+    while (payload_len > BLE_PHY_MAX_PDU_LEN) {
+        max_usecs += chain_usecs;
+        payload_len -= BLE_PHY_MAX_PDU_LEN;
+    }
+
+    max_usecs += ble_ll_pdu_tx_time_get(payload_len, advsm->sec_phy);
+
+    itvl_usecs = (uint32_t)itvl * BLE_LL_ADV_PERIODIC_ITVL;
+
+    return max_usecs < itvl_usecs;
```
andrzej-kaczmarek: `max_usecs` is not calculated properly. `payload_len` is just length of data, but the actual payload of each PDU includes also ext adv header so there's more data to send thus it will take more usecs.

PR #737, opened 2020-01-23
nimble/ll: Use S=8 coding by default

andrzej-kaczmarek:
It makes more sense to use S=8 by default as we can assume that LE Coded
is used for Long Range so we should use best possible coding scheme at
the expense of data speed.

### DIFF
--- a/nimble/controller/include/controller/ble_phy.h
+++ b/nimble/controller/include/controller/ble_phy.h
@@ -179,17 +179,15 @@ void ble_phy_resolv_list_enable(void);
 void ble_phy_resolv_list_disable(void);
 
 /*
- * These definitions are used for the 'phy' parameters in the API listed below.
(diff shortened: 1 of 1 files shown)


nimble/controller/include/controller/ble_phy.h
```diff
@@ -218,16 +216,14 @@ static inline int ble_ll_phy_to_phy_mode(int phy, int phy_options)
     int phy_mode;
 
     /*
-     * Mode values are set in a way that 1M, 2M and Coded(S=2) are equivalent
-     * to 1M, 2M and Coded in HCI. The only conversion is needed for Coded(S=8)
-     * which uses non-HCI value.
+     * 'phy' value can be used as 'phy_mode' value unless S=2 coding is explicitly
+     * required. By default we'll use S=2 for Coded.
```
sjanc: S=8 ?

PR #740, opened 2020-01-24
nimble/ll: Add possible to set BT v5.2 flag plus some minor changes.

rymanluk:


nimble/controller/src/ble_ll.c
```diff
@@ -1277,6 +1280,46 @@ ble_ll_read_supp_features(void)
...
 
+/**
+ * Sets the features controlled by the host.
+ *
+ * @return HCI command status
+ */
+int
+ble_ll_set_host_feat(const uint8_t *cmdbuf, uint8_t len)
+{
+#if MYNEWT_VAL(BLE_VERSION) >= 52
```
sjanc: this is not needed since caller is already doing #ifdef foo

nimble/controller/src/ble_ll.c
```diff
@@ -1277,6 +1280,46 @@ ble_ll_read_supp_features(void)
...
+    const struct ble_hci_le_set_host_feat_cp *cmd = (const void *) cmdbuf;
```
sjanc: you also need to check if len == sizeof(*cmd) here

nimble/controller/include/controller/ble_ll.h
```diff
@@ -199,44 +199,52 @@ extern STATS_SECT_DECL(ble_ll_stats) ble_ll_stats;
...
  * but host will want to use this procedure, we will try. If not
  * succeed, feature bit will be cleared.
  * Look at LL Features above to find out what is allowed
  */
 #define BLE_LL_CONN_INITIAL_FEATURES    (0x00000022)
-
 #define BLE_LL_CONN_CLEAR_FEATURE(connsm, feature)   (connsm->conn_features &= ~(feature))
 
+/* All the features which can be controlled by the Host */
+#define BLE_LL_HOST_CONTROLLED_FEATURES (BLE_LL_ISO_HOST_SUPPORT)
```
sjanc: I think this should be 0 until some initial ISO implementation lands in

nimble/controller/include/controller/ble_ll.h
```diff
@@ -199,44 +199,52 @@ extern STATS_SECT_DECL(ble_ll_stats) ble_ll_stats;
...
+#define BLE_LL_FEAT_CTE_TX           (0x0000080000)
+#define BLE_LL_FEAT_CTE_RX           (0x0000100000)
+#define BLE_LL_FEAT_CTE_AOD          (0x0000200000)
+#define BLE_LL_FEAT_CTE_AOA          (0x0000400000)
+#define BLE_LL_FEAT_CTE_RECV         (0x0000800000)
+#define BLE_LL_FEAT_SYNC_SEND        (0x0001000000)
+#define BLE_LL_FEAT_SYNC_RECV        (0x0002000000)
+#define BLE_LL_FEAT_SCA_UPDATE       (0x0004000000)
+#define BLE_LL_FEAT_REM_PKEY         (0x0008000000)
+#define BLE_LL_FEAT_CTE_RX           (0x0000100000)
```
andrzej-kaczmarek: duplicated def

nimble/controller/include/controller/ble_ll.h
rymanluk: This is as per spec. Where we say that our controller supports it or not is in different place: 
``/** Our supported features which can be controller by the host */
uint64_t g_ble_ll_supported_host_features = 0;``

nimble/controller/include/controller/ble_ll.h
rymanluk: thx

nimble/controller/src/ble_ll.c
rymanluk: ah you are right

nimble/controller/src/ble_ll.c
rymanluk: thx

nimble/include/nimble/hci_common.h
```diff
@@ -825,6 +825,13 @@ struct ble_hci_le_set_default_periodic_sync_transfer_params_cp {
 #define BLE_HCI_OCF_LE_GENERATE_DHKEY_V2                 (0x005E)
 #define BLE_HCI_OCF_LE_MODIFY_SCA                        (0x005F)
 
+#define BLE_HCI_OCF_LE_SET_HOST_FEAT                     (0x0074)
+struct ble_hci_le_set_host_feat_cp
+{
```
sjanc: this is not following coding style

PR #741, opened 2020-01-24
Set bleprph main task stack size to 512.

mlaz:
When using mcumgr's image management uploads on this app at least 502 words would be needed for stack. This is a similar problem to the one described on https://github.com/apache/mynewt-mcumgr/pull/59 so I came up with a value slightly higher than what was measured.

### DIFF
--- a/apps/bleprph/syscfg.yml
+++ b/apps/bleprph/syscfg.yml
@@ -58,7 +58,7 @@ syscfg.vals:
     CONFIG_MGMT: 1
 
     # OS main/default task
-    OS_MAIN_STACK_SIZE: 468
+    OS_MAIN_STACK_SIZE: 512
 
     # Lots of smaller mbufs are required for smp using typical BLE ATT MTU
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek:
would be actually better if someone check why mcumgr consumes so much stack instead of just increasing stack size...

vrahane:
@andrzej-kaczmarek We figured out already :-) there was a related discussion on the mcumgr slack channel, so there is a 512 byte flat buffer which gets allocated on the stack for the upload request. we decided to keep the flat buf as it is on the stack since putting it in the bss or making it a global would be wasting memory all the time and mbufs would mean adding APIs for making the upload requests OS agnostic. You can follow the mcumgr slack channel for the discussion.

andrzej-kaczmarek:
so this should be explained in commit message

mlaz:
> so this should be explained in commit message

Commit message updated.

PR #753, opened 2020-02-10
npl/riot: Improve timer glue code

bergzand:
This PR modifies and cleans the RIOT NPL glue code. 

The first commit replaces the constants in the milllisecond-to-microsecond and microsecond-to-millisecond conversions to the RIOT defines for this to improve readability.
The second commit mitigates an incorrect `remaining_ticks` result when the RIOT microsecond timer overflows or when the Nimble thread is interrupted between setting the `target_ticks` and setting the timer.

@haukepetersen This might be of interest to you.

### DIFF
--- a/porting/npl/riot/include/nimble/nimble_npl_os.h
+++ b/porting/npl/riot/include/nimble/nimble_npl_os.h
@@ -100,7 +100,8 @@ ble_npl_eventq_get(struct ble_npl_eventq *evq, ble_npl_time_t tmo)
     } else if (tmo == BLE_NPL_TIME_FOREVER) {
         return (struct ble_npl_event *)event_wait(&evq->q);
     } else {
-        return (struct ble_npl_event *)event_wait_timeout(&evq->q, (tmo * 1000));
+        return (struct ble_npl_event *)event_wait_timeout(&evq->q,
+                                                          (tmo * US_PER_MS));
     }
 }
 
@@ -227,7 +228,7 @@ ble_npl_callout_set_arg(struct ble_npl_callout *co, void *arg)
 static inline uint32_t
 ble_npl_time_get(void)
 {
-    return xtimer_now_usec() / 1000;
+    return xtimer_now_usec64() / US_PER_MS;
 }
 
 static inline ble_npl_error_t
@@ -259,7 +260,7 @@ ble_npl_time_ticks_to_ms32(ble_npl_time_t ticks)
 static inline void
 ble_npl_time_delay(ble_npl_time_t ticks)
 {
-    xtimer_usleep(ticks * 1000);
+    xtimer_usleep(ticks * US_PER_MS);
 }
 
 static inline uint32_t
--- a/porting/npl/riot/include/nimble/nimble_npl_os.h
+++ b/porting/npl/riot/include/nimble/nimble_npl_os.h
@@ -51,7 +51,7 @@ struct ble_npl_eventq {
 
 struct ble_npl_callout {
     xtimer_t timer;
-    ble_npl_time_t target_ticks;
+    uint64_t target_us;
     struct ble_npl_event e;
     event_queue_t *q;
 };
@@ -100,8 +100,8 @@ ble_npl_eventq_get(struct ble_npl_eventq *evq, ble_npl_time_t tmo)
     } else if (tmo == BLE_NPL_TIME_FOREVER) {
         return (struct ble_npl_event *)event_wait(&evq->q);
     } else {
-        return (struct ble_npl_event *)event_wait_timeout(&evq->q,
-                                                          (tmo * US_PER_MS));
+        return (struct ble_npl_event *)event_wait_timeout64(&evq->q,
+                                                            tmo * US_PER_MS);
     }
 }
 
@@ -216,7 +216,7 @@ ble_npl_callout_is_active(struct ble_npl_callout *c)
 static inline ble_npl_time_t
 ble_npl_callout_get_ticks(struct ble_npl_callout *co)
 {
-    return co->target_ticks;
+    return (ble_npl_time_t)(co->target_us / US_PER_MS);
 }
 
 static inline void
@@ -225,10 +225,10 @@ ble_npl_callout_set_arg(struct ble_npl_callout *co, void *arg)
     co->e.arg = arg;
 }
 
-static inline uint32_t
+static inline ble_npl_time_t
 ble_npl_time_get(void)
 {
-    return xtimer_now_usec64() / US_PER_MS;
+    return (ble_npl_time_t)(xtimer_now_usec64() / US_PER_MS);
 }
 
 static inline ble_npl_error_t
@@ -260,7 +260,7 @@ ble_npl_time_ticks_to_ms32(ble_npl_time_t ticks)
 static inline void
 ble_npl_time_delay(ble_npl_time_t ticks)
 {
-    xtimer_usleep(ticks * US_PER_MS);
+    xtimer_usleep64(ticks * US_PER_MS);
 }
 
 static inline uint32_t
--- a/porting/npl/riot/src/npl_os_riot.c
+++ b/porting/npl/riot/src/npl_os_riot.c
@@ -39,7 +39,8 @@ ble_npl_sem_pend(struct ble_npl_sem *sem, ble_npl_time_t timeout)
     struct timespec abs;
     uint64_t time;
 
-    time = xtimer_now_usec64() + ble_npl_time_ticks_to_ms32(timeout) * US_PER_MS;
+    time = xtimer_now_usec64() +
+        (ble_npl_time_ticks_to_ms32(timeout) * US_PER_MS);
     abs.tv_sec = (time_t)(time / US_PER_SEC);
     abs.tv_nsec = (long)((time % US_PER_SEC) * NS_PER_US);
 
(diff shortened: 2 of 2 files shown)


porting/npl/riot/include/nimble/nimble_npl_os.h
```diff
@@ -215,7 +215,7 @@ ble_npl_callout_is_active(struct ble_npl_callout *c)
 static inline ble_npl_time_t
 ble_npl_callout_get_ticks(struct ble_npl_callout *co)
 {
-    return co->target_ticks;
+    return (ble_npl_time_t)co->target_ticks;
```
sjanc: this looks a bit odd,  shouldn't ble_npl_time_t in riot port be change to uint64_t ?

bergzand: I agree that it is odd and there is a bug here.

I don't think it is necessary to change `ble_npl_time_t` to `uint64_t`. The main reason to change from `uint32_t` to `uint64_t` is in the places where NimBLE provides a time in milliseconds and RIOT expects it in microseconds. The conversion to microseconds should be paired with a conversion to `uint64_t` to compensate for the additional range provided by value as millisecond compared to microseconds.

The bug is this line is that `target_ticks` is not in milliseconds but in microseconds. The RIOT port specifies that [1 tick == 1 millisecond](https://github.com/apache/mynewt-nimble/blob/master/porting/npl/riot/include/nimble/nimble_npl_os.h#L248), so this function should also properly return a value in milliseconds and not in microseconds.

I prefer to stick to `uint32_t` for the `ble_npl_time_t` and make sure that values converted from microseconds to milliseconds wrap around at `2**32` milliseconds and not at `2**32/1000` milliseconds

sjanc: fair enough,  I'll leave final call to Hauke though :)

haukepetersen:
@bergzand yes, this is indeed interesting :-)

Changes are valid and the code looks good to me, testing it as we speak.

porting/npl/riot/include/nimble/nimble_npl_os.h
```diff
@@ -100,7 +100,8 @@ ble_npl_eventq_get(struct ble_npl_eventq *evq, ble_npl_time_t tmo)
     } else if (tmo == BLE_NPL_TIME_FOREVER) {
         return (struct ble_npl_event *)event_wait(&evq->q);
     } else {
-        return (struct ble_npl_event *)event_wait_timeout(&evq->q, (tmo * 1000));
+        return (struct ble_npl_event *)event_wait_timeout(&evq->q,
+                                                          (tmo * US_PER_MS));
```
haukepetersen: If I am not mistaken, a source for error remains for this call: if `tmo`s value is greater than (UINT32T_MAX / 1000 -> 1,19 hours), the timeout will be set incorrectly. Looking at RIOTs µs-based timer APIs it seems this can not be so easily solved. But how about we a) put in an assert that triggers once this overflow happens, or b) return NULL in that case, which subsequently would also in most cases trigger an assertion by the calling code...

bergzand: I have to admit that I intentionally ignored the edge case here, as you mention, there is no easy way to solve the edge case here.

haukepetersen:
One more: the overflow issues is also present in `ble_npl_time_delay()`.

Thinking about his a little more, I think that we should actually expose fitting functions in RIOT so that we can cleanly map the effected functions:
- expose `xtimer_usleep64()` in RIOT and use it to map `ble_npl_time_delay()` properly
- add `xtimer_set_timeout_flag64()` and `event_wait_timeout64` to the corresponding RIOT APIs and use them in `ble_npl_eventq_get()`.

@bergzand what do you think?

porting/npl/riot/include/nimble/nimble_npl_os.h
bergzand: I'm not that familiar with Bluetooth timings, do you consider it likely that this assertion will trigger during the lifetime of a node? Are there any timing durations in the Bluetooth spec itself that are longer than this 1.19 hours?

bergzand:
> One more: the overflow issues is also present in `ble_npl_time_delay()`.

These should show up only when the time in milliseconds causes an overflow in the `uint32_t` microsecond representation. (right?) Whereas with the `ble_npl_callout_remaining_ticks` it will always cause issues when the microsecond representation is near enough to `UINT32_MAX`.

I think adding assertions to `ble_npl_time_delay()` and `ble_npl_eventq_get` should be sufficient for now, assuming Nimble doesn't require timeouts longer than 1.19 hours in regular operations.

To mitigate this with a more long-term solution I think we either have to add the functions you've proposed to RIOT or merge and migrate to [ztimer](https://github.com/RIOT-OS/RIOT/pull/11874/files). ztimer allows for setting millisecond-based timers and would mitigate most of the issues present here.

haukepetersen:
> assuming Nimble doesn't require timeouts longer than 1.19 hours in regular operations.

for normal operation I don't think so, but I do not know for sure. I guess the assertions will telll...

I agree, that `ztimer` shoud be the mid-term goal. However, at least `event_wait_timeout()` would still prove to be a small challenge.

So how about:
1. catch overflows using asserts in `ble_npl_time_delay()` and `ble_npl_eventq_get()` for now, so we can get this PR merged and benefit from the fixes it provides
2. add API functions in RIOT to `xtimer` and `event` as intermediate
3. remove asserts from (1.) in favor of new functions
4. move to `ztimer` in general, once `ztimer` has stabilized a little further

bergzand:
>     1. catch overflows using asserts in `ble_npl_time_delay()` and `ble_npl_eventq_get()` for now, so we can get this PR merged and benefit from the fixes it provides

Done

bergzand:
And I think I broke something related to the announcements, I'll look into it

haukepetersen:
Nice, I am just looking into doing the needed adaptions in RIOT for step 2...

porting/npl/riot/include/nimble/nimble_npl_os.h
```diff
@@ -100,7 +100,10 @@ ble_npl_eventq_get(struct ble_npl_eventq *evq, ble_npl_time_t tmo)
     } else if (tmo == BLE_NPL_TIME_FOREVER) {
         return (struct ble_npl_event *)event_wait(&evq->q);
     } else {
-        return (struct ble_npl_event *)event_wait_timeout(&evq->q, (tmo * 1000));
+        uint64_t tmo_us64 = tmo * US_PER_MS;
+        assert(tmo_us64 <= UINT32_MAX);
+        return (struct ble_npl_event *)event_wait_timeout(&evq->q,
+                                                          (tmo_us64));
```
haukepetersen: in the current state, `tmo_us64` should be explicitly casted to `uint32_t` I'd say

bergzand:
> Nice, I am just looking into doing the needed adaptions in RIOT for step 2...

Thanks!

I think my code is fine, but somehow the RTT is broken on my test board. On a nRF52dk everything seems fine I receive the BLE advertisements.

porting/npl/riot/include/nimble/nimble_npl_os.h
```diff
@@ -259,7 +262,9 @@ ble_npl_time_ticks_to_ms32(ble_npl_time_t ticks)
 static inline void
 ble_npl_time_delay(ble_npl_time_t ticks)
 {
-    xtimer_usleep(ticks * 1000);
+    uint64_t us64 = ticks * US_PER_MS;
+    assert(us64 <= UINT32_MAX);
+    xtimer_usleep(us64);
```
haukepetersen: getting RIOT ready was easier than expected: https://github.com/RIOT-OS/RIOT/pull/13370

So we can drop the assert again and use `xtimer_usleep64()` directly here.

porting/npl/riot/include/nimble/nimble_npl_os.h
haukepetersen: or we simply wait for https://github.com/RIOT-OS/RIOT/pull/13373 to be merged :-)

bergzand: Waited for RIOT-OS/RIOT#13373 and modified to `event_wait_timeout64`

porting/npl/riot/include/nimble/nimble_npl_os.h
bergzand: > So we can drop the assert again and use `xtimer_usleep64()` directly here.

Done!

porting/npl/riot/include/nimble/nimble_npl_os.h
```diff
@@ -227,7 +228,7 @@ ble_npl_callout_set_arg(struct ble_npl_callout *co, void *arg)
 static inline uint32_t
 ble_npl_time_get(void)
 {
-    return xtimer_now_usec() / 1000;
+    return xtimer_now_usec64() / US_PER_MS;
```
haukepetersen: just notice something that could be fixed while we are at it: `ble_npl_time_get()` should actually return a value of type `ble_npl_time_t` (see `nimble/include/nimble/nimble_npl.h` line 139), so we could just adapt the function declaration in this PR.

Furthre: it would be cleaner looking if the return value would be explicitly casted to `ble_npl_time_t`
```c
return (ble_npl_time_t)(xtimer_now_usec64() / US_PER_MS);

File: porting/npl/riot/src/npl_os_riot.c [hunk 6]
```diff
@@ -39,9 +39,9 @@ ble_npl_sem_pend(struct ble_npl_sem *sem, ble_npl_time_t timeout)
     struct timespec abs;
     uint64_t time;
 
-    time = xtimer_now_usec64() + ble_npl_time_ticks_to_ms32(timeout) * 1000;
-    abs.tv_sec = (time_t)(time / 1000000);
-    abs.tv_nsec = (long)((time % 1000000) * 1000);
+    time = xtimer_now_usec64() + ble_npl_time_ticks_to_ms32(timeout) * US_PER_MS;
```
haukepetersen: nitpick: parenthesis would help readability here :-)

bergzand:
I have 09b619b427df801774708bedbe0ebf79f900bfdf running on a device here connected to my phone with periodic BLE traffic between the two. I will leave it running overnight and see if anything pops up.

bergzand:
> Feel free to squash

Squashed!



> and lets hear what your experiment results for tonight were :-)

Still connected to the phone as of now and I haven't detected any issues with the BLE connection.

haukepetersen:
Perfect, then lets go as soon as Travis is happy as well.

bergzand:
@haukepetersen All green!

haukepetersen:
all green -> go.

bergzand:
Thanks!

haukepetersen:
Thank you, awesome work!

PR #755, opened 2020-02-13
npl/riot: disable SM in default configuration

haukepetersen:
This PR is a continuation of #508 and it disables the security manager in the default NimBLE configuration for RIOT. 

Besides saving some resources when not using the SM anyway, I lately run into some trouble where `ble_gap_connect()` was giving me a HCI `BLE_HS_EROLE` (0x212) error, although I am not using any SM related functionality. Disabling `BLE_SM_LEGACY` did resolve that issue for me.

@michal-narajowski I was (once more) not able to regenerate the `syscfg.h` file from the values given in `syscfg.yml`, so I manually adapted the values in question. Does this work for you?


michal-narajowski:
Yes, I think it works. I didn't test it with this particular PR but I just tested it with master and it looks like there are some changes. A reminder how to do it:

If you do not have nimble project:
- `newt new <project name>`
- `cd <project-name>`
- edit project.yml file and change the version of `apache-mynewt-core` repo to `0.0.0`
- `newt upgrade`
- now you should have all the repositories downloaded and set to `master`

Once you have nimble set up:
- make sure you are in `<project-name>` directory
- `newt build riot` (riot target is already in the nimble repo and this will compile it)
- Then you can you can copy generated files: `cp bin/@apache-mynewt-nimble/targets/riot/generated/include repos/apache-mynewt-nimble/porting/npl/riot/ -r`

haukepetersen:
@michal-narajowski thanks once more for your newbie-support! I finally got it to work, it turned out that my `newt` version was quite outdated and hence prevented me from building the `mynewt` master properly. Not that I manually build `newt` everything works as expected.

Looking at the diffs of the generated headers it seems that a couple of things have changed unrelated to this PR. So I would like to merge this PR as is (as it is tested and verified), and then prepare a follow-up to bring the header in line with the rest of the latest syscfg changes. Sound good?

haukepetersen:
perfect -> and go.

PR #763, opened 2020-02-28
[DNM] travis: Add utility to check missing porting config changes

michal-narajowski:
Use porting/update_generated_files.sh script to check if
we need to update config files in porting before merging a PR.

### DIFF
--- a/nimble/host/syscfg.yml
+++ b/nimble/host/syscfg.yml
@@ -121,7 +121,7 @@ syscfg.defs:
     # Security manager settings.
     BLE_SM_LEGACY:
         description: 'Security manager legacy pairing.'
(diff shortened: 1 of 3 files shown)


nimble/host/syscfg.yml
```diff
@@ -121,7 +121,7 @@ syscfg.defs:
     # Security manager settings.
     BLE_SM_LEGACY:
         description: 'Security manager legacy pairing.'
-        value: 1
+        value: 0
```
rymanluk, line 124: this seems to not be part of this patch

michal-narajowski: Yes, this PR is just for testing

PR #766, opened 2020-03-06
Add unittest target

utzig:
A recent newt change removed final-atom support (https://github.com/apache/mynewt-newt/pull/381); this has a side-effect that results in the "unittest" target not being found if is not located in the repo where tests are being run (it does not search for "unittest" inside @apache-mynewt-core anymore). So add the same "unittest" target that already exists in @apache-mynewt-core repo here.


utzig:
@sjanc @ccollins476ad I think this is a better solution than re-introducing the "final-atom" specifically for unittest targets in `newt`, but I would not mind doing the change there if you think it is a better solution (I tested it already!).

PR #768, opened 2020-03-11
Added advertiser, scanner, peripheral and central to apps

Reynevan94:


apps/advertiser/pkg.yml
```diff
@@ -0,0 +1,37 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: "apps/advertiser"
+pkg.type: app
+pkg.description: "Description of your app"
+pkg.author: "You <you@you.org>"
```
utzig: I like your email? :-)

Reynevan94: Fixed :)

utzig:
I don't usually review nimble PRs, but since this was opened in mynewt-core before I was a bit curious. So there are some things like functions with `name(void)` vs `name()` (with no type), `case` not aligned below the `switch`, etc. I don't have strong preferences on style but I like consistency. @sjanc @andrzej-kaczmarek @rymanluk Is nimble following the same style that core uses? In that case maybe it would be interesting to add the style checker here as well!

rymanluk:
@utzig adding style checker here would be great :)

PR #769, opened 2020-03-11
nimble/host: Couple of fixes to l2cap coc

rymanluk:


nimble/host/src/ble_l2cap_sig.c
```diff
@@ -744,7 +744,6 @@ ble_l2cap_sig_credit_base_reconfig_req_rx(uint16_t conn_handle,
          * remote will timeout.
          */
         BLE_HS_LOG(ERROR, "No memory for the response\n");
-        ble_hs_unlock();
```
sjanc: ?? this will leave lock locked

rymanluk: ups, thx

Issue #776, opened 2020-03-20
NimBLE Host: Advertisement and Scanning state should be reset when `ble_hs_reset` is called


prasad-alatkar:
If NimBLE host stack resets because of some controller issue, it should clear the advertising and scanning states.

**Brief scenario:**
 
Currently I am running into scenario, wherein `bleprph` gets DISCONNECT event and followed by host reset (`ble_hs_reset`). Now in `DISCONNECT` event, we start advertisement successfully (`bleprph_adv`). Once `ble_hs_reset` is called it resets host and again starts advertisement in host sync callback (`bleprph_on_sync`). Here we face failure to start advertisement with error code `BLE_HS_EALREADY`. It is happening because in `ble_gap_adv_validate`, this check fails --> `if (ble_gap_slave[0].op != BLE_GAP_OP_NULL)`.

PR #777, opened 2020-03-20
nimble/host: Clear master and slave states on host reset

prasad-alatkar:
Added `ble_gap_reset_master_slave_states` to clear advertising and discovery states when `ble_hs_reset` is called. This PR addresses  #776

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -1020,7 +1020,7 @@ ble_gap_slave_set_timer(uint32_t ticks_from_now)
 }
 #endif
 
-#if NIMBLE_BLE_CONNECT
+#if (NIMBLE_BLE_CONNECT || NIMBLE_BLE_SCAN)
 /**
  * Called when an error is encountered while the master-connection-fsm is
  * active.
@@ -1034,12 +1034,22 @@ ble_gap_master_failed(int status)
         ble_gap_master_connect_failure(status);
         break;
 
+#if NIMBLE_BLE_SCAN
+    case BLE_GAP_OP_M_DISC:
+        STATS_INC(ble_gap_stats, initiate_fail);
+        ble_gap_disc_complete();
+        ble_gap_master_reset_state();
+        break;
+#endif
+
     default:
         BLE_HS_DBG_ASSERT(0);
         break;
     }
 }
+#endif
 
+#if NIMBLE_BLE_CONNECT
 static void
 ble_gap_update_failed(uint16_t conn_handle, int status)
 {
@@ -1235,6 +1245,46 @@ ble_gap_adv_active_instance(uint8_t instance)
 {
     /* Assume read is atomic; mutex not necessary. */
     return ble_gap_slave[instance].op == BLE_GAP_OP_S_ADV;
+}
+
+/**
+ * Clears advertisement and discovery state.  This function is necessary
+ * when the controller loses its active state (e.g. on stack reset).
+ */
+void
+ble_gap_reset_state(int reason)
+{
+    uint16_t conn_handle;
+
+    while (1) {
+        conn_handle = ble_hs_atomic_first_conn_handle();
+        if (conn_handle == BLE_HS_CONN_HANDLE_NONE) {
+            break;
+        }
+
+        ble_gap_conn_broken(conn_handle, reason);
+    }
+
+#if NIMBLE_BLE_ADVERTISE
+#if MYNEWT_VAL(BLE_EXT_ADV)
+    uint8_t i;
+    for (i = 0; i < BLE_ADV_INSTANCES; i++) {
+        if (ble_gap_adv_active_instance(i)) {
+            /* Indicate to application that advertising has stopped. */
+            ble_gap_adv_finished(i, reason, 0, 0);
+        }
+    }
+#else
+    if (ble_gap_adv_active_instance(0)) {
+        /* Indicate to application that advertising has stopped. */
+        ble_gap_adv_finished(0, reason, 0, 0);
+    }
+#endif
+#endif
+
+#if (NIMBLE_BLE_SCAN || NIMBLE_BLE_CONNECT)
+    ble_gap_master_failed(reason);
+#endif
 }
 
 #if NIMBLE_BLE_CONNECT
--- a/nimble/host/src/ble_gap_priv.h
+++ b/nimble/host/src/ble_gap_priv.h
@@ -136,6 +136,7 @@ void ble_gap_preempt(void);
 void ble_gap_preempt_done(void);
 
 int ble_gap_terminate_with_conn(struct ble_hs_conn *conn, uint8_t hci_reason);
+void ble_gap_reset_state(int reason);
 void ble_gap_conn_broken(uint16_t conn_handle, int reason);
 int32_t ble_gap_timer(void);
 
--- a/nimble/host/src/ble_hs.c
+++ b/nimble/host/src/ble_hs.c
@@ -361,7 +361,6 @@ ble_hs_sync(void)
 static int
 ble_hs_reset(void)
 {
-    uint16_t conn_handle;
     int rc;
 
     STATS_INC(ble_hs_stats, reset);
@@ -376,14 +375,8 @@ ble_hs_reset(void)
 
     ble_hs_clear_rx_queue();
 
-    while (1) {
-        conn_handle = ble_hs_atomic_first_conn_handle();
-        if (conn_handle == BLE_HS_CONN_HANDLE_NONE) {
-            break;
-        }
-
-        ble_gap_conn_broken(conn_handle, ble_hs_reset_reason);
-    }
+    /* Clear adverising and scanning states. */
+    ble_gap_reset_state(ble_hs_reset_reason);
 
     /* Clear configured addresses. */
     ble_hs_id_reset();


prasad-alatkar:
@sjanc @rymanluk @christ PTAL

nimble/host/src/ble_gap.c
```diff
@@ -737,6 +737,30 @@ ble_gap_slave_reset_state(uint8_t instance)
...
+    for (i = 0; i < BLE_ADV_INSTANCES; i++) {
+        ble_gap_slave_reset_state(i);
+    }
+#else
+    ble_gap_slave_reset_state(0);
+#endif
+#endif
+
+#if NIMBLE_BLE_SCAN
+    ble_gap_master_reset_state();
```
rymanluk: For the master role, imho we need bit more:
1) when BLE_GAP_OP_M_DISC is ongoing, we need to call  ble_gap_disc_complete()
2) when BLE_GAP_OP_M_CONN is ongoing we probably should call ble_gap_master_connect_failure()

nimble/host/src/ble_gap_priv.h
```diff
@@ -136,6 +136,7 @@ void ble_gap_preempt(void);
 void ble_gap_preempt_done(void);
 
 int ble_gap_terminate_with_conn(struct ble_hs_conn *conn, uint8_t hci_reason);
+void ble_gap_master_slave_reset_state(void);
```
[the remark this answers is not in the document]
prasad-alatkar: Sure, but `ble_gap_reset_state` appears quite generic, doesn't it ?

nimble/host/src/ble_gap.c
```diff
@@ -737,6 +737,30 @@ ble_gap_slave_reset_state(uint8_t instance)
...
+ */
+void
+ble_gap_master_slave_reset_state(void)
+{
+#if NIMBLE_BLE_ADVERTISE
+#if MYNEWT_VAL(BLE_EXT_ADV)
+    uint8_t i;
+
+    for (i = 0; i < BLE_ADV_INSTANCES; i++) {
+        ble_gap_slave_reset_state(i);
```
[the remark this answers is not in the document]
prasad-alatkar: Yes, we should, have made the changes.

nimble/host/src/ble_gap.c
prasad-alatkar: 1. Yes, we should call ` ble_gap_disc_complete`, have made changes.
2. Yes, however in `ble_hs_reset`, we also call `ble_gap_conn_broken` which ultimately calls GAP event `BLE_GAP_EVENT_DISCONNECT` to notify app. But I do understand, you want to cover the cases where  there may be `connection establishment procedure` and may not be any `active connection (established)`.

nimble/host/src/ble_gap.c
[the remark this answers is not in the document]
rymanluk: but, don't  we need to call it only for the active instances?

nimble/host/src/ble_gap.c
```diff
@@ -737,6 +748,39 @@ ble_gap_slave_reset_state(uint8_t instance)
...
+        /* Indicate to application that advertising has stopped. */
+        ble_gap_adv_finished(i, reason, conn_handle, 0);
+    }
+#else
+    ble_gap_slave_reset_state(0);
+    /* Indicate to application that advertising has stopped. */
+    ble_gap_adv_finished(0, reason, conn_handle, 0);
```
rymanluk: same here, we need to do in only when active.

nimble/host/src/ble_gap.c
```diff
@@ -737,6 +748,39 @@ ble_gap_slave_reset_state(uint8_t instance)
...
+ble_gap_reset_state(uint16_t conn_handle, int reason)
+{
+#if NIMBLE_BLE_ADVERTISE
+#if MYNEWT_VAL(BLE_EXT_ADV)
+    uint8_t i;
+
+    for (i = 0; i < BLE_ADV_INSTANCES; i++) {
+        ble_gap_slave_reset_state(i);
+        /* Indicate to application that advertising has stopped. */
+        ble_gap_adv_finished(i, reason, conn_handle, 0);
```
rymanluk: I don't think we need `conn_handle` here. It is used when advertising is completed because connection has been established and it is not a case here for sure

nimble/host/src/ble_gap.c
```diff
@@ -737,6 +748,39 @@ ble_gap_slave_reset_state(uint8_t instance)
 #endif
 }
 
+/**
+ * Clears advertisement and discovery state.  This function is necessary
+ * when the controller loses its active state (e.g. on stack reset).
+ */
+void
+ble_gap_reset_state(uint16_t conn_handle, int reason)
```
rymanluk: I would remove `conn_handle` at all because it is not needed. I explained below why.

nimble/host/src/ble_hs.c
```diff
@@ -388,6 +388,13 @@ ble_hs_reset(void)
     /* Clear configured addresses. */
     ble_hs_id_reset();
 
+    /* Clear adverising and scanning states. */
+    if (conn_handle == BLE_HS_CONN_HANDLE_NONE) {
```
rymanluk: we should just call `ble_gap_reset_state` with a `ble_hs_reset_reason`.  Imho `conn_handle` at this point is always `BLE_HS_CONN_HANDLE_NONE`.

BTW later I think that we can move following block
```
 while (1) {

        conn_handle = ble_hs_atomic_first_conn_handle();
        if (conn_handle == BLE_HS_CONN_HANDLE_NONE) {
            break;
        }

        ble_gap_conn_broken(conn_handle, ble_hs_reset_reason);
    }
```
to the new `ble_gap_reset_state()`

nimble/host/src/ble_gap.c
rymanluk: ad 2. Because application started connection procedure and is waiting for a result of it. 
As the example, if device is not around and application sent Connect, host has internal timeout for Connection procedure. When it timeouts, then host do connect cancel and when connection complete arrives host is calling `ble_gap_master_connect_cancelled`. In our case we don't want to send connect cancel but we want to notify application that connection is failed.

nimble/host/src/ble_gap.c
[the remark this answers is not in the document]
prasad-alatkar: Yes, but I thought as we are anyway resetting the host stack, so resetting slave_state should be all right. However I understand, as far as `ble_gap_adv_finished` is concerned, we should only call it when actually active adv instance was there.

nimble/host/src/ble_gap.c
prasad-alatkar: Thank you for the explanation :+1:

nimble/host/src/ble_hs.c
prasad-alatkar: Makes sense :+1:

nimble/host/src/ble_gap.c
prasad-alatkar: Done !!

nimble/host/src/ble_gap.c
prasad-alatkar: True.

nimble/host/src/ble_gap.c
```diff
@@ -1237,6 +1237,53 @@ ble_gap_adv_active_instance(uint8_t instance)
...
+    if (ble_gap_adv_active_instance(0)) {
+        ble_gap_slave_reset_state(0);
+        /* Indicate to application that advertising has stopped. */
+        ble_gap_adv_finished(0, reason, 0, 0);
+    }
+#endif
+#endif
+
+#if NIMBLE_BLE_SCAN
+    ble_gap_master_reset_state();
```
rymanluk: this will reset state before we check it below

nimble/host/src/ble_gap.c
```diff
@@ -1237,6 +1237,53 @@ ble_gap_adv_active_instance(uint8_t instance)
...
+    if (ble_gap_disc_active()) {
+        ble_gap_disc_complete();
+    } else if (ble_gap_conn_active()) {
+        ble_gap_master_failed(reason);
```
rymanluk: hmm when you look inside ` ble_gap_master_failed`  function, there is a switch.  Maybe it is worth to add there case for `BLE_GAP_OP_M_DISC` and then call that function? I would also add there `ble_gap_master_reset_state()`

nimble/host/src/ble_gap.c
```diff
@@ -1237,6 +1237,53 @@ ble_gap_adv_active_instance(uint8_t instance)
...
+    for (i = 0; i < BLE_ADV_INSTANCES; i++) {
+        if (ble_gap_adv_active_instance(i)) {
+            ble_gap_slave_reset_state(i);
+            /* Indicate to application that advertising has stopped. */
+            ble_gap_adv_finished(i, reason, 0, 0);
+        }
+    }
+#else
+    if (ble_gap_adv_active_instance(0)) {
+        ble_gap_slave_reset_state(0);
```
rymanluk: I don't think we need to call that as this is called later in `ble_gap_adv_finished() ->ble_gap_slave_extract_cb()->ble_gap_slave_reset_state()`

nimble/host/src/ble_gap.c
```diff
@@ -1237,6 +1237,53 @@ ble_gap_adv_active_instance(uint8_t instance)
...
+
+        ble_gap_conn_broken(conn_handle, reason);
+    }
+
+#if NIMBLE_BLE_ADVERTISE
+#if MYNEWT_VAL(BLE_EXT_ADV)
+    uint8_t i;
+    for (i = 0; i < BLE_ADV_INSTANCES; i++) {
+        if (ble_gap_adv_active_instance(i)) {
+            ble_gap_slave_reset_state(i);
```
rymanluk: I don't think we need to call that as this is called later in ble_gap_adv_finished() ->ble_gap_slave_extract_cb()->ble_gap_slave_reset_state()

nimble/host/src/ble_hs.c
```diff
@@ -376,18 +375,12 @@ ble_hs_reset(void)
...
-        }
-
-        ble_gap_conn_broken(conn_handle, ble_hs_reset_reason);
-    }
-
     /* Clear configured addresses. */
     ble_hs_id_reset();
 
+    /* Clear adverising and scanning states. */
+    ble_gap_reset_state(ble_hs_reset_reason);
```
rymanluk: I would call it before the `ble_hs_id_reset()`. It probably will change nothing, but since before we were calling `ble_gap_conn_broken()` before that I don't see a reason to change that.

nimble/host/src/ble_gap.c
prasad-alatkar: Yes !!

nimble/host/src/ble_gap.c
prasad-alatkar: Yes !!

nimble/host/src/ble_hs.c
prasad-alatkar: > It probably will change nothing

Had the same though in mind :smile: will make the changes.

nimble/host/src/ble_gap.c
prasad-alatkar: Ok !!

nimble/host/src/ble_gap.c
prasad-alatkar: sorry for that, done !!

nimble/host/src/ble_gap.c
```diff
@@ -1237,6 +1243,46 @@ ble_gap_adv_active_instance(uint8_t instance)
...
+    }
+#else
+    if (ble_gap_adv_active_instance(0)) {
+        /* Indicate to application that advertising has stopped. */
+        ble_gap_adv_finished(0, reason, 0, 0);
+    }
+#endif
+#endif
+
+#if NIMBLE_BLE_SCAN
```
rymanluk: #if( NIMBLE_BLE_SCAN || NIMBLE_BLE_CONNECT) ?

nimble/host/src/ble_gap.c
```diff
@@ -1034,6 +1034,12 @@ ble_gap_master_failed(int status)
         ble_gap_master_connect_failure(status);
         break;
 
+    case BLE_GAP_OP_M_DISC:
+        STATS_INC(ble_gap_stats, initiate_fail);
+        ble_gap_disc_complete();
+        ble_gap_master_reset_state();
```
rymanluk: could you double check if ` ble_gap_master_reset_state();` we need only for `BLE_GAP_OP_M_DISC:` and not for `BLE_GAP_OP_M_CONN' ?

rymanluk: BTW - just noticed that above `ble_gap_master_failed(int status)` there is #if NIMBLE_BLE_CONNECT

I believe we need this #if inside that function. Also #if NIMBLE_BLE_SCAN should be there.

prasad-alatkar: > could you double check if ble_gap_master_reset_state(); we need only for BLE_GAP_OP_M_DISC: and not for `BLE_GAP_OP_M_CONN' ?

For `BLE_GAP_OP_M_CONN`, `ble_gap_master_connect_failure` --> `ble_gap_master_extract_state` --> calls `ble_gap_master_reset_state`. So we need it only for `BLE_GAP_OP_M_DISC`

prasad-alatkar: `#if NIMBLE_BLE_CONNECT` is applied for both functions: `ble_gap_master_failed` & `ble_gap_update_failed`, I will make changes accordinly. 

> I believe we need this #if inside that function. Also #if NIMBLE_BLE_SCAN should be there.

Yes, I will make the required changes.

prasad-alatkar: Hi @rymanluk I am getting CI failure

```
echo 'Building target=nordic_pca10056_bttester'

Building target=nordic_pca10056_bttester

+newt build -q -l info nordic_pca10056_bttester

2020/03/26 11:46:51.604 [INFO] Ignoring /home/travis/gopath/src/github.com/apache/mynewt-nimble/repos/apache-mynewt-core/hw/mcu/nordic/src/ext/nrfx/drivers/src/nrfx_power.c because package dictates it.

+rc=0

+[[ 0 -ne 0 ]]

+[[ 0 -eq 127 ]]

+for target in '${TARGETS}'

+echo 'Building target=nordic_pca10056_ext_advertiser'

Building target=nordic_pca10056_ext_advertiser

+newt build -q -l info nordic_pca10056_ext_advertiser

2020/03/26 11:47:43.511 [INFO] Ignoring /home/travis/gopath/src/github.com/apache/mynewt-nimble/repos/apache-mynewt-core/hw/mcu/nordic/src/ext/nrfx/drivers/src/nrfx_power.c because package dictates it.

Error: nimble/host/src/ble_gap.c:1028:1: error: 'ble_gap_master_failed' defined but not used [-Werror=unused-function]

 ble_gap_master_failed(int status)

 ^~~~~~~~~~~~~~~~~~~~~

cc1: all warnings being treated as errors

```

prasad-alatkar: I think it is better if we keep `#if NIMBLE_BLE_CONNECT` or `#if NIMBLE_BLE_SCAN` outside of `ble_gap_master_failed` function.

nimble/host/src/ble_gap.c
prasad-alatkar: Hi @rymanluk Do we really need `NIMBLE_BLE_CONNECT` here ? I mean given that we want to reset state of only master, I was thinking `NIMBLE_BLE_SCAN` should suffice.

Issue #780, opened 2020-03-23
Should 40s timeout apply to connection parameter updates?


zacwbond:
I have a connection using a connection interval of 120, latency of 90, and timeout of 3000.  The spec says in section 6.B.5.4, 

>The master should allow a minimum of 6 connection events that the slave will be listening for before the instant occurs, considering that the slave may only be listening once every connSlaveLatency events.

Since a connection update isn't done until the new parameters are applied, that means in my case a parameter update will take 120 *1.25ms * 90 * 6 = 81000ms, or 1:21.  

But in `ble_gap.c` there's a hard-coded timeout of `BLE_GAP_UPDATE_TIMEOUT_MS` = 40s applied to connection parameter update events.  I assume 40s was chosen because of 6.B.5.2, but that section also says there's no timeout rules for connection update procedures.  

The result is that in my application, when the peripheral asks for new connection parameters, the master times out and disconnects the peripheral before the process is expected to have completed.

I think the timeout for connection parameter update events should be based on the current connection parameters rather than a hardcoded value.

rymanluk:
Fixed with https://github.com/apache/mynewt-nimble/pull/782

PR #781, opened 2020-03-23
Make connection update process timeout variable

zacwbond:
This is a fix for #780, an issue I submitted earlier today.  

I am connecting as a central to a remote peripheral using a connection interval of 120, latency of 90, and timeout of 3000.  Due to the hard-coded 40 second timeout, it is impossible to perform a connection parameter update on this connection.  Trying to do so results in a timeout followed by a disconnect 100% of the time.

My attempt to fix the problem is to compute the timeout based on the current connection parameters instead of just hard-coding a constant value.

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -88,7 +88,7 @@
 #define BLE_GAP_CANCEL_RETRY_TIMEOUT_MS         100 /* ms */
 
 #define BLE_GAP_UPDATE_TIMEOUT_MS(itvl, latency) \
-            (6 * ((itvl) * BLE_HCI_CONN_ITVL / 1000) * ((latency) + 1))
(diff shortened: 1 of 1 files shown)


zacwbond:
@rymanluk Your a) makes sense to me.  If I get a timeout, I'd tend to assume that the event that timed out is never going to happen, so it would be odd to send a timeout event if the procedure is eventually going to succeed.

Another problem I've noticed is that I frequently see a call to my reset_cb with error code 19 when the gap connection parameter update procedure is initiated.  I think it must be associated with the function `ble_hs_hci_wait_for_ack()`.  That makes it sound like an HCI problem and not a Nimble problem, but it only ever happens with the parameter update procedure, and only intermittently.

rymanluk:
@zacwbond could you update your PR so it implements a) ?

the other problem you mentioned is something with your controller I believe.

zacwbond:
Closing because I mangled my forks somehow.  Will re-open momentarily.

PR #783, opened 2020-03-25
Make ble_gap_rx_l2cap_update_req() behave like ble_gap_rx_param_req()

zacwbond:
`BLE_GAP_EVENT_L2CAP_UPDATE_REQ` and `BLE_GAP_EVENT_CONN_UPDATE_REQ` share the same `conn_update_req` event structure when they are sent to the application callback, but they do not behave the same way.  

`ble_gap_rx_param_req()` creates a self_params, allowing the application to adjust the parameters to something acceptable.  `ble_gap_rx_l2cap_update_req()` just leaves the `self_params` structure NULL and otherwise ignores it.  Now they both behave in the same way.

In my specific case, the ESP32 platform I am using crashes if I don't enforce a specific set of parameters when multiple peripherals connect.  Since the peripherals use the L2CAP method to make the request, this change eliminates the crash.


h2zero:
I’ve been seeing this happen as well on esp32. Only way to prevent it I have found is to reject the peer request then manually change the parameters after. This PR makes more sense.

zacwbond:
> The reson why it differs between BLE_GAP_EVENT_L2CAP_UPDATE_REQ and BLE_GAP_EVENT_CONN_UPDATE_REQ is basically because over L2CAP you can only accept or reject the requested parameters. There is no option to modify the parameters here.
> Having this patch, we could end up in the situation that remote devices requested for parameters set A, we reply over L2CAP that we accepted those, but in fact we set parameters B, which is not nice.
> If master wants to change parameters, he can always do it, but if remote ask for the specific parameters set, we either Reject or Accept and set it.
> 
> I believe, that you need to fix you application and on BLE_GAP_EVENT_L2CAP_UPDATE_REQ does not modify self_parameters.
> 
> Hope that explains.
> If you agree please close this PR.

If the peripheral requests an interval between 16 and 20, and I force the radio through this mechanism to use an interval of 18, it seems to me I've still accepted the peripheral's request because the connection interval will still match the peripheral's request.

I think the only invalid behavior would be to accept the request but to choose a  parameter out of the range requested by the peripheral (or to change one of the parameters that isn't given as a range).

h2zero:
@zacwbond What @rymanluk is saying is that the procedure you are suggesting would go against the bluetooth core spec, although it does help with the esp32 issues. 

The esp32 has some weird issues with connection parameters stuffing it up with multiple connections, this is a specific case though. I think it's better to handle it manually by just rejecting the request and then calling `ble_gap_update_params()` afterward. This may cause issues if your peripherals are picky about accepting their parameters though, in which case there's not much else to do but maintain this patch in your own code.

rymanluk:
@zacwbond Actually on the second thought, why not to allow application to fine tune that parameters. It is up to application to behave correctly here.

If there will be no voices against that I will merge it. Thanks

andrzej-kaczmarek:
if we are going to allow adjusting min/max interval by application, we also need to make sure that selected range is within original range, otherwise we violate spec as @rymanluk said before. adding this extra check would both resolve original issue and would not make our host accept parameters that violate spec.

rymanluk:
@andrzej-kaczmarek  I agree.

zacwbond:
> @zacwbond Could you please add additional check as @andrzej-kaczmarek suggested?

Got it, thanks.  I'll make the change as soon as I am able.

nimble/host/src/ble_gap.c
```diff
@@ -1806,15 +1806,40 @@ int
...
     rc = ble_gap_call_conn_event_cb(&event, conn_handle);
+
+    if ((self_params.itvl_min <= self_params.itvl_max)
+         && (self_params.itvl_min >= params->itvl_min)
+         && (self_params.itvl_max <= params->itvl_max)) {
+             params->itvl_max = self_params.itvl_max;
+             params->itvl_min = self_params.itvl_min;
+    }
+
+    if ((self_params.min_ce_len <= self_params.max_ce_len)
```
h2zero, line 1834: Unless I'm mistaken we probably shouldn't restrict the min/max_ce_len values as they are not requested by the peer and actually just initialized [here](https://github.com/apache/mynewt-nimble/blob/6ced1dd4ca63e6f0bd2df7dcc558afcc2896f1c2/nimble/host/src/ble_l2cap_sig.c#L414) to:
```
    params.min_ce_len = BLE_GAP_INITIAL_CONN_MIN_CE_LEN;
    params.max_ce_len = BLE_GAP_INITIAL_CONN_MAX_CE_LEN;
```
before this gets called.

zacwbond: I checked in a fix, but realized I might have done the opposite thing to what you wanted.  Do you want to allow the callback to change the ce_len to whatever it wants, or do you want to prevent it from changing them at all?  I did the latter, but I think you may have meant the former.

h2zero: Yes, I think the app should be able to change the `min/max_ce_len` to whatever it wants. 
That's just my opinion, we'll have to see what @rymanluk thinks.

Also, since you're using the esp32 as well you might be interested in [this](https://github.com/espressif/esp-nimble/issues/4) for some insight into my opinion.

nimble/host/include/host/ble_gap.h
```diff
@@ -610,6 +610,11 @@ struct ble_gap_event {
              * like to use.  The application callback should fill this in.  By
              * default, this struct contains the requested parameters (i.e.,
              * it is a copy of 'peer_params').
+             *
+             * For BLE_GAP_EVENT_L2CAP_UPDATE_REQ, only connection interval
+             * min/max and connection length min/max can be changed.  If they
```
rymanluk, line 615: Connection length?

nimble/host/include/host/ble_gap.h
```diff
@@ -610,6 +610,11 @@ struct ble_gap_event {
...
+             * are changed, they must remain within the original range.
```
rymanluk, line 616: I would wrote, that if new interval is not in the range of the one provide by the peripheral, it will be ignored.

nimble/host/include/host/ble_gap.h
```diff
@@ -610,6 +610,16 @@ struct ble_gap_event {
              * like to use.  The application callback should fill this in.  By
              * default, this struct contains the requested parameters (i.e.,
              * it is a copy of 'peer_params').
+             *
+             * For BLE_GAP_EVENT_L2CAP_UPDATE_REQ, only connection interval
+             * min/max and connection event length min/max can be changed.
```
rymanluk, line 615: @zacwbond we are saying here that min/max CE might be changed but code does not copy those parameters from self_params.

Issue #788, opened 2020-03-30
nimble/store: `ble_gap_unpair_oldest_peer` in `ble_store_util_status_rr`(store callback) can cause infinite loop for CCCDs


prasad-alatkar:
In cases where CCCDs are exceeding `MYNEWT_VAL(BLE_STORE_MAX_CCCDS)` (overflow event), we can end up in continuous while loop. 

Whenever `BLE_STORE_EVENT_OVERFLOW` is called in store callback, if we do not have any previously bonded peer, `ble_gap_unpair_oldest_peer` will return `0` without deleting any CCCD. This in turn results in continuous while loop in `ble_store_write` as it will continuously end up calling `BLE_HS_ESTORE_CAP` evwnt.

h2zero:
I found this as well, currently testing a fix in my own fork. works for my purposes.
You can see it [here](https://github.com/h2zero/NimBLE-Arduino/blob/bae2eb26d0128d649eb854bc23a4846d7d7581fc/src/nimble/host/src/ble_gap.c#L5305)

prasad-alatkar:
@h2zero your proposed solution may avoid this issue when "Bonding" is not enabled, but the issue will eventually pop-up in case of bonding is enabled (i.e. `ble_hs_cfg.sm_bonding = 1`).

h2zero:
Yes, you are correct. It was just a quick patch to get by while developing the cpp library to avoid that error. Didn’t want to do any big changes to the NimBLE core code yet.

I do not think we should store cccd data if not bonding though, maybe a combination of my fix and a proper round robin cccd store patch?

rymanluk:
This is fixed. Closing.

PR #790, opened 2020-03-31
nimble/store: Fix nimble store behavior when CCCDs exceed static defined limit

prasad-alatkar:
**Issue reference**: #788 

**Description:**
 In cases where CCCDs are exceeding MYNEWT_VAL(BLE_STORE_MAX_CCCDS) (overflow event), we can end up in continuous while loop.

**Change list:**
1. In `ble_gap_unpair_oldest_peer` changed return value from `0` to `BLE_HS_ENOENT` when there is no entry found.
2. Modified `ble_store_util_status_rr` to address overflow event of CCCDs.

Previous to this PR, if CCCDs used to exceed the MAX_CCCDs then in `ble_store_write` call there used to come OVERFLOW event, which used to call `ble_store_delete_oldest_peer`, now here as we do not have any peer to be deleted, we should have identified that we need to make space for CCCDs and not `peer_sec` or `our_sec`. The PR tries to address this issue.

### DIFF
--- a/nimble/host/include/host/ble_gap.h
+++ b/nimble/host/include/host/ble_gap.h
@@ -1896,6 +1896,20 @@ int ble_gap_unpair(const ble_addr_t *peer_addr);
  */
 int ble_gap_unpair_oldest_peer(void);
 
+/**
+ * Similar to `ble_gap_unpair_oldest_peer()`, except it makes sure that the
+ * peer received in input parameters is not deleted.
+ *
+ * @param peer_addr             Address of the peer (not to be deleted)
+ *
+ * @return                      0 on success;
+ *                              A BLE host HCI return code if the controller
+ *                                  rejected the request;
+ *                              A BLE host core return code on unexpected
+ *                                  error.
+ */
+int ble_gap_unpair_oldest_except(const ble_addr_t *peer_addr);
+
 #define BLE_GAP_PRIVATE_MODE_NETWORK        0
 #define BLE_GAP_PRIVATE_MODE_DEVICE         1
 
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -5605,6 +5605,36 @@ ble_gap_unpair_oldest_peer(void)
     return 0;
 }
 
+int
+ble_gap_unpair_oldest_except(const ble_addr_t *peer_addr)
+{
+    ble_addr_t peer_id_addrs[MYNEWT_VAL(BLE_STORE_MAX_BONDS)];
+    int num_peers;
+    int rc, i;
+
+    rc = ble_store_util_bonded_peers(
+            &peer_id_addrs[0], &num_peers, MYNEWT_VAL(BLE_STORE_MAX_BONDS));
+    if (rc != 0) {
+        return rc;
+    }
+
+    if (num_peers == 0) {
+        return BLE_HS_ENOENT;
+    }
+
+    for (i = 0; i < num_peers; i++) {
+        if (ble_addr_cmp(peer_addr, &peer_id_addrs[i]) != 0) {
+            break;
+        }
+    }
+
+    if (i >= num_peers) {
+        return BLE_HS_ENOMEM;
+    }
+
+    return ble_gap_unpair(&peer_id_addrs[i]);
+}
+
 void
 ble_gap_passkey_event(uint16_t conn_handle,
                       struct ble_gap_passkey_params *passkey_params)
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -5594,7 +5594,7 @@ ble_gap_unpair_oldest_peer(void)
     }
 
     if (num_peers == 0) {
-        return 0;
+        return BLE_HS_ENOENT;
     }
 
     rc = ble_gap_unpair(&oldest_peer_id_addr);
--- a/nimble/host/src/ble_store_util.c
+++ b/nimble/host/src/ble_store_util.c
@@ -233,13 +233,15 @@ ble_store_util_status_rr(struct ble_store_status_event *event, void *arg)
     switch (event->event_code) {
     case BLE_STORE_EVENT_OVERFLOW:
         switch (event->overflow.obj_type) {
-            case BLE_STORE_OBJ_TYPE_OUR_SEC:
-            case BLE_STORE_OBJ_TYPE_PEER_SEC:
-            case BLE_STORE_OBJ_TYPE_CCCD:
-                return ble_gap_unpair_oldest_peer();
+        case BLE_STORE_OBJ_TYPE_OUR_SEC:
+        case BLE_STORE_OBJ_TYPE_PEER_SEC:
+            return ble_gap_unpair_oldest_peer();
+        case BLE_STORE_OBJ_TYPE_CCCD:
+            /* Try unpairing oldest peer except current peer */
+            return ble_gap_unpair_oldest_except(&event->overflow.value->cccd.peer_addr);
 
-            default:
-                return BLE_HS_EUNKNOWN;
+        default:
+            return BLE_HS_EUNKNOWN;
         }
 
     case BLE_STORE_EVENT_FULL:


prasad-alatkar:
@h2zero That is a valid point, will update if I find anything. Maybe @rymanluk can explain it to us.

nimble/host/src/ble_gap.c
```diff
@@ -5594,7 +5594,7 @@ ble_gap_unpair_oldest_peer(void)
     }
 
     if (num_peers == 0) {
-        return 0;
+        return BLE_HS_ENOENT;
```
rymanluk, line 5597: Can we make it as separate patch?

nimble/host/src/ble_store_util.c
```diff
@@ -230,13 +299,35 @@ ble_store_util_delete_oldest_peer(void)
...
+    ble_addr_t peer_id_addr;
+    int num_peers;
+
     switch (event->event_code) {
     case BLE_STORE_EVENT_OVERFLOW:
         switch (event->overflow.obj_type) {
             case BLE_STORE_OBJ_TYPE_OUR_SEC:
             case BLE_STORE_OBJ_TYPE_PEER_SEC:
             case BLE_STORE_OBJ_TYPE_CCCD:
-                return ble_gap_unpair_oldest_peer();
```
rymanluk, line 239: I believe that for CCCD we should have different handling and if we want to unpair older peer we should not unpair peer for which we store CCCD.  If there is nothing to unpair just return the error.

In the CCCD event there is and ble_addr we should use to make sure that one is no unpaired.

nimble/host/src/ble_gap.c
prasad-alatkar: Done !!

nimble/host/src/ble_store_util.c
prasad-alatkar: @rymanluk agreed !! I have made relevant changes. For CCCD,  if `ble_gap_unpair_oldest_peer` is unable to delete any peer, we continue to find other peers occupying CCCDs in storage. I have also modified callback function (`ble_store_util_iter_peer_cccd`) to skip the current (i.e. for which CCCDs need to be stored) peer entry.

nimble/host/src/ble_store_util.c
```diff
@@ -230,16 +315,43 @@ ble_store_util_delete_oldest_peer(void)
...
-            case BLE_STORE_OBJ_TYPE_CCCD:
-                return ble_gap_unpair_oldest_peer();
-
-            default:
-                return BLE_HS_EUNKNOWN;
+        case BLE_STORE_OBJ_TYPE_OUR_SEC:
+        case BLE_STORE_OBJ_TYPE_PEER_SEC:
+            return ble_gap_unpair_oldest_peer();
+        case BLE_STORE_OBJ_TYPE_CCCD:
+            if ((rc = ble_gap_unpair_oldest_peer()) == BLE_HS_ENOENT) {
```
rymanluk, line 329: I was thinking to call here other version of `unpair_oldest` which would take as a parameter `ble_addr_t ` of device which we don't want to unpair.
If the only device which could be unpair is the one provided, then unpair is not done and this function returns error.

In such a case we know that we cannot store CCC and there is nothing we can do more here. All the other code is not needed then.

prasad-alatkar: @rymanluk That is what I have done actually in `ble_store_util_subscribed_cccds` !! What I think is `ble_gap_unpair_oldest_peer` is not anyway going to unpair current peer (the one for which we want to make space), so introducing similar function with `ble_addr_t` as input won't be really that helpful. Please let me know your thoughts on this.

The issue I believe is only confined to CCCDs, so  `ble_gap_unpair_oldest_peer()) == BLE_HS_ENOENT` will make sure that we do not have any bonded devices and still we are out of storage space, which can be used to conclude that we need to clean old CCCDs (except the current one).

rymanluk: @prasad-alatkar are you sure that `ble_gap_unpair_oldest_peer()` will not unpair current peer? For me it looks it will.

h2zero: I believe @rymanluk is correct, if the current peer exceeds the cccd limit it will likely unpair that peer and disconnect. I like the do nothing if full approach here.

prasad-alatkar: @rymanluk @h2zero , disregard my earlier comment, yes current peer will be unpaired, and we should simply return error in that case.

I was thinking of adding utility function (`ble_store_clean_old_cccd`) which will delete CCCDs of all unbonded peers. This can be called first in case of `CCCD overflow` and if it fails to delete peer then we can continue with  modified `ble_gap_unpair_oldest_peer()`. Please let me know your views on it.

rymanluk: I believe that `ble_store_util_delete_peer()` deletes all what we need.

prasad-alatkar: Hi @rymanluk I think for some peers `CCCD` will exist but `peer_sec` and `our_sec` won't exist (Ref: [subscribe-event ble_gatts_clt_cfg_access](https://github.com/apache/mynewt-nimble/blob/9e26fbe97b3d0bb3a22f15ed9c6035be648d1104/nimble/host/src/ble_gatts.c#L782)). So we need mechanism to clean these stored CCCDs, that is why I have introduced `ble_store_clean_old_cccd()` function.

nimble/host/src/ble_gap.c
```diff
@@ -5605,6 +5605,41 @@ ble_gap_unpair_oldest_peer(void)
     return 0;
 }
 
+int
+ble_gap_unpair_oldest_except_curr(const ble_addr_t *curr_peer)
+{
+    ble_addr_t oldest_peer_id_addr[MYNEWT_VAL(BLE_STORE_MAX_BONDS)];
```
rymanluk, line 5611: nitpick: please rename to `peer_id_addrs`

nimble/host/include/host/ble_gap.h
```diff
@@ -1896,6 +1896,20 @@ int ble_gap_unpair(const ble_addr_t *peer_addr);
...
+ *
+ * @param peer_addr             Address of the current peer (not to be deleted)
+ *
+ * @return                      0 on success;
+ *                              A BLE host HCI return code if the controller
+ *                                  rejected the request;
+ *                              A BLE host core return code on unexpected
+ *                                  error.
+ */
+int ble_gap_unpair_oldest_except_curr(const ble_addr_t *curr_peer);
```
rymanluk, line 1911: let us call it `ble_gap_unpair_oldest_except` () and change the parameter name to `peer_addr`. 
This functionality is not really dedicated only to current device. You might want to have device which you don't want to unpair no matter what :)

nimble/host/src/ble_store_util.c
```diff
@@ -230,16 +372,24 @@ ble_store_util_delete_oldest_peer(void)
...
-                return ble_gap_unpair_oldest_peer();
-
-            default:
-                return BLE_HS_EUNKNOWN;
+        case BLE_STORE_OBJ_TYPE_OUR_SEC:
+        case BLE_STORE_OBJ_TYPE_PEER_SEC:
+            return ble_gap_unpair_oldest_peer();
+        case BLE_STORE_OBJ_TYPE_CCCD:
+            /* Try to remove unbonded CCCDs first */
+            if ((rc = ble_store_clean_old_cccds((void *) &event->overflow.value->cccd.peer_addr)) == BLE_HS_ENOENT) {
```
rymanluk, line 384: I don't think we need to clean unbodend CCCDs as we do not keep such in the storage. Unless I miss something, but CCCDs are stored only for bonded devices, and once unbonded, we do clean CCCDs. If it does not work like this (I believe it does) then we have problem somewhere else.

In this case we just need to call the new function and that is it. Nothing else is needed.

nimble/host/src/ble_gap.c
```diff
@@ -5605,6 +5605,41 @@ ble_gap_unpair_oldest_peer(void)
...
+    if (rc != 0) {
+        return rc;
+    }
+
+    if (num_peers == 0) {
+        return BLE_HS_ENOENT;
+    }
+
+    for (i = 0; i < num_peers; i++) {
+        if (memcmp(curr_peer, &oldest_peer_id_addr[i], sizeof (ble_addr_t)) != 0) {
```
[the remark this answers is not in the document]
prasad-alatkar: Sure !!

nimble/host/include/host/ble_gap.h
prasad-alatkar: Agreed !!

nimble/host/src/ble_store_util.c
prasad-alatkar: Hi @rymanluk Please let me know if I am missing something. As per my understanding CCCDs for unbonded devices can get stored as `ble_gatts_clt_cfg_access --> ble_store_write_cccd --> ble_store_write(BLE_STORE_OBJ_TYPE_CCCD, store_value)`.

h2zero: > I don't think we need to clean unbodend CCCDs as we do not keep such in the storage. Unless I miss something, but CCCDs are stored only for bonded devices, and once unbonded, we do clean CCCDs. If it does not work like this (I believe it does) then we have problem somewhere else.
> 
> In this case we just need to call the new function and that is it. Nothing else is needed.

@rymanluk CCCD's do get stored if the devices pair but one or both of them does not accept bonding, so cleaning those up makes sense in this case. Although it would be better if that never happened to begin with.

rymanluk: @h2zero actually CCCD's do get stored when devices are bonded.  That happens after pairing is completed and both devices have bonding flag in set. 

If you do see an issue that CCCDs are stored but device is not actually bonded, please report it as an issue.

PR #796, opened 2020-04-09
Apps: central added

KKopyscinski:
Added central application which presents how to connect to peripheral.

### DIFF
--- a/apps/central/src/main.c
+++ b/apps/central/src/main.c
@@ -1,0 +1,186 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "sysinit/sysinit.h"
+#include "os/os.h"
+#include "console/console.h"
+#include "host/ble_hs.h"
+#include "host/util/util.h"
+#include "console/console.h"
+#include "log/log.h"
+
+static uint8_t g_own_addr_type;
+
+static void
+ble_app_set_addr(void)
+{
+    ble_addr_t addr;
+    int rc;
+
+    /* generate new non-resolvable private address */
+    rc = ble_hs_id_gen_rnd(0, &addr);
+    assert(rc == 0);
+
+    /* set generated address */
+    rc = ble_hs_id_set_rnd(addr.val);
+    assert(rc == 0);
+}
+
+/* scan_event() calls scan(), so forward declaration is required */
+static void scan(void);
+
+/* connection has separate event handler from scan */
+static int
+conn_event(struct ble_gap_event *event, void *arg)
+{
+    switch (event->type) {
+    case BLE_GAP_EVENT_CONNECT:
+        if (event->connect.status == 0) {
+            MODLOG_DFLT(INFO,"Connection was established\n");
+            ble_gap_terminate(event->connect.conn_handle, 0x13);
+        } else {
+            MODLOG_DFLT(INFO,"Connection failed, error code: %i\n",
+                event->connect.status);
+        }
+        break;
+    case BLE_GAP_EVENT_DISCONNECT:
+        MODLOG_DFLT(INFO,"Disconnected, reason code: %i\n",
+        event->disconnect.reason);
+        scan();
+        break;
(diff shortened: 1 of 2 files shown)


sjanc:
some general comments on how I'd like to see this sample:

- no need to parse full adv data and print those,  just look for specified UUID and print info when it is found
- then initiate connection
- when connected (or failed) print proper info about  that
- update conn params, notify when updated
- change PHY eg to 2M and notify when done,
- disconnect
- wait for few sconds and go back to 1

apps/central/src/main.c
```diff
@@ -0,0 +1,190 @@
...
+                                             BLE_GAP_LE_PHY_2M_MASK,
+                                             BLE_GAP_LE_PHY_2M_MASK,
+                                             BLE_GAP_LE_PHY_CODED_ANY);
+            if (rc == 0) {
+                MODLOG_DFLT(INFO, "rx and tx PHYs set to 2M\n");
+            } else {
+                MODLOG_DFLT(INFO, "failed to set rx and tx PHYs\n");
+            }
+            ble_gap_terminate(event->connect.conn_handle, 0x13);
+            os_time_ms_to_ticks(delay_ms, &ticks);
```
rymanluk, line 70: os_time_delay(os_time_ms_to_ticks32(2000)); ?
Also please write comment why it is here.

apps/central/src/main.c
```diff
@@ -0,0 +1,190 @@
...
+            rc = ble_gap_set_prefered_le_phy(event->connect.conn_handle,
+                                             BLE_GAP_LE_PHY_2M_MASK,
+                                             BLE_GAP_LE_PHY_2M_MASK,
+                                             BLE_GAP_LE_PHY_CODED_ANY);
+            if (rc == 0) {
+                MODLOG_DFLT(INFO, "rx and tx PHYs set to 2M\n");
+            } else {
+                MODLOG_DFLT(INFO, "failed to set rx and tx PHYs\n");
+            }
+            ble_gap_terminate(event->connect.conn_handle, 0x13);
```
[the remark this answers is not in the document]
Reynevan94: As far as I understand, this GAP event should be handled by peripheral and on central side we don't have easy way to find out if it completed, but if I'm mistaken, I'm happy to change that.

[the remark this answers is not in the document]
rymanluk: Event is sent on  both sides.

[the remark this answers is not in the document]
Reynevan94: I added this feature, but I couldn't get this event called. Maybe other side is configured in a way this change can't be made, so this should be checked.

apps/central/src/main.c
rymanluk: this seems to be unresolved.

Reynevan94: That was simply overlooked and it's fixed now. os_time_delay(os_time_ms_to_ticks32(delay_ms)); works perfect

apps/central/src/main.c
```diff
@@ -0,0 +1,213 @@
...
+#include "os/os.h"
+#include "console/console.h"
+#include "host/ble_hs.h"
+#include "host/util/util.h"
+#include "console/console.h"
+#include "log/log.h"
+
+static uint8_t g_own_addr_type;
+static uint32_t delay_ms = 2000;
+static os_time_t ticks;
```
rymanluk, line 30: this is not needed anymore.

apps/central/src/main.c
```diff
@@ -0,0 +1,213 @@
...
+#include "sysinit/sysinit.h"
+#include "os/os.h"
+#include "console/console.h"
+#include "host/ble_hs.h"
+#include "host/util/util.h"
+#include "console/console.h"
+#include "log/log.h"
+
+static uint8_t g_own_addr_type;
+static uint32_t delay_ms = 2000;
```
rymanluk, line 29: this delay is also not needed here - you ca put is as value where it is used

KKopyscinski: I agree with both - it's fixed now

apps/central/src/main.c
```diff
@@ -0,0 +1,211 @@
...
+                                             BLE_GAP_LE_PHY_CODED_ANY);
+            if (rc == 0) {
+                MODLOG_DFLT(INFO,"PHY update successful\n");
+            } else {
+                MODLOG_DFLT(INFO,"PHY update failed\n");
+            }
+            /* Before calling scan() again, wait predefined time delay_ms.
+               This value has to be converted to ticks first before calling
+               os_time_delay() */
+            os_time_delay(os_time_ms_to_ticks32(2000));
```
sjanc, line 72: you are blocking host task with that, you should not do that

I think for central app we can skip phy updates for now and just have separate sample for playing with PHY (where we could test all combinations, including asymetric)

KKopyscinski: Now waiting is omitted and there's no PHY update, just simple connect -> disconnect -> scan procedure.

apps/central/src/main.c
```diff
@@ -0,0 +1,196 @@
...
+    }
+    return 0;
+}
+
+static int
+scan_event(struct ble_gap_event *event, void *arg)
+{
+    /* predef_uuid stores information about UUID of device,
+       that we connect to */
+    const uint8_t predef_uuid[16] = {
```
sjanc, line 93: why not make this ble_uuid and use stuf from ble_uuid.h to handle uuid operations?

KKopyscinski: In retrospect it's much better and elegant approach.

PR #800, opened 2020-04-17
nimble/ll: Add common setting for SCA

andrzej-kaczmarek:
Currently we have two separate SCA settings: actual SCA as ppm value and
enumerated one based on ranges of SCA value. This means user needs to
update both values and they ideally should be in sync.

To make our life easier, we can just have single BLE_LL_SCA setting to
provide ppm value and enumerated one can be easily determined at build
time.

### DIFF
--- a/nimble/controller/include/controller/ble_ll.h
+++ b/nimble/controller/include/controller/ble_ll.h
@@ -69,6 +69,29 @@ extern "C" {
 /* Timing jitter as per spec is +/16 usecs */
 #define BLE_LL_JITTER_USECS         (16)
 
+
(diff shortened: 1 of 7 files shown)


nimble/controller/include/controller/ble_ll.h
```diff
@@ -69,6 +69,29 @@ extern "C" {
 /* Timing jitter as per spec is +/16 usecs */
 #define BLE_LL_JITTER_USECS         (16)
 
+
```
rymanluk, line 72: spare empty line

PR #801, opened 2020-04-17
Prevent storing of CCCD's if peer is not bonded.

h2zero:
Currently a peer that we have paired with but not bonded will have it's
CCCD status stored. We should not store CCCD data unless bonded,
this will prevent that.

#788

### DIFF
--- a/nimble/host/src/ble_gatts.c
+++ b/nimble/host/src/ble_gatts.c
@@ -1678,28 +1678,29 @@ ble_gatts_bonding_established(uint16_t conn_handle)
 
     conn = ble_hs_conn_find(conn_handle);
     BLE_HS_DBG_ASSERT(conn != NULL);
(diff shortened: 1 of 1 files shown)


rymanluk:
@h2zero thanks for that PR, I  see the issue now which is triggered when bonding is actually not used. 
imho in such a case we shall not call `ble_gatts_bonding_established() ` this is why there is BLE_HS_DBG_ASSERT(conn->bhc_sec_state.bonded)

I went ahead and did other fix - could you please try that?
https://github.com/apache/mynewt-nimble/pull/804

h2zero:
@rymanluk Many ways to skin a cat 😃. #804 is certainly a more correct way that I considered as well, I just didn't have the time and opted for a quick fix. I will close this in favor of yours.

PR #802, opened 2020-04-19
nimble/host: Add return parameter to the ble_hs_misc_conn_chan_find_reqd()

h2zero:
`ble_hs_misc_conn_chan_find_reqd()` did not return an error code if
the connection and or the channel were not found, i.e. in a disconnected state.
When debug is not enabled and `ble_hs_misc_conn_chan_find_reqd()` is called and
the device has disconnected from the peer various functions may attempt to access
memory that is not valid causing an null pointer exception. This changes it to return an 
error code that is tested before accessing the memory,

### DIFF
--- a/nimble/host/src/ble_hs_misc.c
+++ b/nimble/host/src/ble_hs_misc.c
@@ -56,7 +56,7 @@ ble_hs_misc_conn_chan_find(uint16_t conn_handle, uint16_t cid,
     return rc;
 }
 
-void
+int
 ble_hs_misc_conn_chan_find_reqd(uint16_t conn_handle, uint16_t cid,
                                 struct ble_hs_conn **out_conn,
                                 struct ble_l2cap_chan **out_chan)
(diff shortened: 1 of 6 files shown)


nimble/host/src/ble_hs_misc.c
```diff
@@ -66,14 +66,15 @@ ble_hs_misc_conn_chan_find_reqd(uint16_t conn_handle, uint16_t cid,
     int rc;
 
     rc = ble_hs_misc_conn_chan_find(conn_handle, cid, &conn, &chan);
-    BLE_HS_DBG_ASSERT_EVAL(rc == 0);
```
rymanluk, line 69: We could exit here when `rc != 0`

h2zero:
@rymanluk Yeah, sloppy title haha. I will clean this up when I get a chance later today.

nimble/host/src/ble_hs_misc.c
h2zero: Agreed I will add it.

h2zero: Done!

PR #804, opened 2020-04-21
nimble/gap: Fix storing CCC for bonded devices.

rymanluk:
This PR make sure that ble_gatts_bonding_established is called only when
bonded.

Note: renamed not used "persist_keys" to "bonded" and make use of it

Bug can be easly reproduced when
BLE_SM_BONDING: 0
BLE_HS_DEBUG: 1

Program received signal SIGTRAP, Trace/breakpoint trap.
ble_gatts_bonding_established (conn_handle=conn_handle@entry=1) at repos/apache-mynewt-nimble/nimble/host/src/ble_gatts.c:1681
1681	    BLE_HS_DBG_ASSERT(conn->bhc_sec_state.bonded);
(gdb)

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -5654,7 +5654,8 @@ ble_gap_passkey_event(uint16_t conn_handle,
 }
 
 void
-ble_gap_enc_event(uint16_t conn_handle, int status, int security_restored)
+ble_gap_enc_event(uint16_t conn_handle, int status,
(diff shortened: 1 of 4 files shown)


rymanluk:
@h2zero could you please have a look on that?

h2zero:
@rymanluk Just put it through testing and works as it should 👍.

A quick note for @prasad-alatkar for this PR to work on the esp32 we also need #730

PR #805, opened 2020-04-21
nimble/store: Make it clear store/ram is deprecated

rymanluk:
With this patch when store/ram is used then new will warn about it:

2020/04/21 15:51:21.077 [WARNING] Use of deprecated setting BLE_STORE_RAM_DEPRECATED_FLAG in @apache-mynewt-nimble/nimble/host/store/ram: Package store/ram is deprecated and store/config shall be used with BLE_STORE_CONFIG_PERSIST set to 0


rymanluk:
@edoput fyi

PR #806, opened 2020-04-22
nimble/ll: Some LL fixes, cleaning and improvements

rymanluk:


nimble/controller/src/ble_ll_ctrl.c
```diff
@@ -1040,7 +1060,44 @@ ble_ll_ctrl_rx_periodic_sync_ind(struct ble_ll_conn_sm *connsm, uint8_t *dptr)
...
+ *
+ * @param connsm
+ * @param dptr
+ *
+ * @return uint8_t
+ */
+static uint8_t
+ble_ll_ctrl_rx_sca_rsp(struct ble_ll_conn_sm *connsm, uint8_t *dptr)
+{
+    if (!(connsm->cur_ctrl_proc == BLE_LL_CTRL_PROC_SCA_UPDATE)) {
```
sjanc, line 1100: if (foo != BAR)  maybe ?:)

nimble/controller/include/controller/ble_ll_conn.h
```diff
@@ -58,6 +58,8 @@ extern "C" {
 /* Definition for RSSI when the RSSI is unknown */
 #define BLE_LL_CONN_UNKNOWN_RSSI        (127)
 
+#define BLE_LL_CONN_ISO_HANDLE_OFFSET   (0x0100)
```
sjanc, line 61: I'd name this BLE_LL_CONN_HANDLE_ISO_OFFSET (so that is it similar to BLE_HS_CONN_HANDLE_*)

nimble/syscfg.yml
```diff
@@ -77,6 +77,19 @@ syscfg.defs:
             integer for easy comparison.
         range: 50, 51, 52
         value: 50
+    BLE_ISO:
```
sjanc, line 80: I think we should have BLE_LL_FEAT_ISO (or similar) for controller configuration that defaults to BLE_ISO (which should set both host and controller support). This should keep this similar to other LL config options

PR #810, opened 2020-04-30
WIP: use port dependent macros instead of FreeRTOS functions

edoput:
Management of critical section is done through a combination of the FreeRTOS
API and port dependent configuration.

Most of the time the port uses critical section management exposed by FreeRTOS by
defining as such

```C
extern void vPortEnterCritical(void);
#define portENTER_CRITICAL()     vPortEnterCritical();
...
```

Moreover FreeRTOS does not expose an API for critical section detection
as it is port dependent. This will rely on the macro `portIS_IN_CRITICAL` that
a port should implement when using nimble in conjunction with FreeRTOS.


edoput:
I would also like to point in the docs that this macro should be defined by the port
but I'm not finding any docs for the port yet.

PR #815, opened 2020-05-12
nimble/ll: Add controller-to-host flow control support

andrzej-kaczmarek:


nimble/controller/src/ble_ll_hci.c
```diff
@@ -1375,6 +1425,19 @@ ble_ll_hci_ctlr_bb_cmd_proc(const uint8_t *cmdbuf, uint8_t len, uint16_t ocf,
             rc = ble_ll_reset();
         }
         break;
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_CTRL_TO_HOST_FLOW_CONTROL)
+    /*
+     * Note:
+     * we do not handle BLE_HCI_OCF_CB_HOST_NUM_COMP_PKTS here since this
```
rymanluk, line 1431: I would add case and assert along with this comment.

nimble/controller/src/ble_ll_hci.c
```diff
@@ -1343,6 +1343,56 @@ ble_ll_hci_cb_set_event_mask(const uint8_t *cmdbuf, uint8_t len)
     return BLE_ERR_SUCCESS;
 }
 
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_CTRL_TO_HOST_FLOW_CONTROL)
+static int
+ble_ll_hci_cb_set_ctrlr_to_host_fc(const uint8_t *cmdbuf, uint8_t len)
+{
+    const struct ble_hci_cb_ctlr_to_host_fc_cp *cmd = (const void *) cmdbuf;
+
```
rymanluk, line 1351: if (len != sizeof (*cmd)) {
        return BLE_ERR_INV_HCI_CMD_PARMS;
    }

nimble/controller/src/ble_ll_hci.c
```diff
@@ -1343,6 +1343,56 @@ ble_ll_hci_cb_set_event_mask(const uint8_t *cmdbuf, uint8_t len)
...
+    return BLE_ERR_SUCCESS;
+}
+
+static int
+ble_ll_hci_cb_host_buf_size(const uint8_t *cmdbuf, uint8_t len)
+{
+    const struct ble_hci_cb_host_buf_size_cp *cmd = (const void *) cmdbuf;
+    uint16_t acl_num;
+    uint16_t acl_data_len;
+
```
rymanluk, line 1370: also len of command shall be verified.

rymanluk:
In general looks good to me, just couple of comments.

nimble/controller/src/ble_ll_conn.c
```diff
@@ -3598,8 +3769,10 @@ int
 ble_ll_conn_rx_isr_end(uint8_t *rxbuf, struct ble_mbuf_hdr *rxhdr)
 {
     int rc;
-    int is_ctrl;
+    uint8_t is_ctrl;
+    uint8_t is_data;
```
rymanluk, line 3773: looks like `is_data` is used only when
MYNEWT_VAL(BLE_LL_CFG_FEAT_CTRL_TO_HOST_FLOW_CONTROL  is set.

nimble/controller/src/ble_ll_conn.c
```diff
@@ -3612,22 +3611,33 @@ ble_ll_conn_rx_isr_end(uint8_t *rxbuf, struct ble_mbuf_hdr *rxhdr)
...
 
     /* Retrieve the header and payload length */
     hdr_byte = rxbuf[0];
     rx_pyld_len = rxbuf[1];
 
+    /*
+     * No need to alloc rxpdu for packets with invalid CRC, we would throw them
+     * away instantly from LL anyway.
+     */
+    if (!BLE_MBUF_HDR_CRC_OK(rxhdr)) {
```
rymanluk, line 3800: Even we throw away packet in LL, it does not mean that this packet is not important. e.g. when initiating connection using ext commands it is import for us to handle invalid AUX_CONNECT_RSP to do some cleaning and restart scanning. Maybe this is a good idea to handle this way but we need more work on that and in the same time remove all the CRC checks in higher layer, if we decide to drop such packets in here.

rymanluk: ah you are right, we are in "ble_ll_conn_rx_isr_end" which when connection is up. So indeed it should be fine.

PR #819, opened 2020-05-19
Add USB transport to NimBLE controller

kasjer:
This adds USB transport to NimBLE controller.

It requires that https://github.com/hathach/tinyusb/pull/413 is merged on TinyUSB side first.

### DIFF
--- a/nimble/transport/usb/pkg.yml
+++ b/nimble/transport/usb/pkg.yml
@@ -1,0 +1,39 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: nimble/transport/usb
+pkg.description: Provides HCI transport over USB interface
+pkg.author: "Apache Mynewt <dev@mynewt.apache.org>"
+pkg.homepage: "http://mynewt.apache.org/"
+pkg.keywords:
+    - ble
+    - bluetooth
+    - usb
+
+pkg.deps:
+    - "@apache-mynewt-core/hw/hal"
+    - "@apache-mynewt-core/kernel/os"
+    - "@apache-mynewt-core/util/mem"
+    - nimble
+
+pkg.apis:
+    - ble_transport
+
+pkg.init:
+    ble_hci_usb_init: 'MYNEWT_VAL(BLE_TRANS_USB_SYSINIT_STAGE)'
--- a/nimble/transport/usb/src/ble_hci_usb.c
+++ b/nimble/transport/usb/src/ble_hci_usb.c
@@ -1,0 +1,422 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <assert.h>
+#include <stddef.h>
+#include "syscfg/syscfg.h"
+#include "sysinit/sysinit.h"
+#include "os/os.h"
+#include "mem/mem.h"
+
+#include "nimble/ble.h"
+#include "nimble/ble_hci_trans.h"
+#include "nimble/hci_common.h"
+
+#include <class/bth/bth_device.h>
+
+/*
+ * The MBUF payload size must accommodate the HCI data header size plus the
+ * maximum ACL data packet length. The ACL block size is the size of the
+ * mbufs we will allocate.
+ */
+#define ACL_BLOCK_SIZE  OS_ALIGN(MYNEWT_VAL(BLE_ACL_BUF_SIZE) \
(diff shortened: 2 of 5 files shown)


nimble/transport/usb/src/ble_hci_usb.c
```diff
@@ -0,0 +1,420 @@
...
+
+void
+ble_hci_usb_init(void)
+{
+    int rc;
+
+    /* Ensure this function only gets called by sysinit. */
+    SYSINIT_ASSERT_ACTIVE();
+
+    rc = os_mempool_init(&ble_hci_pool_acl_mempool,
```
rymanluk, line 383: these two calls `os_mempool_init` and `os_mbuf_pool_init` you chould replace with a single call `mem_init_mbuf_pool`

nimble/transport/usb/src/ble_hci_usb.c
```diff
@@ -0,0 +1,420 @@
...
+    struct ble_hci_pkt *acl = STAILQ_FIRST(&ble_hci_tx_acl_queue.queue);
+
+    assert(acl != NULL);
+    om = acl->data;
+    assert(om != NULL && om->om_len >= sent_bytes);
+    os_mbuf_adj(om, sent_bytes);
+
+    while (om != NULL && om->om_len == 0) {
+        acl->data = SLIST_NEXT(om, om_next);
+        os_mbuf_free(om);
```
rymanluk, line 151: did you consider to use  `os_mbuf_free_chain` instead?

nimble/transport/usb/src/ble_hci_usb.c
```diff
@@ -0,0 +1,420 @@
...
+        STAILQ_REMOVE_HEAD(&ble_hci_tx_acl_queue.queue, next);
+        os_memblock_put(&ble_hci_pkt_pool, acl);
+        acl = STAILQ_FIRST(&ble_hci_tx_acl_queue.queue);
+        if (acl != NULL) {
+            om = acl->data;
+        }
+    }
+
+    if (om != NULL) {
+        tud_bt_acl_data_send(om->om_data, om->om_len);
```
rymanluk, line 165: It looks like mbuf chaining is not used. Do we make sure somewhere that memory is flat in that mbuf?

nimble/transport/usb/src/ble_hci_usb.c
```diff
@@ -0,0 +1,420 @@
...
+    uint8_t *data;
+    uint32_t len;
+    struct os_mbuf *om = incoming_acl_data;
+    int rc;
+
+    if (om == NULL) {
+        om = ble_hci_trans_acl_buf_alloc();
+        assert(om != 0);
+    }
+    assert(om->om_len + data_len <= MYNEWT_VAL(BLE_ACL_BUF_SIZE) + 4);
```
rymanluk, line 204: ah yes, here it is the check I was asking above.

nimble/transport/usb/src/ble_hci_usb.c
kasjer: I think ```os_mbuf_free_chain()``` can not be used in this context

rymanluk: you are right. Here you iterate on each chain element.

nimble/transport/usb/src/ble_hci_usb.c
kasjer: done

nimble/transport/usb/src/ble_hci_usb.c
```diff
@@ -0,0 +1,415 @@
...
+{
+    uint8_t *buf;
+    int rc = -1;
+
+    buf = ble_hci_trans_buf_alloc(BLE_HCI_TRANS_BUF_CMD);
+    assert(buf != NULL);
+    memcpy(buf, hci_cmd, cmd_len);
+
+    assert(ble_hci_usb_rx_cmd_ll_cb);
+    if (ble_hci_usb_rx_cmd_ll_cb) {
```
andrzej-kaczmarek, line 231: you could just check this at the beginning and skip alloc/free

nimble/transport/usb/pkg.yml
```diff
@@ -0,0 +1,39 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: nimble/transport/usb
+pkg.description: XXX
```
andrzej-kaczmarek, line 21: leave empty or add something more verbose

nimble/transport/usb/include/transport/usb/ble_hci_usb.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+#ifndef H_BLE_HCI_USB_
+#define H_BLE_HCI_USB_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void ble_hci_usb_init(void);
```
andrzej-kaczmarek, line 27: you don't need prototype for pkg_init since newt generates it automatically in sysinit

nimble/transport/usb/src/ble_hci_usb.c
```diff
@@ -0,0 +1,415 @@
...
+tud_bt_acl_data_received_cb(void *acl_data, uint16_t data_len)
+{
+    uint8_t *data;
+    uint32_t len;
+    struct os_mbuf *om = incoming_acl_data;
+    int rc;
+
+    if (om == NULL) {
+        om = ble_hci_trans_acl_buf_alloc();
+        assert(om != 0);
```
andrzej-kaczmarek, line 203: `!= NULL`

nimble/transport/usb/src/ble_hci_usb.c
```diff
@@ -0,0 +1,415 @@
...
+
+    pkt = os_memblock_get(&ble_hci_pkt_pool);
+    if (pkt == NULL) {
+        ble_hci_trans_buf_free(hci_ev);
+        return BLE_ERR_MEM_CAPACITY;
+    }
+
+    pkt->data = hci_ev;
+    OS_ENTER_CRITICAL(sr);
+    bool first = STAILQ_EMPTY(&ble_hci_tx_evt_queue.queue);
```
andrzej-kaczmarek, line 292: local variables should be declared at the beginning

andrzej-kaczmarek: local variables should be declared at the top of function body

nimble/transport/usb/src/ble_hci_usb.c
```diff
@@ -0,0 +1,415 @@
...
+        assert(rc == 0);
+    } else {
+        assert(usb_ble_hci_pool_cmd.allocated);
+        usb_ble_hci_pool_cmd.allocated = 0;
+    }
+}
+
+/**
+ * Unsupported; the USB transport does not have a dedicated ACL data packet
+ * pool.
```
andrzej-kaczmarek, line 362: this is technically incorrect because it does have ACL pool for incoming packets. however this function can be removed entirely because it's used only by NimBLE host so will never be referenced by a controller. and perhaps for this reason would be good to add simple restriction to syscfg or some #error to make sure no one tries to build this with host.

PR #821, opened 2020-05-25
nimble/hs: Fix flow control

andrzej-kaczmarek:
Current implementation of hs flow control uses mbuf user header to
associate mbuf with a connection handle. This unfortunately does not
work well in most cases.

One major problem is that host can reassemble data. This is done by
os_mbuf_concat() which strips packet header from 2nd mbuf and this
means we basically lose connection handle information.

While this problem can be worked around by some extra mbuf counting,
another way to lose connection handle information is to basically pass
received buffer to application. For example, OIC uses user header for
own purposes so it will either overwrite connection handle information
(if our user header is large enough for OIC) or strip our packet header
if it decides to use own mbuf for packet header.

So it's better not to rely on user header to track store connection
handle information. This patch uses a simple private array to store
connection handle for each mbuf - array is indexed by mbuf index in a
mempool. This is a bit of a hack since index of an mbuf in a mempool
has to be calculated using mempool buffer address and block size, but
I do not really see any simple alternative here.

### DIFF
--- a/nimble/host/src/ble_hs_flow.c
+++ b/nimble/host/src/ble_hs_flow.c
@@ -40,6 +40,23 @@ static ble_npl_event_fn ble_hs_flow_event_cb;
 
 static struct ble_npl_event ble_hs_flow_ev;
 
+/* Connection handle associated with each mbuf in ACL pool */
+static uint16_t ble_hs_flow_mbuf_conn_handle[ MYNEWT_VAL(BLE_ACL_BUF_COUNT) ];
+
+static inline int
+ble_hs_flow_mbuf_index(const struct os_mbuf *om)
+{
+    const struct os_mempool *mp = om->om_omp->omp_pool;
+    uintptr_t addr = (uintptr_t)om;
+    int idx;
+
+    idx = (addr - mp->mp_membuf_addr) / mp->mp_block_size;
+
+    BLE_HS_DBG_ASSERT(mp->mp_membuf_addr + idx * mp->mp_block_size == addr);
+
+    return idx;
+}
+
 static int
 ble_hs_flow_tx_num_comp_pkts(void)
 {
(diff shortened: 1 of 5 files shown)


nimble/host/src/ble_hs_flow.c
```diff
@@ -40,6 +40,23 @@ static ble_npl_event_fn ble_hs_flow_event_cb;
...
 static struct ble_npl_event ble_hs_flow_ev;
 
+/* Connection handle associated with each mbuf in ACL pool */
+static uint16_t ble_hs_flow_mbuf_conn_handle[ MYNEWT_VAL(BLE_ACL_BUF_COUNT) ];
+
+static inline int
+ble_hs_flow_mbuf_index(const struct os_mbuf *om)
+{
+    const struct os_mempool *mp = om->om_omp->omp_pool;
+    uint32_t addr = (uint32_t)om;
```
ccollins476ad, line 50: I would use `uintptr_t` here.

nimble/include/nimble/ble.h
```diff
@@ -177,8 +177,8 @@ struct ble_mbuf_hdr
 #define BLE_MBUF_MEMBLOCK_OVERHEAD      \
     (sizeof(struct os_mbuf) + BLE_MBUF_PKTHDR_OVERHEAD)
 
-/* Length of host user header.  Only contains the peer's connection handle. */
-#define BLE_MBUF_HS_HDR_LEN     (2)
+#define BLE_MBUF_HS_HDR_LEN     (0)
```
ccollins476ad, line 180: This comment is definitely in "nitpick territory" so feel free to ignore it.  IMO, this constant reduces clarity without any benefit.  It would be more consistent and easily understood to just hardcode 0 at the allocation site.

andrzej-kaczmarek: I tend to agree, although the reason I left this constant was just in case we want to restore pkthdr later. But since it was removed because it does not quite work as expected on host side, I do not think this will happen any time soon :-)

PR #825, opened 2020-05-28
porting: Update targets and updating script

michal-narajowski:
We need to update pkg names to represent at path to the targets.
Also update the update_generated_files.sh script to
refer to the targets using a path. This is required
after recent change to Newt.


michal-narajowski:
The change in `newt` that caused this: https://github.com/apache/mynewt-newt/pull/381

haukepetersen:
On paper the changes look valid to me. However, I am having trouble to get them running...

So from what I know about `newt` by now, I expected to do the following:
- create a new (empty) project, `newt new foo && cd foo`
- change nimble repo version to `0.0.0` (master) and update repos  `newt upgrade`
- run `./repos/apache-mynewt-nimble/porting/update_generated_files.sh`
- or trigger selected porting targets manually -> `newt build riot`? or `newt build @apache-mynewt-nimble/porting/targets/riot`?

But:
a) how to set the NimBLE repo to use the remote branch from this commit?
b) when calling the target manually from within the `foo` project (`newt build "@apache-mynewt-nimble/porting/targets/riot"`) I get `Error: Could not resolve target name: @apache-mynewt-nimble/porting/targets/riot`
So after all, I seem to be unable to build those targets included in the nimble repo...

michal-narajowski:
@haukepetersen 

a) not sure if you can do that using `newt upgrade` but you can go into nimble repo and either:
- add my fork as a remote and then fetch the branch or
- fetch the PR branch directly `git fetch origin pull/ID/head:BRANCHNAME`

b) That's because the target names need to be changed too. If you open pkg.yml in riot target and change pkg.name to `porting/targets/riot`, then you should be able to build using `newt build @apache-mynewt-nimble/porting/targets/riot`

haukepetersen:
of course, setting the branch manually inside of my newt project is something I didn't think of... I actually spend some time figuring out, how I can configure my project to use my local nimble repository I normally use for development, which is located outside of the newt project...

Anyway, now it works and I can confirm that the changes in this PR are working as expected. Except different values for 
`MYNEWT_VAL_REPO_HASH_APACHE_MYNEWT_CORE`, `MYNEWT_VAL_REPO_HASH_MCUBOOT`, 
`MYNEWT_VAL_REPO_HASH_APACHE_MYNEWT_NIMBLE`, `MYNEWT_VAL_REPO_HASH_APACHE_MYNEWT_MCUMGR`, `MYNEWT_VAL_REPO_HASH_APACHE_MYNEWT_CORE`, and `MYNEWT_VAL_REPO_VERSION_APACHE_MYNEWT_CORE`, my generated files are identical to the ones in this PR.

michal-narajowski:
Actually now that you mentioned it, these `MYNEWT_VAL_REPO_HASH_*` defines kind of defeat the purpose of this script as they will change all the time. I want to use this script in a CI to automatically remind contributors to update these generated files. So as a workaround I will filter out these lines using a script.

michal-narajowski:
@haukepetersen I updated the script and generated files.

haukepetersen:
Works like a charm, the change in #828 now only triggers that one change in the generated syscfg.h file, perfect :-)

Before merging, you might want to move the latest change from the 2nd into the 1st commit, should be easier to track.

PR #826, opened 2020-05-29
host/mesh/src/prov.c fixes

KKopyscinski:
prov_handlers and rx_buf were defined as static, but called only when BLE_MESH_PG_GATT or BLE_MESH_PB_ADV (for prov_handlers) and BLE_MESH_PB_GATT (for rx_buf) were set.
Appropriate preprocessor directives were added for cheching these values, preventing compiling errors (defined but not used).


rymanluk:
@KKopyscinski please check travis errors.

Issue #818, opened 2020-05-18
L2CAP COC: link throughput inconsistent and fluctuating


haukepetersen:
For my research I am lately conducting some experiments involving measuring the raw data throughput of BLE links between selected nodes. While doing so, I noticed that the throughput between two nodes is quite unreliable and showing some funny patterns. To investigate, I simplified my setup so that I currently do the following:
- open a L2CAP connection between two nodes
- open a connection oriented channel using NimBLE's L2CAP API
- let the L2CAP client (`source` node) send a defined number of defined size chunks to the l2cap server (`sink` node)

Below are some exemplary numbers when sending 25000 chunks of 100byte payload each from the source to the `sink`. Each decimal denotes the number of chunks per 250ms that were received at the `sink`, counting every `BLE_L2CAP_EVENT_COC_DATA_RECEIVED` event. One can see clearly, that for some reason the link is throttled to ~8kb/s (80 chunks per sec) for a random amount of time. But then something happens and the remainder of the chunks are send at a much higher speed of ~65kb/s (~650 chunks per sec), which is actually somewhere near of what I would expect as sensible throughput on nrf52-based platforms...

```
0 0 0 0 0 0 0 0 36 20 20 20 20 20 20 20 20 23 20 22 22 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 21 21
22 20 20 20 20 20 20 20 20 20 20 23 20 20 20 20 18 20 19 136 165 125 137 165 165 162 165 166 165 162 165 165 154 165 165 165 166 165 112 165
109 152 165 165 166 165 165 165 165 162 165 165 161 166 166 165 165 165 165 128 165 165 165 162 165 165 163 148 149 166 165 165 165 165 152 164 163 166 159 165
165 133 165 162 162 165 165 141 156 162 165 166 165 148 149 140 160 165 162 151 146 165 164 163 162 164 166 165 162 147 166 149 148 165 135 165 165 165 158 165
158 165 165 165 165 152 164 163 162 165 165 166 142 158 165 165 165 163 75 149 162 165 166 165 132 136 165 165 165 165 162 165 162 166 165 165 165 165 162 165
126 161 165 165 129 149 147 165 150 156 11 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
```
Next run (no node is restarted etc, I simply trigger a shell command to start spamming the link again):
```
[...]

PR #827, opened 2020-06-04
l2cap/coc: make ble_l2cap_coc_continue_tx() thread safe

haukepetersen:
When debugging my favorite L2CAP throughput issues (#818, #373), I sometimes noticed some packets being dropped. In these (mostly rare) cases `l2cap_send()` returned `BLE_HS_ENOMEM` even though there was sufficient space in the used `msys` mbuf pool. 

Looking into this, I found that in `ble_l2cap_coc_continue_tx()` the `chan->coc_tx->data_offset` value as well as the value for `OS_MBUF_PKTLEN(tx->sdu)` were corrupted, leading `os_mbuf_appendfrom()` fail.

See the commit description for my explanation for this:
```
ble_l2cap_coc_continue_tx() is executed in different tasks: it is
(i) called from ble_l2cap_coc_send() in one or more different task
contexts from user application(s) and (ii) called from
ble_l2cap_coc_le_credits_update() from the NimBLE host task.
Typically, the host task has a higher priority then the user tasks,
so that ble_l2cap_coc_continue_tx() can be interrupted by itself.
This leads to potentially inconsistent state for
`chan->coc_tx->credits` and `chan->coc_tx->data_offset`, consequently
leading to dropped packets.
This commit fixes this by putting all the code code working on these
variables into a critical section.
```

This solution feels a little bit brute force, but I could not think of anything else to make `ble_l2cap_coc_continue_tx()` thread safe. My initial idea was to put an event into the hosts event queue from `ble_l2cap_coc_send()`, instead of calling `ble_l2cap_coc_continue_tx()` directly. But then we would need to allocate an event for each l2cap channel. But more severe we would not be able to synchronously pass the return code of `ble_l2cap_coc_continue_tx()` to `ble_l2cap_coc_send()`. So it seemed to me locking the host for the duration of `ble_l2cap_coc_continue_tx()` is the easier path?!

### DIFF
--- a/nimble/host/src/ble_l2cap_coc.c
+++ b/nimble/host/src/ble_l2cap_coc.c
@@ -469,12 +469,12 @@ ble_l2cap_coc_continue_tx(struct ble_l2cap_chan *chan)
             txom = NULL;
             goto failed;
         } else {
-            tx->credits --;
+            tx->credits--;
             tx->data_offset += len - sdu_size_offset;
         }
 
         BLE_HS_LOG(DEBUG, "Sent %d bytes, credits=%d, to send %d bytes \n",
-                  len, tx->credits, OS_MBUF_PKTLEN(tx->sdu)- tx->data_offset );
+                  len, tx->credits, OS_MBUF_PKTLEN(tx->sdu)- tx->data_offset);
 
         if (tx->data_offset == OS_MBUF_PKTLEN(tx->sdu)) {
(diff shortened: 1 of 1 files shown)


haukepetersen:
fixed two more style issues pointed out by the `apache-mynewt-bot`

haukepetersen:
I am still trying to come up with a more elegant solution, but so far I was not able to make anything work. So lets go for now.

PR #828, opened 2020-06-04
npl/riot: change dflt value for BLE_ACL_BUF_COUNT

haukepetersen:
As stated in #818, the old default value of `4` did lead to degraded performance concerning link throughput. The new value of `24` also seems to be the default value for other NimBLE ports, so it seems valid to me to use that as default for the RIOT port as well. Also this value has no impact on memory usage, so it seems to me there is no penalty in doing so.

For this PR I edited both files manually, as I was not able to re-generate the syscfg.h file using the newt tool. The steps listed in #648 were not working for me: I was not able to build the `riot` target included in the nimble repo (using Apache Newt 1.9.0-dev / 0886033 / 2020-06-04_18:03).


michal-narajowski:
Hi @haukepetersen, recently there was a change in newt tool that broke the update script. Please take a look at https://github.com/apache/mynewt-nimble/pull/825.

haukepetersen:
rebased this PR on top of #825 and used the `update_generated_files.sh` script to re-generate syscfg headers.

As of now, I blindly committed all generated files, also including non-related ones for non-riot ports. I have to say I don't like it though. Maybe we can find a way that those version related values (as also pointed out in #825) are ignored or handled otherwise?

haukepetersen:
updated the PR on base of the latest changes in #825 -> changes for the regenerated riot syscfg.h file look much better now!

haukepetersen:
#825 is merged. Rebased this PR on master, so should be good for merge :-)

haukepetersen:
And go.

PR #829, opened 2020-06-05
nimble/host: check read permission in ble_att_svr_build_read_mult_rsp

KKopyscinski:
'ble_att_svr_build_read_mult_rsp()' always tried to read without
checking permissions. If attribute had no read permissions,
ble_att_svr_read_handle() returned an error. This approach wasn't
sufficient for cases, in which long attribute with read permission was
followed by another without this permission. Because iteration over the
list handles happened only when both handles are left and response
isn't full, filling the response caused, that second attribute was
never checked for permissions, and success was returned instead of
ATT: Error Response. Now, permissions are checked first, and then read
is performed.

This was affecting GATT/SR/GAR/BI-21-C

### DIFF
--- a/nimble/host/src/ble_att_svr.c
+++ b/nimble/host/src/ble_att_svr.c
@@ -1557,6 +1557,8 @@ ble_att_svr_build_read_mult_rsp(uint16_t conn_handle,
                                 uint16_t *err_handle)
 {
     struct os_mbuf *txom;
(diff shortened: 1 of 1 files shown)


KKopyscinski:
@rymanluk Could you review this?

KKopyscinski:
Updated commit message and rebased commit

KKopyscinski:
As of PTS 8.0.2 this is not required

PR #836, opened 2020-06-29
LL: Fix few issues in scanner

andrzej-kaczmarek:


nimble/controller/src/ble_ll_scan.c
```diff
@@ -3021,7 +3021,8 @@ ble_ll_scan_rx_pkt_in_on_legacy(uint8_t pdu_type, struct os_mbuf *om,
 
     if (!BLE_MBUF_HDR_DEVMATCH(hdr) ||
         !BLE_MBUF_HDR_CRC_OK(hdr) ||
-        BLE_MBUF_HDR_IGNORED(hdr)) {
+        BLE_MBUF_HDR_IGNORED(hdr) ||
+        !scansm->scan_enabled) {
```
rymanluk, line 3025: isn't ignored in  ble_ll_scan_rx_isr_end()?

 ```
case BLE_ADV_PDU_TYPE_ADV_EXT_IND:
        rc = ble_ll_scan_rx_isr_on_aux(pdu_type, rxbuf, hdr, &addrd);
        break;
```

rc will be -1 on legacy scan

rymanluk: Ok, this is handling situation when scan is disabled while we can have already received data. Therefore, this fix make sense.

PR #837, opened 2020-06-30
Allow BLE_ROLE_OBSERVER without BLE_ROLE_CENTRAL or BLE_EXT_ADV

hasty:
If BLE_ROLE_OBSERVER is set, but neither BLE_ROLE_CENTRAL or BLE_EXT_ADV is set, the build fails with a missing ble_gap_master_set_timer. This adds BLE_ROLE_OBSERVER to the list for ble_gap_master_set_timer

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -998,7 +998,7 @@ ble_gap_update_next_exp(int32_t *out_ticks_from_now)
 
 }
 
-#if MYNEWT_VAL(BLE_ROLE_CENTRAL) || MYNEWT_VAL(BLE_ROLE_OBSERVER)
+#if NIMBLE_BLE_SCAN
 static void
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_gap.c
```diff
@@ -998,7 +998,7 @@ ble_gap_update_next_exp(int32_t *out_ticks_from_now)
 
 }
 
-#if MYNEWT_VAL(BLE_ROLE_CENTRAL)
+#if MYNEWT_VAL(BLE_ROLE_CENTRAL) || MYNEWT_VAL(BLE_ROLE_OBSERVER)
```
rymanluk, line 1001: maybe we just need here `#if NIMBLE_BLE_SCAN` ?

hasty: Yep, that's cleaner. Updated the PR.

PR #838, opened 2020-07-02
apps: privacy added

KKopyscinski:
This commit adds two applications working together in similar way central and
peripheral did; the difference is, they are waiting until RPA timeout expires
before reconnecting.


sjanc:
ok, so this is not testing privacy, I think to test it properly boards should connect (by uuid), pair and get ID address of peer, disconnect, wait for RPA to rotate and then reconnect without using UUID for discovery, once reconnected verify that RPA changed but resolves to ID

apps/privacy_cent/pkg.yml
```diff
@@ -0,0 +1,38 @@
...
+    - "@apache-mynewt-core/kernel/os"
+    - "@apache-mynewt-core/sys/console/full"
+    - "@apache-mynewt-core/sys/log/full"
+    - "@apache-mynewt-core/sys/stats/full"
+    - "@apache-mynewt-nimble/nimble/host"
+    - "@apache-mynewt-nimble/nimble/host/util/"
+    - "@apache-mynewt-nimble/nimble/host/store/config"
+    - "@apache-mynewt-nimble/nimble/transport"
+    - "@apache-mynewt-nimble/nimble/host/services/gap"
+    - "@apache-mynewt-nimble/nimble/controller"
```
andrzej-kaczmarek, line 37: there should be no explicit dependency to controller - it will be selected by transport package if needed

PR #843, opened 2020-07-13
nimble/porting: Explicitly NULL terminate string buffer after using strncpy.

prasad-alatkar:
os_mempool.c: Using `strncpy`, buffer will not be `NULL` terminated if size of source string is same as the buffer size. Hence relpace `strncpy` with `strlcpy`.

### DIFF
--- a/porting/nimble/src/os_mempool.c
+++ b/porting/nimble/src/os_mempool.c
@@ -336,7 +336,8 @@ os_mempool_info_get_next(struct os_mempool *mp, struct os_mempool_info *omi)
     omi->omi_num_blocks = cur->mp_num_blocks;
     omi->omi_num_free = cur->mp_num_free;
     omi->omi_min_free = cur->mp_min_free;
-    strncpy(omi->omi_name, cur->name, sizeof(omi->omi_name));
+    strncpy(omi->omi_name, cur->name, sizeof(omi->omi_name) - 1);
+    omi->omi_name[sizeof(omi->omi_name) - 1] = '\0';
 
     return (cur);
 }


porting/nimble/src/os_mempool.c
```diff
@@ -336,7 +336,7 @@ os_mempool_info_get_next(struct os_mempool *mp, struct os_mempool_info *omi)
     omi->omi_num_blocks = cur->mp_num_blocks;
     omi->omi_num_free = cur->mp_num_free;
     omi->omi_min_free = cur->mp_min_free;
-    strncpy(omi->omi_name, cur->name, sizeof(omi->omi_name));
+    strlcpy(omi->omi_name, cur->name, sizeof(omi->omi_name));
```
andrzej-kaczmarek, line 339: `strlcpy` is not a standard library function so keep `strncpy` and terminate manually instead

prasad-alatkar: @andrzej-kaczmarek Just pushed the changes using `strncpy` only after seeing travis build failures. Thanks :smiley:

h2zero:
> this should be fixed in mynewt-core and then ported to mynewt-nimble

@andrzej-kaczmarek this was fixed in mynewt-core quite some time ago apache/mynewt-core#1865

andrzej-kaczmarek:
> > this should be fixed in mynewt-core and then ported to mynewt-nimble
> 
> @andrzej-kaczmarek this was fixed in mynewt-core quite some time ago [apache/mynewt-core#1865](https://github.com/apache/mynewt-core/pull/1865)

ok, I did not know that. so file in mynewt-nimble should be just updated to match mynewt-core instead of making different fix here.

PR #844, opened 2020-07-17
Fix big-endian serialization issues

hbcam:
I have been porting NimBLE to a big endian system, and found two endianness issues: one missing cpu to le, and a `htole16` on a `uint8_t` value.

First PR related to #847.

### DIFF
--- a/nimble/host/src/ble_att_svr.c
+++ b/nimble/host/src/ble_att_svr.c
@@ -1360,7 +1360,7 @@ done:
         *att_err = 0;
 
         /* Fill the response base. */
-        rsp->batp_length = htole16(sizeof(*data) + prev_attr_len);
+        rsp->batp_length = sizeof(*data) + prev_attr_len;
     }
 
     *out_txom = txom;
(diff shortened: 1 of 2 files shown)


nimble/host/src/ble_att_svr.c
```diff
@@ -1360,7 +1360,7 @@ ble_att_svr_build_read_type_rsp(uint16_t conn_handle,
         *att_err = 0;
 
         /* Fill the response base. */
-        rsp->batp_length = htole16(sizeof(*data) + prev_attr_len);
```
sjanc, line 1363: Hi, I'm not sure if I get this one on how it would fix anything, could you explain it a bit?

hbcam: `batp_length` type is `uint8_t`, so calling `htole16` basically discards the value on a big-endian system.

sjanc: ahh right!

PR #846, opened 2020-07-24
apps/blestress - compilation error

KKopyscinski:
Building the app returned error: multiple definition of `stress_timer_callout'. Moving the it's declaration to stress.c fixes the issue - this structure is only used there.

### DIFF
--- a/apps/blestress/src/stress.c
+++ b/apps/blestress/src/stress.c
@@ -19,6 +19,8 @@
 
 #include "stress.h"
 
+static struct os_callout stress_timer_callout;
+
 void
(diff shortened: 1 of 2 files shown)


apps/blestress/src/stress.c
```diff
@@ -19,6 +19,8 @@
 
 #include "stress.h"
 
+struct os_callout stress_timer_callout;
```
rymanluk, line 22: don't we want it static then?

KKopyscinski: Yes, I agree

PR #849, opened 2020-07-27
apps/blestress - test 10 (L2CAP) fix

KKopyscinski:
In this test L2CAP frames are badly send and/or formed. Changes contain:
- switch from PSM = 1 to PSM = 0x80 (128) - from dynamic range
- in `stress_fill_mbuf_with_pattern()` filling mbuf begins at start of pattern
  buffer, as the full length of that buffer if being copied
- removed appending length of pattern data, as that is performed either way
  by functions called by l2cap API

### DIFF
--- a/apps/blestress/src/rx_stress.c
+++ b/apps/blestress/src/rx_stress.c
@@ -1365,7 +1365,7 @@ rx_stress_start(int test_num)
         break;
     case 10:
         console_printf("Stress L2CAP send\033[0m\n");
-        rc = ble_l2cap_create_server(1, STRESS_COC_MTU,
+        rc = ble_l2cap_create_server(TEST_PSM, STRESS_COC_MTU,
                                      rx_stress_10_l2cap_event, NULL);
         assert(rc == 0);
         rx_stress_simple_adv(&rx_stress_adv_sets[10]);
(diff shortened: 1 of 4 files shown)


apps/blestress/src/rx_stress.c
```diff
@@ -1365,7 +1357,7 @@ rx_stress_start(int test_num)
         break;
     case 10:
         console_printf("Stress L2CAP send\033[0m\n");
-        rc = ble_l2cap_create_server(1, STRESS_COC_MTU,
+        rc = ble_l2cap_create_server(0x80, STRESS_COC_MTU,
```
rymanluk, line 1360: this could be #define TEST_PSM

apps/blestress/src/rx_stress.c
```diff
@@ -837,14 +837,6 @@ rx_stress_10_l2cap_event(struct ble_l2cap_event *event, void *arg)
     MODLOG_DFLT(INFO, "Data buf %s\n", data_buf ? "OK" : "NOK");
     assert(data_buf != NULL);
 
-    /* The first 2 bytes of data is the size of appended pattern data. */
```
rymanluk, line 840: indeed l2cap uses 2 octects with a size, but maybe this was used for the test purposes?

KKopyscinski: I don't think so - the other board doesn't really read that data, just receives it, detects MTU and calculates bitrate. In theory, this could be used to check the length of the data - but in current state it's done just by using OS_MBUF_PKTLEN().

PR #854, opened 2020-08-28
apps/blestress: error correction

KKopyscinski:
Failed sending data via L2CAP in test 10 was always assumed to be
BLE_HS_ESTALLED error, and it's not always the case.


rymanluk (approved):
lgtm

PR #856, opened 2020-08-31
controller: properly seed srand() during initialization

haukepetersen:
This one took me a long time to find and it had devastating effects: the STDLIBs pseudo random number generator was not properly seeded by the controller. As the comment in the code, wisely starting with `XXX`, perfectly stated: `This really doesn't belong here, as the address probably has not been set yet.`. At least when using the NimBLE controller with RIOT, that comment is absolutely true... So when running multiple nodes, all use the same seed making their random number identical.

When running my deployment using 15 nRF52xxx nodes in a multi-hop network, I observed the following faulty behaviors:
- application data was duplicated and received on nodes that did not have a direct connection to nodes sending this data
- L2CAP COC flow control getting out of sync: RX or TX credit counts exceeding their initial limits

As far as I can tell by now, this was all caused by multiple connections using the same access addresses, in term caused by all nodes using the same seed value.

Now this PR fixes the seeding issue by simply using the hardware random number generator to seed the pseudo random number generator. This way it should be guaranteed, that nodes always use different seed values. Also the runtime-overhead should be acceptable, since this is only called once during controller initialization.

### DIFF
--- a/nimble/controller/src/ble_ll.c
+++ b/nimble/controller/src/ble_ll.c
@@ -1203,8 +1203,6 @@ ble_ll_task(void *arg)
     /* Tell the host that we are ready to receive packets */
     ble_ll_hci_send_noop();
 
-    ble_ll_rand_start();
-
     while (1) {
         ev = ble_npl_eventq_get(&g_ble_ll_data.ll_evq, BLE_NPL_TIME_FOREVER);
         assert(ev);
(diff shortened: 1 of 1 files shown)


nimble/controller/src/ble_ll.c
```diff
@@ -1682,15 +1663,17 @@ ble_ll_init(void)
...
-     * been set yet.
-     */
-    ble_ll_seed_prng();
-
-    lldata->ll_supp_features = features;
+    /* Start the random number generator */
+    ble_ll_rand_start();
+    /* Use the random number generator to seed the STDLIBs pseudo-number
+     * generator */
+    unsigned seed;
```
sjanc, line 1674: good catch!
nitpick : declare seed at function top :)

haukepetersen: done

haukepetersen:
addressed comment.

sjanc (approved):

haukepetersen:
@sjanc any chance you could verify the current version? Would be nice to have this in :-)

sjanc:
this needs to wait until https://github.com/apache/mynewt-nimble/pull/863 is merged
otherwise unit tests will fail

andrzej-kaczmarek:
@haukepetersen #863 is merged so CI should be happy if you rebase it now

haukepetersen:
done :-)

sjanc (approved):

haukepetersen:
all green -> go.

PR #857, opened 2020-09-01
nimble/hs: remove recursive call

corecode:
On esp32, tail call optimization does not exist, which can lead to a
stack overflow.

### DIFF
--- a/nimble/host/src/ble_hs_conn.c
+++ b/nimble/host/src/ble_hs_conn.c
@@ -481,29 +481,52 @@ ble_hs_conn_timer(void)
     int32_t time_diff;
     uint16_t conn_handle;
 
-    conn_handle = BLE_HS_CONN_HANDLE_NONE;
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_hs_conn.c
```diff
@@ -552,7 +554,7 @@ ble_hs_conn_timer(void)
      */
     if (conn_handle != BLE_HS_CONN_HANDLE_NONE) {
         ble_gap_terminate(conn_handle, BLE_ERR_REM_USER_CONN_TERM);
-        return ble_hs_conn_timer();
+        goto again;
```
andrzej-kaczmarek, line 557: you need to update comment as well
and `while` or `do {} while` would be nicer than `goto`

corecode: will do!

corecode:
How do you like this?

andrzej-kaczmarek (approved):

PR #859, opened 2020-09-03
Enable NimBLE on Dialog CMAC

andrzej-kaczmarek:
This adds phy, transport and some other changes required to run full NimBLE controller on Dialog CMAC. This means we will now be able to run NimBLE controller on Dialog DA1469x MCU family instead of using binary image with proprietary controller.


sjanc:
Is tinycrypt copy necessary? We already have it in ext/

wes3 (approved):
Took a quick look at things and it looks reasonable to me

andrzej-kaczmarek:
> Is tinycrypt copy necessary? We already have it in ext/

for now - yes. I'll dealt with it later.

sjanc:
OK, just add note to LICENSE about TC seconds copy

sjanc (approved):

sjanc (approved):

benpicco:
Hi, do you think this could also be used for the DA1458x MCU family or are the differences too big?

PR #861, opened 2020-09-07
host/l2cap_coc: harden TX thread safety

haukepetersen:
I also stumbled upon some more non-thread-safe lines in the `ble_l2cap_coc` code while debugging erratic node behavior in the context of #865.

Mainly the operation of setting `tx->sdu = sdu_tx;` in `ble_l2cap_coc_send()`, called by any user thread, could theoretically be interrupted from the NimBLE host thread and be overwritten before `ble_l2cap_coc_send()` returns. This is now prevented by doing this assignment inside a locked host. 

To further prevent any context switch to interfere, I moved both calls to `ble_l2cap_event_coc_unstalled()` inside a locked host environment. With this it is guaranteed, that any changes to the sending state (possible by `ble_l2cap_coc_send()` or `ble_l2cap_coc_le_credits_update()`) are always complete before any interruption can occur.

I have been successfully using this fix together with #865 for a little while now in my experiment runs.

### DIFF
--- a/nimble/host/src/ble_l2cap_coc.c
+++ b/nimble/host/src/ble_l2cap_coc.c
@@ -392,6 +392,8 @@ ble_l2cap_event_coc_unstalled(struct ble_l2cap_chan *chan, int status)
     chan->cb(&event, chan->cb_arg);
 }
 
+/* WARNING: this function is called from different task contexts. We expect the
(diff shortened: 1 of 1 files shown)


haukepetersen:
fixed style issues.

rymanluk (approved):
LGTM

haukepetersen:
Thanks! Lets go then...

PR #866, opened 2020-09-21
NimBLE host: Change min/max macros to inline functions for C++

prasad-alatkar:
In case of C++ project, the min/max macros defined in `porting/nimble/include/os/os.h` can cause compilation error if the C++ project also has min/max function definition in namespace/template. This issue happens because these min/max macros will simply replace the function definition in other part of code during preprocessing. The error seen for me is as below:
```
error: expected unqualified-id before 'const'
 inline const _T & min(const _T & a, const _T & b)
                       ^~~~~
/esp-idf/components/bt/host/nimble/nimble/porting/nimble/include/os/os.h:30:21: note: in definition of macro 'min'
 #define min(a, b) ((a)<(b)?(a):(b))
```  
Reference for solution: https://stackoverflow.com/a/51642531/6100298

### DIFF
--- a/porting/nimble/include/os/os.h
+++ b/porting/nimble/include/os/os.h
@@ -24,14 +24,28 @@
 
 #ifdef __cplusplus
 extern "C" {
-#endif
 
+inline int
+max(int a, int b)
+{
+    return((a) > (b) ? a : b);
+}
+
+inline int
+min(int a, int b)
+{
+    return((a) < (b) ? a : b);
+}
+
+#else
 #ifndef min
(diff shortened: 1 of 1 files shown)


andrzej-kaczmarek (changes requested):
this won't work as expected if you want min/max on unsigned type (e.g. uint32_t) because both values will be converted to signed type. you could try templates here instead but I do not really know if they work with C code.

prasad-alatkar:
Yes @andrzej-kaczmarek this won't work for uint32_t and templates do not work for C code. Thank you for your response, will close this PR.

PR #867, opened 2020-09-30
ble_store_config: building error if BLE_STORE_CONFIG_PERSIST set to 0

KKopyscinski:
As apache-mynewt-core/sys/config is dependency of
BLE_STORE_CONFIG_PERSIST, it should only be included when
BLE_STORE_CONFIG_PERSIST is set to 1, preventing build errors.

### DIFF
--- a/nimble/host/store/config/src/ble_store_config.c
+++ b/nimble/host/store/config/src/ble_store_config.c
@@ -23,7 +23,9 @@
 #include "sysinit/sysinit.h"
 #include "syscfg/syscfg.h"
 #include "host/ble_hs.h"
+#ifdef BLE_STORE_CONFIG_PERSIST
 #include "config/config.h"
+#endif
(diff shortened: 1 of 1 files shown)


nimble/host/store/config/src/ble_store_config.c
```diff
@@ -23,7 +23,9 @@
 #include "sysinit/sysinit.h"
 #include "syscfg/syscfg.h"
 #include "host/ble_hs.h"
+#ifdef BLE_STORE_CONFIG_PERSIST
```
kasjer, line 26: That was also my first solution then I found out that it seems that content of included file is not used at all, hence #889

KKopyscinski: Ok, so as your PR has better solution I think this one can be closed.

Issue #858, opened 2020-09-02
Possible NULL dereference in porting/nimble/src/os_mempool.c


tluio:
On line 81, the pointer **membuf** is checked in 
  if (membuf != NULL)
This indicates that **membuf** can be NULL.
If so, a NULL dereference bug will occur on line 114:
  SLIST_NEXT(block_ptr, mb_next)

sjanc:
if membuf is NULL, blocks must be 0 (otherwise function returns in line 81) so that while() never loops and thus SLIST_NEXT(block_ptr, mb_next) is not called

tluio:
Sorry, but I wonder if our code is the same? If **blocks** is 0, SLIST_NEXT(block_ptr, mb_next) is also called in line 114.

sjanc:
Looks like this was already fixed in core but nimble copy wasn't updated. We should probably sync copy with core

KKopyscinski:
Fixed in merged patch

PR #869, opened 2020-10-02
porting/nimble/src/os_mempool - prevent possible NULL dereference

KKopyscinski:
This ports the core implementation of os_mempool_init(), preventing
possible dereference of NULL pointer if provided mbuf is NULL.

This fixes https://github.com/apache/mynewt-nimble/issues/858


KKopyscinski:
As disgussed offline, whole directory should be synced, not only single function.

PR #871, opened 2020-10-06
nimble/ll: Fix stuck in the encryption process

rymanluk:
Background:
When Central sends LL_ENC_REQ, Peripheral is allowed to send unencrypted
data until LL_ENC_RSP is sent back to Central. This is why, Nimble LL puts
LL_ENC_RSP to the tail of the conn_txq.

Issue description:
When lot of traffic is ongoing in the link while encryption, we might
stuck in the process.
This is because enc_state CONN_ENC_S_LTK_REQ_WAIT is set just after
LL_ENC_REQ has been received. If there are still packets in the conn_txq
to send, those will not be sent because of the enc_state. Instead empty
packets will be send and LL_ENC_RSP will stuck in the conn_txq.

Fix:
Add addition state for Peripheral role and make sure that LL is able to
send unencypted packets from its conn_txq until LL_ENC_RSP is sent.


nimble/controller/src/ble_ll_conn.c
```diff
@@ -1015,8 +1015,14 @@ ble_ll_conn_tx_pdu(struct ble_ll_conn_sm *connsm)
...
          * kinds of LL control PDU's. If none is enqueued, send empty pdu!
+         *
+         * In Slave role, we are allowed to send unencrypted packets until
+         * LL_ENC_RSP is sent.
          */
-        if (connsm->enc_data.enc_state > CONN_ENC_S_ENCRYPTED) {
+        if (((connsm->enc_data.enc_state > CONN_ENC_S_ENCRYPTED) &&
+              CONN_IS_MASTER(connsm)) ||
+              ((connsm->enc_data.enc_state > CONN_ENC_S_ENC_RSP_TO_BE_SENT) &&
+              !CONN_IS_MASTER(connsm))) {
```
sjanc, line 1025: we also have CONN_IS_SLAVE macro :)

andrzej-kaczmarek (approved):
lgtm. no regressions verified on le harmony.

Issue #864, opened 2020-09-16
Porting nimBLE to NuttX


v01d:
I'm experimenting with porting nimBLE to NuttX. We're in the process of completing a Linux-like socket interface via BTPROTO_HCI and BTPROTO_L2CAP. There's already a host layer on NuttX but I would like to try nimBLE as it already supports Linux and it shouldn't be much harder to support.

I'm finding some issues that I wanted to mention, not sure if these are possibles areas of improvement of nimBLE or I'm simply not fully understanding how to do the port (I couldn't find any documentation). 

For starters, I copied linux/ in npl/ and modified it to suit NuttX. As it is mostly Linux compatible, no big changes are needed.  ~~However one issue I find is that NuttX requires an initial include in every .c and .h as: `#include <nuttx/config.h>` which holds the output of the Kconfig configuration system. For the OS specific files in npl/ I can add this include, but for the rest of the code I would have to manually insert it everywhere. I imagine one way would be to extend the porting layer to have a header which would be included everywhere in nimBLE where I could add this include ([LVGL](https://github.com/littlevgl/lvgl) for example does it this way).~~ This is not really a problem

Another issue I find is there there are some macros which are not defined anywhere in nimBLE repo. For now I found (but I think there are many others):
* NIMBLE_CFG_CONTROLLER
* OS_CFG_ALIGN_4
* OS_CFG_ALIGNMENT
* BLE_LL_CFG_FEAT_LL_PRIVACY
* LOG_LEVEL_*

I'm sensing that there should be some configuration process for the port which defines these but it does not appear to be part of this repo (maybe done in mynewt repo?)

I also started by attempting to build the linux "test" applications but it failed due to `ble_npl_eventq_run` not being declared. Maybe this test is not maintained? If you don't think it is reasonable to spend time trying to build this test I can skip it. I was now moving on to trying to build the linux "example".

One final question: in Linux there's BTPROTO_L2CAP and on NuttX we're following the same path of exposing L2CAP over a socket. I see nimBLE only uses the HCI socket and solves L2CAP on its own. Is there any plan to make this optional and have BTPROTO_L2CAP sockets supported? This would not only better integrate nimble with NuttX but also Linux IMHO.

That's it so far. If you could give me some points on the porting process it would really be helpful.

v01d:
[...]

PR #878, opened 2020-10-26
Add support for NuttX

v01d:
This PR adds support for Apache NuttX RTOS. As NuttX is POSIX compatible, I mostly used the Linux porting layer unchanged and only introduce some differences (I also wrote the event queue using POSIX message queues, which removed the need for the C++ file). Support for the required interface in NuttX has just been merged so this depends on master version of NuttX at the moment. nimBLE is used on NuttX just for its host layer.

I'm marking this as draft because I would like to get some feedback in order to check that I'm in the right direction and then clean it up a bit (there are some changes that you may want to be merged separately). Also, I needed to distinguish Linux from NuttX case in the HCI socket interface file but I don't know how to create a new variable to do so, so I would need to fix that as well.

I've been using nimBLE extensively on NuttX, both in the "sim" build (where NuttX uses the PC's bluetooth stack and simply passes the HCI messages around) and on nRF52832 chip (I wrote a link-layer for NuttX which is not yet merged).

Integration will be completed in NuttX by having an extra "nimble" application which downloads this repo and builds the required userspace component (the host-layer).


nimble/host/src/ble_hs_hci.c
```diff
@@ -339,11 +339,13 @@ ble_hs_hci_cmd_tx(uint16_t opcode, const void *cmd, uint8_t cmd_len,
 static void
 ble_hs_hci_rx_ack(uint8_t *ack_ev)
 {
+#if 0
     if (ble_npl_sem_get_count(&ble_hs_hci_sem) > 0) {
         /* This ack is unexpected; ignore it. */
         ble_hci_trans_buf_free(ack_ev);
         return;
     }
+#endif
```
v01d, line 348: This is ifdefed since it is a race condition

v01d:
Ok, I just addressed the style issues and also found how to define a variable to distinguish NuttX from Linux. There are just two things worth mentioning:
- a check on a semaphore which is actually a race condition (see annotation above) since it checks the semaphore at a time which may be incorrect (in NuttX this was triggered incorrectly, possibly on Linux it was slower and it wasn't triggered)
- I had to disable LOG_* defines in log.h, due to conflicts of syslog() defines which NuttX has. It was suggested in #864 that I could redefine include path to my own log.h, and to do this I would have to change all includes of `log.h` from using `"` to `<`/`>` for that to work. But I feel hesitant to do so since it sounds a bit hack to do that only for that include. If you confirm it is OK to do so, I will add that as well. Otherwise, I think the simpler change is to remove those LOG_* macros which are not used in nimBLE (or maybe conditionally define them based on MYNEWT?)

nimble/host/services/bas/include/services/bas/ble_svc_bas.h
```diff
@@ -28,4 +28,6 @@
 
 int ble_svc_bas_battery_level_set(uint8_t level);
 
+void ble_svc_bas_init(void);
```
v01d, line 31: This declaration was missing

v01d:
I tried fixing the remaining style issues but I'm not sure if you actually want me to change the alignment on that macro definition on log_common.h

v01d:
Hi @ccollins476ad @andrzej-kaczmarek any chance you could take a look at the PR and offer some feedback?

v01d:
Hi, sorry for insisting but is there a chance for this to be reviewed? I don't think it is really that different than the Linux case.
I also would like to continue BLE work on NuttX using nimBLE and having this integrated upstream would really help.

andrzej-kaczmarek:
@v01d sorry it takes so long, I was quite busy recently but I should find some time to check it this week

v01d:
great, thanks @andrzej-kaczmarek

nimble/host/src/ble_hs_hci.c
sjanc: could you comment on this a bit? it looks like this code was here since day 1

porting/examples/nuttx/README.md
```diff
@@ -0,0 +1,73 @@
...
+## Overview
+
+See (https://mynewt.apache.org/network/ble/ble_intro/).
+
+## Building
+
+NimBLE is usually built as a part of Apache Mynewt OS, but ports for
+other RTOS-es are also available.
+
+### Linux
```
sjanc, line 33: this readme should be either updated to reflect nuttx env or removed

porting/examples/nuttx/include/syscfg.h
```diff
@@ -0,0 +1,941 @@
+/**
```
sjanc, line 1: is this file needed?

porting/nimble/include/log/log.h
```diff
@@ -30,11 +30,13 @@ log_dummy(void *log, ...)
     (void)log;
 }
 
+#if 0
```
sjanc, line 33: why is this needed?

porting/targets/nuttx/pkg.yml
```diff
@@ -0,0 +1,40 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+pkg.name: "porting/targets/nuttx"
+pkg.type: "target"
+pkg.description: This target is used to generate syscfg.h file and other artifacts for linux example app.
```
sjanc, line 21: "for nuttx example app"

nimble/transport/socket/src/ble_hci_socket.c
```diff
@@ -205,6 +208,8 @@ static struct ble_hci_sock_state {
 static int s_ble_hci_device = MYNEWT_VAL(BLE_SOCK_TCP_PORT);
 #elif MYNEWT_VAL(BLE_SOCK_USE_LINUX_BLUE)
 static int s_ble_hci_device = MYNEWT_VAL(BLE_SOCK_LINUX_DEV);
+#elif MYNEWT_VAL(BLE_SOCK_USE_LINUX_BLUE) || MYNEWT_VAL(BLE_SOCK_USE_NUTTX)
```
andrzej-kaczmarek, line 211: `MYNEWT_VAL(BLE_SOCK_USE_LINUX_BLUE)` is always false here since it's already checked earlier

nimble/transport/socket/src/ble_hci_socket.c
```diff
@@ -281,6 +343,9 @@ ble_hci_sock_cmdevt_tx(uint8_t *hci_ev, uint8_t h4_type)
     memset(&msg, 0, sizeof(msg));
     memset(iov, 0, sizeof(iov));
 
+    msg.msg_name = (void *)btaddr;
+    msg.msg_namelen = 6;
+
```
andrzej-kaczmarek, line 348: why do we need this here? there's separate code for nuttx below. if this fixes something else than nuttx, please create separate PR.

porting/npl/nuttx/src/os_callout.c
```diff
@@ -0,0 +1,159 @@
...
+    return c->c_active;
+}
+
+int ble_npl_callout_inited(struct ble_npl_callout *c)
+{
+    return (c->c_timer != NULL);
+}
+
+ble_npl_error_t ble_npl_callout_reset(struct ble_npl_callout *c,
+				      ble_npl_time_t ticks)
```
andrzej-kaczmarek, line 79: use spaces for indentation, return type should be in separate line. the same for other functions in NPL

porting/npl/nuttx/src/os_callout.c
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    if (ticks == 0) {
+        ticks = 1;
+    }
+
+    c->c_ticks = ble_npl_time_get() + ticks;
+
+    its.it_interval.tv_sec = 0;
+    its.it_interval.tv_nsec = 0;                     // one shot
```
andrzej-kaczmarek, line 94: do not use C++ style comments

andrzej-kaczmarek (changes requested):
overall this looks good, few minor comments
please make sure NPL files comply with NimBLE coding style. other files that are "extras" for NuttX code (like tests) can be left as-is.

nimble/host/src/ble_hs_hci.c
v01d: I found that this check would return true unexpectedly and it would not get to the release. It seemed to be a race condition due to the fact that the this function would get called too fast under certain conditions. I think the sequence was that the lock was actually set after the if check.

nimble/transport/socket/src/ble_hci_socket.c
v01d: Right. I initially tried to use sendmsg() for NuttX but that is not implemented yet for Bluetooth sockets, this was leftover code. I just removed it.

porting/examples/nuttx/README.md
v01d: I'll remove it since the build starts on NuttX side and the documentation should be there.

porting/nimble/include/log/log.h
v01d: Please see previous comment: https://github.com/apache/mynewt-nimble/pull/878#issuecomment-718964467

porting/targets/nuttx/pkg.yml
v01d: Fixed

porting/npl/nuttx/src/os_callout.c
v01d: ok, I fixed coding style, let me know if you find other issues

porting/npl/nuttx/src/os_callout.c
v01d: done

porting/examples/nuttx/include/syscfg.h
v01d: No, I just removed it.

v01d:
I just force pushed the changes. I still don't know what to do about the coding style error reported by CI.

nimble/host/src/ble_hs_hci.c
v01d: Ok, I looked further into the issue and it was not a race condition after all. The problem is that ble_npl_sem_get_count() returns unsigned integer and NuttX returns negative numbers in sem_getvalue() to indicate number of lock count (this is a valid option as per the man page). So this would generate 65536 as the sem value and the if would get triggered.
It now works as expected.

porting/nimble/include/log/log.h
v01d: Any suggestion on what to do about it? I think it is the only remaining issue in the PR

sjanc: Hmm I think for now we could just put all this under #ifndef LOG_DEBUG

in long term I think we should add own NPL_LOG_* and use those instead of relying on modlog module defining those

porting/nimble/include/log_common/log_common.h
```diff
@@ -26,14 +26,116 @@
...
+extern struct log_info g_log_info;
+
+/** @typedef log_append_cb
+ * @brief Callback that is executed each time the corresponding log is appended
+ * to.
+ *
+ * @param log                   The log that was just appended to.
+ * @param idx                   The index of newly appended log entry.
+ */
+typedef void log_append_cb(struct log *log, uint32_t idx);
```
sjanc, line 130: I think we need to #include <stdint.h>  in this file, this should fix CI error for mesh sample

porting/nimble/include/log/log.h
v01d: I think that would require an include to be present before defining LOG_DEBUG which may not always be the case. What about `#if MYNEWT`?

porting/nimble/include/log_common/log_common.h
v01d: Ok, I'll add the include

porting/nimble/include/log/log.h
andrzej-kaczmarek: I think `#if MYNEWT` is ok for now, we'll need to resolve this anyway in some other way as @sjanc proposed

v01d:
It seems last few changes made CI happy. I think those were all requested changes. Do you think it is OK to merge?

sjanc (approved):

andrzej-kaczmarek (approved):

PR #880, opened 2020-10-30
nimble/ll: Use true RNG instead of rand

sjanc:
Don't use stdlib pseudo rand in LL code. This also means LL code
will no longer seed stdlib srand() on init as this should be
application responsibility anyway.

### DIFF
--- a/nimble/controller/src/ble_ll_rand.c
+++ b/nimble/controller/src/ble_ll_rand.c
@@ -120,6 +120,18 @@ ble_ll_rand_data_get(uint8_t *buf, uint8_t len)
     return BLE_ERR_SUCCESS;
 }
 
+/* Simple wrapper to allow easy replacement of rand() */
+int
+ble_ll_rand(void)
+{
+    int data;
+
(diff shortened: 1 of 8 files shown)


nimble/controller/src/ble_ll_rand.c
```diff
@@ -120,6 +120,18 @@ ble_ll_rand_data_get(uint8_t *buf, uint8_t len)
...
+/* Simple wrapper to allow easy replacement of rand() */
+int
+ble_ll_rand(void)
+{
+    int data;
+
+    ble_ll_rand_data_get((uint8_t *)&data, sizeof(data));
+
+    /* rand returns signed integer but in range 0-RAND_MAX */
+    return data & RAND_MAX;
```
andrzej-kaczmarek, line 132: I would just use `uint32_t` as return type and return data directly - this does not need to compatible with `rand()` since our code does not really care, it will limit returned value anyway if needed.

sjanc:
abandoning for now

PR #883, opened 2020-11-13
nimble/ll: Use jrand48 for pseudo random generator

sjanc:
Don't use rand() for for pseudo random generation as this shares global
state. Use jrand48 instead and provide wrapper for late initialization.
This is to avoid seeding on init beacouse trng might not be accessible
yet.

This also means LL code will no longer seed stdlib srand() on init as
this should be application responsibility anyway.

### DIFF
--- a/nimble/controller/src/ble_ll_rand.c
+++ b/nimble/controller/src/ble_ll_rand.c
@@ -17,6 +17,9 @@
  * under the License.
  */
 
+/* for jrand48 */
+#define _XOPEN_SOURCE
+#include <stdlib.h>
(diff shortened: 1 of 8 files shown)


nimble/controller/src/ble_ll_rand.c
```diff
@@ -20,6 +20,7 @@
 #include <stdint.h>
```
benpicco, line 20:
```suggestion
#define _XOPEN_SOURCE
#include <stdint.h>
```

andrzej-kaczmarek (approved):

PR #886, opened 2020-11-16
cmac: Update initialization and calibration routines

andrzej-kaczmarek:
Make sure we initialize RF part only once, that is we only need to read
trim values, TX power configuration and do full calibration once after
boot.

Partial calibration is done on host request in idle time when RF is not
used.


wes3 (approved):
LGTM

PR #894, opened 2020-12-17
targets: Add blehci over USB for pca10056

kasjer:
Target to use nimble controller on Linux with USB.

Target uses some randomly selected USB VID/PID and BT address.

TinyUSB repository must be in project.yml

```
repository.tinyusb:
    type: github
    vers: 0.7.0
    user: hathach
    repo: tinyusb
```

### DIFF
--- a/targets/nordic_pca10056-blehci-usb/syscfg.yml
+++ b/targets/nordic_pca10056-blehci-usb/syscfg.yml
@@ -1,0 +1,6 @@
+syscfg.vals:
+    BLE_HCI_TRANSPORT: usb
+    USBD_BTH: 1
+
+    USBD_PID: 0xC01A
+    USBD_VID: 0xC0CA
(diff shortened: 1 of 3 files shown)


targets/nordic_pca10056-blehci-usb/syscfg.yml
```diff
@@ -0,0 +1,7 @@
+syscfg.vals:
+    BLE_HCI_TRANSPORT: usb
+    BLE_PUBLIC_DEV_ADDR: "(uint8_t[6]){0x81, 0x19, 0x1A, 0xC0, 0xCA, 0xC0}"
```
sjanc, line 3: I'm not sure if we should hardcode public addr in samples, and BlueZ is able to handle controllers without one just fine

rymanluk: @kasjer please just remove BLE_PUBLIC_DEV_ADDR

kasjer: done

sjanc (approved):

PR #895, opened 2020-12-18
ble_gap: fix warning about implicit cast form pointer to int

v01d:
This solves the following warning:

returning 'int (*)(struct ble_gap_event *, void *)' from a function with return type '_Bool8' {aka 'unsigned char'} makes integer from pointer without a cast [-Werror=int-conversion]

which was found when building nimBLE as part of NuttX


andrzej-kaczmarek (approved):

v01d:
Thanks!

PR #898, opened 2021-01-05
NimBLE host: Modify `ble_gap_wl_set` to help remove all addresses from controller whitelist

prasad-alatkar:
In NimBLE host, we only have `ble_gap_wl_set` API. This API can be leveraged to add and remove addresses in whitelist. However, issue comes when `white_list_count = 0`. Instead of cluttering `ble_gap_wl_set` API with generic implementation, this PR adds API in host to remove addresses from whitelist.  

Edit: As per suggestion by @sjanc , I have removed addition of new API. The PR only modifies `ble_gap_wl_set` to clear complete whitelist stored in controller.

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -2169,11 +2169,6 @@ ble_gap_wl_set(const ble_addr_t *addrs, uint8_t white_list_count)
 
     ble_hs_lock();
 
-    if (white_list_count == 0) {
-        rc = BLE_HS_EINVAL;
-        goto done;
-    }
-
     for (i = 0; i < white_list_count; i++) {
         if (addrs[i].type != BLE_ADDR_PUBLIC &&
             addrs[i].type != BLE_ADDR_RANDOM) {
--- a/nimble/host/test/src/ble_gap_test.c
+++ b/nimble/host/test/src/ble_gap_test.c
@@ -449,9 +449,10 @@ TEST_CASE_SELF(ble_gap_test_case_wl_bad_args)
 
     ble_gap_test_util_init();
 
-    /*** 0 white list entries. */
+    /*** 0 white list entries. This is acceptable now with the wl_set API
+     * change. */
     rc = ble_hs_test_util_wl_set(NULL, 0, 0, 0);
-    TEST_ASSERT(rc == BLE_HS_EINVAL);
+    TEST_ASSERT(rc == 0);
 
     /*** Invalid address type. */
     rc = ble_hs_test_util_wl_set(


prasad-alatkar:
@rymanluk @ccollins476ad Could you guys help to get these changes merged ?

sjanc (changes requested):
This patch is missing check if host is allowed to touch WL in first place.

But from what I see in code restriction on device list >0 is not really needed so why not just remove it from ble_gap_wl_set()  and allow 0 length (which would be equivalent to clearing WL)?   Seems like following would work just fine:

```
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -2169,11 +2169,6 @@ ble_gap_wl_set(const ble_addr_t *addrs, uint8_t white_list_count)
 
     ble_hs_lock();
 
-    if (white_list_count == 0) {
-        rc = BLE_HS_EINVAL;
-        goto done;
-    }
-
     for (i = 0; i < white_list_count; i++) {
         if (addrs[i].type != BLE_ADDR_PUBLIC &&
             addrs[i].type != BLE_ADDR_RANDOM) {
```

prasad-alatkar:
> This patch is missing check if host is allowed to touch WL in first place.
> 
> But from what I see in code restriction on device list >0 is not really needed so why not just remove it from ble_gap_wl_set() and allow 0 length (which would be equivalent to clearing WL)? Seems like following would work just fine:
> 
> ```
> --- a/nimble/host/src/ble_gap.c
> +++ b/nimble/host/src/ble_gap.c
> @@ -2169,11 +2169,6 @@ ble_gap_wl_set(const ble_addr_t *addrs, uint8_t white_list_count)
>  
>      ble_hs_lock();
>  
> -    if (white_list_count == 0) {
> -        rc = BLE_HS_EINVAL;
> -        goto done;
> -    }
> -
>      for (i = 0; i < white_list_count; i++) {
>          if (addrs[i].type != BLE_ADDR_PUBLIC &&
>              addrs[i].type != BLE_ADDR_RANDOM) {
> ```

Yes @sjanc , I do agree that your suggested code changes can help get the job done here. However, what would you suggest about having a separate API for whitelist remove ? If addition of API is acceptable then I can add check `MYNEWT_VAL(BLE_WHITELIST)`, if you are not in favor of adding a new API, I can change the PR to remove `white_list_count == 0` check.

sjanc:
I'd prefer if possible we do not add new APIs as long us current ones can be extended

prasad-alatkar:
@rymanluk @sjanc I have incorporated your suggested changes. PTAL.

sjanc (approved):

Issue #24, opened 2018-03-30
Helper functions for subscribing to notifications / indications


ccollins476ad:
Currently, an application subscribes to notifications or indications by performing a GATT write procedure on the appropriate CCCD.  Knowing which two-byte pattern to write, and remembering to use little-endian, feels a bit too low-level to impose on the application developer.

I suggest we add two helper functions:
```
int
ble_gattc_subscribe_notifications(uint16_t conn_handle, uint16_t cccd_handle,
                                  ble_gatt_attr_fn *cb, void *cb_arg);
int
ble_gattc_subscribe_indications(uint16_t conn_handle, uint16_t cccd_handle,
                                ble_gatt_attr_fn *cb, void *cb_arg);
```

PR #900, opened 2021-01-12
ble_gatt: Add subscribing helper functions

KKopyscinski:
Adds 2 helper functions for subscribing to notification/indication.
Using these functions, app developer doesn't need to know what values
to write to CCCD to subscribe and how these values have to be formatted.

Adds functions mentioned in issue https://github.com/apache/mynewt-nimble/issues/24


KKopyscinski:
@rymanluk Could you review this?

nimble/host/src/ble_gattc.c
```diff
@@ -4366,6 +4366,30 @@ ble_gattc_indicate(uint16_t conn_handle, uint16_t chr_val_handle)
     return ble_gattc_indicate_custom(conn_handle, chr_val_handle, NULL);
 }
 
+int
+ble_gattc_subscribe_notifications(uint16_t conn_handle, uint16_t cccd_handle,
```
sjanc, line 4370: hmm I'd prefer if this take char value handle

sjanc (changes requested):
also how about adding also ble_gattc_unsubscribe() ?

KKopyscinski:
As disgussed offline this would be better to do after implementation of GATT caching, as then user could use handle to characteristic not CCCD

PR #906, opened 2021-01-27
NuttX fixes

v01d:
This introduces a fix for NuttX callout mechanism. SIGEV_THREAD is not entirely the same as in Linux and requires an own thread to work correctly (see commit message). This also adds random address generation to NuttX example so that it can work out of the box with chips which do not normally have a public address set.

### DIFF
--- a/porting/examples/nuttx/ble.c
+++ b/porting/examples/nuttx/ble.c
@@ -97,6 +97,17 @@ static void
 app_ble_sync_cb(void)
 {
     int rc;
+    ble_addr_t addr;
+
+    /* generate new non-resolvable private address */
+
+    rc = ble_hs_id_gen_rnd(1, &addr);
+    assert(rc == 0);
+
+    /* set generated address */
+
+    rc = ble_hs_id_set_rnd(addr.val);
+    assert(rc == 0);
 
     rc = ble_hs_util_ensure_addr(0);
     assert(rc == 0);
--- a/porting/npl/nuttx/src/os_callout.c
+++ b/porting/npl/nuttx/src/os_callout.c
@@ -28,17 +28,52 @@
 
 #include "nimble/nimble_npl.h"
 
-static void
-ble_npl_callout_timer_cb(union sigval sv)
+#ifndef CONFIG_NIMBLE_CALLOUT_THREAD_STACKSIZE
+#define CONFIG_NIMBLE_CALLOUT_THREAD_STACKSIZE 1024
+#endif
+
+struct ble_npl_callout *pending_callout = NULL;
+
+bool thread_started = false;
+pthread_t callout_thread;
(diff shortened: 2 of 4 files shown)


v01d:
@andrzej-kaczmarek it seems CI is not picking up the job. Is it down?

sjanc:
this is due to travis moving workers from travis-ci.org to travis-ci.com

sjanc (approved):

v01d:
sorry, I missed a small change, I just force pushed. The change I added is this:
```
diff --git a/porting/npl/nuttx/src/os_callout.c b/porting/npl/nuttx/src/os_callout.c
index fe12a8e3..e4580da9 100644
--- a/porting/npl/nuttx/src/os_callout.c
+++ b/porting/npl/nuttx/src/os_callout.c
@@ -28,6 +28,10 @@
 
+#ifndef CONFIG_NIMBLE_CALLOUT_THREAD_STACKSIZE
+#define CONFIG_NIMBLE_CALLOUT_THREAD_STACKSIZE 1024
+#endif
+
 struct ble_npl_callout *pending_callout = NULL;
 
 bool thread_started = false;
@@ -81,7 +85,10 @@ ble_npl_callout_init(struct ble_npl_callout *c,
     struct sigevent         event;
 
     if (!thread_started) {
-        pthread_create(&callout_thread, NULL, callout_handler, NULL);
+        pthread_attr_t attr;
+        pthread_attr_init(&attr);
+        pthread_attr_setstacksize(&attr, CONFIG_NIMBLE_CALLOUT_THREAD_STACKSIZE);
+        pthread_create(&callout_thread, &attr, callout_handler, NULL);
         thread_started = true;
     }
 
diff --git a/porting/npl/nuttx/src/os_task.c b/porting/npl/nuttx/src/os_task.c
index 2e020236..ab535a36 100644
--- a/porting/npl/nuttx/src/os_task.c
+++ b/porting/npl/nuttx/src/os_task.c
@@ -60,8 +60,10 @@ ble_npl_task_init(struct ble_npl_task *t, const char *name, ble_npl_task_func_t
     if (err) return err;
     err = pthread_attr_getschedparam (&t->attr, &t->param);
     if (err) return err;
+#if CONFIG_RR_INTERVAL > 0
     err = pthread_attr_setschedpolicy(&t->attr, SCHED_RR);
     if (err) return err;
+#endif
     t->param.sched_priority = prio;
     err = pthread_attr_setschedparam (&t->attr, &t->param);
     if (err) return err;
@@ -69,6 +71,11 @@ ble_npl_task_init(struct ble_npl_task *t, const char *name, ble_npl_task_func_t
     t->name = name;
     err = pthread_create(&t->handle, &t->attr, func, arg);
 
+    if (err == ENOMEM)
+      {
+        err = OS_ENOMEM;
+      }
+
     return err;
 }
```

PR #907, opened 2021-01-28
nimble/phy: Improve tx_late handling

andrzej-kaczmarek:
There is a very small window between setting timer comparator and
configuring field_1 where frame_start event can happen due to match
on that timer. To prevent this from generating a fatal error, let's
disable corresponding error before setting comparator and check if
frame_start event already happened after frame_1 is configured. If
it did happen, assume we did not configure frame_1 on time and do
tx_late.


wes3 (approved):
LGTM

PR #908, opened 2021-02-01
nimble/test: Fix LE SC tests

sjanc:
Some LE SC tests depend on known initial values on crypto side.
With latest update to TinyCrypt provided prng function is used also
for initial seeding and thus tests were failing. To keep things simple
just don't initialize random source function when doign unit tests.


sjanc:
FYI @vikrant-proxy

vrahane (approved):
Makes sense. LGTM.

PR #910, opened 2021-02-05
nimble/phy: Fix tx_late handling

andrzej-kaczmarek:
Disabling some errors temporarily to avoid tx_late does not work
properly because those errors are not reset when BS is reset and we
may still have CMAC error after those errors are enabled.

To make things work let's just keep all relevant errors disabled and
ignore them since we can use exceptions status to detect tx_late pretty
consistently.


wes3 (approved):
LGTM

PR #918, opened 2021-02-18
nimble/ll: Add restriction when enabling periodic advertising

sjanc:
Periodic advertising requires extended advertising to be enabled.


KKopyscinski:
@sjanc I checked it again and CDB is off by default; this fixes issue when CDB is on. Could you change commit title?

andrzej-kaczmarek (approved):

PR #924, opened 2021-02-24
btshell/cmd.c: make strcpy in parse_dev_addr memory safe 

KKopyscinski:


apps/btshell/src/cmd.c
```diff
@@ -99,7 +99,10 @@ parse_dev_addr(const char *prefix, const struct kv_pair *addr_types,
     char name[32];
     int rc;
 
-    /* XXX string operations below are not quite safe, but do we care? */
+    /* Make sure we're not copying/concatenating more than sizeof name  */
+    if (sizeof(prefix)+sizeof("addr")+sizeof("_type") > sizeof(name)) {
```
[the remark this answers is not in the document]
KKopyscinski: fixed, let's just copy up to n valid characters, we know how many that is for the prefix (5). If prefix is wrong it'll return error later

apps/btshell/src/cmd.c
```diff
@@ -99,12 +99,11 @@ parse_dev_addr(const char *prefix, const struct kv_pair *addr_types,
...
     int rc;
 
-    /* XXX string operations below are not quite safe, but do we care? */
-
+    /* Valid prefixes are peer_, own_, or null. Let's copy up to 5 characters */
     if (!prefix) {
         name[0] = '\0';
     } else {
-        strcpy(name, prefix);
+        strncpy(name, prefix, 5);
```
andrzej-kaczmarek, line 106: strncpy will not null terminate name if prefix is longer than 5 characters

KKopyscinski: so maybe just return error when prefix is longer than 5 and avoid copying at all? longest viable prefix is "peer_"

KKopyscinski: ENOENT for example

andrzej-kaczmarek: I do not know if we won't use other prefixes in future. if you want to make it really fool-proof then better use strlcat (I think we support it) everywhere and handle errors.

KKopyscinski: ok, switched to strlcat/cpy

apps/btshell/src/cmd.c
```diff
@@ -104,17 +104,23 @@ parse_dev_addr(const char *prefix, const struct kv_pair *addr_types,
     if (!prefix) {
         name[0] = '\0';
     } else {
-        strcpy(name, prefix);
+        if (strlcpy(name, prefix, sizeof(name)) > sizeof(name)) {
```
sjanc, line 107: this (and others) should be  >=


also, XXX in 102 is no longer needed :)

KKopyscinski: done :)

apps/btshell/src/cmd.c
```diff
@@ -99,22 +99,26 @@ parse_dev_addr(const char *prefix, const struct kv_pair *addr_types,
...
         name[0] = '\0';
     } else {
-        strcpy(name, prefix);
+        if (strlcpy(name, prefix, sizeof(name)) >= sizeof(name)) {
+            return EINVAL;
+        }
     }
 
-    strcat(name, "addr");
+    if (strlcat(name, "addr", sizeof(name)) > sizeof(name)) {
```
sjanc, line 110: >=

KKopyscinski: fixed

sjanc (approved):

PR #936, opened 2021-03-15
nimble/ll: Fix possible race on TX queue

andrzej-kaczmarek:
Removing from ll_tx_pkt_q should be done in critical section since
insertion can be done from isr.

There's no need to protect peek on 1st element since we only insert
in isr and remove in LL so there's no risk of element being removed
after we checked not-null.


benmccrea (approved):

wes3 (approved):
LGTM

PR #939, opened 2021-03-17
sm: fail pairing in SC only mode if keysize is less than 128 bit

KKopyscinski:
this is required to pass GAP/SEC/SEM/BI tests


KKopyscinski:
This is not fully compatible with both Bluetooth and test specs. I close it for now, as solving the issue requires different approach

Issue #940, opened 2021-03-18
blehci-usb app fails on assert when trying to upload an image using newtmgr


ncasaril:
Newtmgr stalls on image upload. 
Not entirely certain that this is an issue in nimble, as it may very well be within newtmgr also. However, newtmgr successfully uploads images using the same hw but with a different hci dongle / built in ble dev. 

To reproduce: 
- setup a pca10056 board / nrf52840-dongle with the blehci app using usb as transport. Should become a hci device when connected in linux. 
- using newtmgr under linux upload an image over ble using the created hci device

The below assert gets hit after a few seconds:

https://github.com/apache/mynewt-nimble/blob/32621d01b444d99dedba8fe85c44a2bb1b2460c4/nimble/transport/usb/src/ble_hci_usb.c#L207

It may very well be that things go wrong before this line as newtmgr has been stuck at 0% for a few seconds already. 

Increasing the BLE_ACL_BUF_SIZE only delays the problem, it doesn't solve it. 

Some output from btmon. The newtmgr upload appears stalled after the ACL TX at 11.608489: 

```
...
< ACL Data TX: Handle 1 flags 0x00 dlen 9                                                                                                           #173 [hci0] 11.591733
      ATT: Write Request (0x12) len 4
        Handle: 0x001a
          Data: 0100
> HCI Event: Number of Completed Packets (0x13) plen 5                                                                                              #174 [hci0] 11.602625
        Num handles: 1
        Handle: 1
        Count: 1
> ACL Data RX: Handle 1 flags 0x02 dlen 5                                                                                                           #175 [hci0] 11.606626
      ATT: Write Response (0x13) len 0
< ACL Data TX: Handle 1 flags 0x00 dlen 255                                                                                                         #176 [hci0] 11.608483
< ACL Data TX: Handle 1 flags 0x01 dlen 4                                                                                                           #177 [hci0] 11.608489
      ATT: Write Command (0x52) len 254
        Handle: 0x0019
[...]

PR #944, opened 2021-03-22
nimble/ll_conn: Fix for frag tx of mbufs >255 bytes

ncasaril:
Further details in #940. 

When using nimble over hci with a perph that is using an MTU larger than 255 bytes the variables changed in this PR overflowed causing the transfer to halt. 

With these changes an MTU of up to 512 can be used, but not 513 and above. The MTU should be able to go to 525?! 

I'm also aware that I'm not very versed in the nimble code base and am open to other solutions to this altogether?

### DIFF
--- a/nimble/controller/src/ble_ll_conn.c
+++ b/nimble/controller/src/ble_ll_conn.c
@@ -959,8 +959,8 @@ ble_ll_conn_tx_pdu(struct ble_ll_conn_sm *connsm)
     uint8_t hdr_byte;
     uint8_t end_transition;
     uint8_t cur_txlen;
-    uint8_t next_txlen;
-    uint8_t cur_offset;
+    uint16_t next_txlen;
+    uint16_t cur_offset;
     uint16_t pktlen;
     uint32_t next_event_time;
     uint32_t ticks;
@@ -1129,11 +1129,10 @@ ble_ll_conn_tx_pdu(struct ble_ll_conn_sm *connsm)
         if ((cur_offset + cur_txlen) < pktlen) {
             next_txlen = pktlen - (cur_offset + cur_txlen);
         } else {
-            if (nextpkthdr->omp_len > connsm->eff_max_tx_octets) {
-                next_txlen = connsm->eff_max_tx_octets;
-            } else {
-                next_txlen = nextpkthdr->omp_len;
-            }
+            next_txlen = connsm->eff_max_tx_octets;
+        }
+        if (next_txlen > connsm->eff_max_tx_octets) {
+            next_txlen = connsm->eff_max_tx_octets;
         }
 
         /*
@@ -1491,7 +1490,7 @@ ble_ll_conn_can_send_next_pdu(struct ble_ll_conn_sm *connsm, uint32_t begtime,
                               uint32_t add_usecs)
 {
     int rc;
-    uint8_t rem_bytes;
+    uint16_t rem_bytes;
     uint32_t ticks;
     uint32_t usecs;
     uint32_t next_sched_time;
@@ -3791,7 +3790,7 @@ ble_ll_conn_rx_isr_end(uint8_t *rxbuf, struct ble_mbuf_hdr *rxhdr)
     uint8_t conn_sn;
     uint8_t conn_nesn;
     uint8_t reply;
-    uint8_t rem_bytes;
+    uint16_t rem_bytes;
     uint8_t opcode = 0;
     uint8_t rx_pyld_len;
     uint32_t begtime;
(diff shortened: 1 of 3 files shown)


nimble/controller/src/ble_ll_conn.c
```diff
@@ -1128,6 +1128,7 @@ ble_ll_conn_tx_pdu(struct ble_ll_conn_sm *connsm)
          */
         if ((cur_offset + cur_txlen) < pktlen) {
             next_txlen = pktlen - (cur_offset + cur_txlen);
+            next_txlen = (next_txlen < connsm->eff_max_tx_octets) ? next_txlen : connsm->eff_max_tx_octets;
```
andrzej-kaczmarek, line 1131: I'd rewrite whole block:
```
if ((cur_offset + cur_txlen) < pktlen) {
    next_txlen = pktlen - (cur_offset + cur_txlen);
} else {
    next_txlen = connsm->eff_max_tx_octets;
}
if (next_txlen > connsm->eff_max_tx_octets) {
    next_txlen = connsm->eff_max_tx_octets;
}
```

nimble/include/nimble/ble.h
```diff
@@ -109,8 +109,8 @@ struct ble_mbuf_hdr_rxinfo
 struct ble_mbuf_hdr_txinfo
 {
     uint8_t flags;
-    uint8_t offset;
-    uint8_t pyld_len;
+    uint16_t offset;
+    uint16_t pyld_len;
     uint8_t hdr_byte;
```
andrzej-kaczmarek, line 114: move `hdr_byte` after `flags` to optimize alignment in struct

andrzej-kaczmarek (changes requested):

nimble/controller/src/ble_ll_conn.c
ncasaril: Agree, much cleaner.

nimble/include/nimble/ble.h
ncasaril: Done

ncasaril:
@wes3 @kasjer, any ideas why this only works up to a MTU of 512 and not higher? 
I'm running out of ideas as to how to push this further.

kasjer:
> @wes3 @kasjer, any ideas why this only works up to a MTU of 512 and not higher?
> I'm running out of ideas as to how to push this further.

I'm checking this from the USB side perspective...

andrzej-kaczmarek (approved):
LGTM. Please update PR title if it's ok to merge. I'll run qualification testcases before merging just to be sure I did not miss anything here.

ncasaril:
Sounds good. We can explore separately why the MTU isn't happy exceeding 512 separately.

nimble/controller/src/ble_ll_conn.c
```diff
@@ -1129,11 +1129,10 @@ ble_ll_conn_tx_pdu(struct ble_ll_conn_sm *connsm)
         if ((cur_offset + cur_txlen) < pktlen) {
             next_txlen = pktlen - (cur_offset + cur_txlen);
         } else {
-            if (nextpkthdr->omp_len > connsm->eff_max_tx_octets) {
-                next_txlen = connsm->eff_max_tx_octets;
-            } else {
-                next_txlen = nextpkthdr->omp_len;
-            }
+            next_txlen = connsm->eff_max_tx_octets;
```
ncasaril, line 1132: This should instead probably be
 
```
     next_txlen = nextpkthdr->omp_len;
```

nimble/include/nimble/ble.h
```diff
@@ -109,9 +109,9 @@ struct ble_mbuf_hdr_rxinfo
 struct ble_mbuf_hdr_txinfo
 {
     uint8_t flags;
-    uint8_t offset;
-    uint8_t pyld_len;
     uint8_t hdr_byte;
+    uint16_t offset;
+    uint16_t pyld_len;
```
ncasaril, line 114: Changing the types to uint16_t here breaks larger replies from a peripheral. For instance listing the images using newtmgr with two images present on the peripheral. Changing the types back to uint8_t makes it work. Adding a dummy byte before the offset / pyld_len breaks download again. I.e.:

```
/* This works */
struct ble_mbuf_hdr_txinfo
 {
     uint8_t flags;
     uint8_t offset;
     uint8_t pyld_len;
     uint8_t hdr_byte;
}; 
```
```
/* This doesn't work */
struct ble_mbuf_hdr_txinfo
{
    uint8_t flags;
    uint8_t offset;
    uint8_t dummy;
    uint8_t pyld_len;
    uint8_t hdr_byte;
}; 
```

Is anyone else seeing this? @andrzej-kaczmarek

PR #945, opened 2021-03-23
l2cap: Fix double disconnect issue

rymanluk:
It might happen that we get into situation that host will try to
disconnect same channel couple times e.g when receiveing log of
corrupted packages.

This patch make sure that host will send l2cap disconnect command only
once.

### DIFF
--- a/nimble/host/src/ble_l2cap_priv.h
+++ b/nimble/host/src/ble_l2cap_priv.h
@@ -62,6 +62,7 @@ typedef int ble_l2cap_rx_fn(struct ble_l2cap_chan *chan);
 
 struct ble_l2cap_chan {
     SLIST_ENTRY(ble_l2cap_chan) next;
+    bool disconnecting;
     uint16_t conn_handle;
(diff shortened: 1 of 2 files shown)


nimble/host/src/ble_l2cap_priv.h
```diff
@@ -62,6 +62,7 @@ typedef int ble_l2cap_rx_fn(struct ble_l2cap_chan *chan);
 
 struct ble_l2cap_chan {
     SLIST_ENTRY(ble_l2cap_chan) next;
+    bool disconnecting;
```
andrzej-kaczmarek, line 65: what about using `flags` instead?

rymanluk: yeah, it was too quick fix and I forgot about our flags. Changing it. Thanks

andrzej-kaczmarek (approved):

sjanc (approved):

KKopyscinski (approved):

PR #948, opened 2021-03-24
transport/usb: Improve host de-synchronization recovery

kasjer:
In case when host did not waited for all events for past
commands and send another command system would assert.
This happens because for every command there is one event generated,
and because this is one to one relation buffer for command and
event is shared.
When event that was generated by controller was not sent yet,
and host stopped execution pending event would block any further
host attempts to start over.

Now when command is received when event was no confirmed by USB yet,
but new command is HCI reset. Code simply patches event queue
so it no longer keeps track of event that occupies same buffer.
When USB finally confirms that data was sent data length is not
compared with already scratched buffer.

If any other command arrives when complete event is not confirmed
by USB new command is silently dropped.


sjanc:
is this stil valid?

PR #953, opened 2021-03-26
nimble/hs: Fix mbuf allocation for combined build

andrzej-kaczmarek:
In combined build we need to allocate mbufs with user header since LL
needs extra space to keep txinfo.

This works so far only because txinfo has 4 bytes and it overwrites
omp_next in packet header which is only used when mbuf is enqueued.
Once dequeued, we don't use so it's "ok" to overwrite it. However if
txinfo is extended with some extra data we will start overwriting
PDU data when writing to txinfo.


andrzej-kaczmarek:
this is required for combined build to work once https://github.com/apache/mynewt-nimble/pull/944 is merged

sjanc (approved):

rymanluk (approved):

PR #969, opened 2021-04-13
transport/cmac: Fix race on LL event write

andrzej-kaczmarek:
HCI commands are processed in LL task context and thus CS/CC events are
written to mbox in the same context. This can lead to following race:
- CS/CC is written to mbox
- interrupt in handled on CMAC *before* command buffer is freed
- M33 reads CS/CC and writes next command to mbox
- CMAC reads new command

This triggers an assert in code because we do not have free buffer for
new command because we were busy handling interrupts and did not free
it yet.

To fix this we should write and free in critical section.


wes3 (approved):
LGTM

PR #971, opened 2021-04-15
nimble/host: Fix potential MITM vulnerability in case of secure connection pairing

prasad-alatkar:
Issue: In case of secure connection pairing and Paskkey Entry method, there is chance that MITM attacker can reply with same public key as received from vulnerable BLE device. This can potentially authenticate MITM attacker. This PR fixes this potential vulnerability.
 
Solution: Check if the peer public key is same as our generated public key. Return fail if the public keys match to avoid 'Impersonation in Passkey entry protocol' vulnerability.


prasad-alatkar:
@sjanc @rymanluk Could you guys take a look ?

sjanc:
do I get it right that without this patch pairing would end with success even though remote peer is not knowing our private key?
How would it calculate valid DHKey and then mackey and LTK?

nimble/host/src/ble_sm_sc.c
```diff
@@ -612,6 +612,14 @@ ble_sm_sc_public_key_rx(uint16_t conn_handle, struct os_mbuf **om,
     }
 
     cmd = (struct ble_sm_public_key *)(*om)->om_data;
+    /* Check if the peer public key is same as our generated public key.
+     * Return fail if the public keys match. */
+    if (ble_sm_sc_keys_generated) {
+        if (memcmp(cmd, ble_sm_sc_pub_key, 64) == 0) {
+            res->enc_cb = 1;
+            res->sm_err = BLE_SM_ERR_AUTHREQ;
```
rymanluk, line 620: shouldn't be here a return?

Could you please create a unit test for that in nimble/host/test/src/ble_sm_sc_test.c

rymanluk (changes requested):

nimble/host/src/ble_sm_sc.c
```diff
@@ -612,6 +612,14 @@ ble_sm_sc_public_key_rx(uint16_t conn_handle, struct os_mbuf **om,
     }
 
     cmd = (struct ble_sm_public_key *)(*om)->om_data;
+    /* Check if the peer public key is same as our generated public key.
+     * Return fail if the public keys match. */
+    if (ble_sm_sc_keys_generated) {
```
sjanc, line 617: I think this check is not needed as ble_sm_sc_ensure_keys_generated() is called before

prasad-alatkar: Yes, the check is not needed.

nimble/host/src/ble_sm_sc.c
prasad-alatkar: Added return. I will create unit test for the same.

sjanc (approved):

rymanluk (commented):
Looks good. Any chance for the unit test for that ?:)

prasad-alatkar:
> Looks good. Any chance for the unit test for that ?:)

@rymanluk yes, I will add the unit test by this weekend :)

prasad-alatkar:
> > Looks good. Any chance for the unit test for that ?:)
> 
> @rymanluk yes, I will add the unit test by this weekend :)


@rymanluk, I have tried to add unit test to verify the scenario. PTAL and suggest if any improvements can be done...

dhrishi:
Hi @rymanluk Can you please review the unit test and merge if it looks okay? Post that, we will pull this into Espressif's NimBLE repository.

rymanluk (approved):
Thanks! Looks good to me.

PR #973, opened 2021-04-19
host/mesh: fix paths in persistent storage

KKopyscinski:
All paths should be using bt_mesh instead of bt/mesh


michal-narajowski (approved):
LGTM

sjanc (approved):

PR #974, opened 2021-04-19
host/gap: fix doxygen for ble_gap_ext_disc()

haukepetersen:
Documentation for the parameters `filter_duplicates` and `filter_policy` for ble_gap_ext_disc() was missing. This made me mix up parameters a couple of times now, so I decided to fix this :-)

Not 100% sure if I understood the usage of those parameters correctly and got the parameter description just right - so any enhancements are welcome!


sjanc (approved):

haukepetersen:
and go.

PR #977, opened 2021-05-12
ble_att_svr: fix keysize check for SC Only mode

KKopyscinski:
Keysize for level 4 security is 16 Bytes not 128 bits.


sjanc:
16 bytes are 128bits :-)  I'd adjust commit message to something like "keysize is kept in bytes and not bits" or similar :-)

KKopyscinski:
Indeed, 16 bytes are 128 bits. Edited commit message :)

PR #978, opened 2021-05-19
mesh: Check if ack is pending in trans_seg

KKopyscinski:
Before scheduling ack it was checked if previous ack reached it's
timeout. If it did, it was assumed it was sent, where in fact,
timeout could be not reached and work could be canceled. Now,
check if work is pending at all, and if not, schedule next one.

This affects MESH/NODE/TNPT/BV-11-C


KKopyscinski:
This is no longer needed (https://github.com/apache/mynewt-nimble/pull/1062)

PR #980, opened 2021-05-20
bttester: add another characteristic requiring authentication

KKopyscinski:
This characteristic is required by GATT/SR/GAR/BI-21-C
(it requires two characteristics requiring authentication for
reading)


KKopyscinski:
As of PTS 8.0.2 this is not required

PR #986, opened 2021-06-08
Blestress fixes

KKopyscinski:


KKopyscinski:
@mkasenberg please review these changes

KKopyscinski:
There is still crash in test 14 - I'll update a PR with fix

KKopyscinski:
All crashes are resolved - the only problem left is silent restart on RX side during test 9 if test 7 was executed before - for now, skip test 7. Reason of this restart is yet unknown - there is no crash visible in debugger, no crash information is printed to console, HCI trace only shows lack of activity - basically, restart causes RTT to deattach.  All tests are working if run independently.

apps/blestress/src/tx_stress.c
```diff
@@ -848,6 +848,16 @@ tx_stress_9_gap_event(struct ble_gap_event *event, void *arg)
...
+    struct ble_gap_conn_params conn_params = {
+        .scan_itvl = 0x0010,
+        .scan_window = 0x0010,
+        .itvl_min = BLE_GAP_INITIAL_CONN_ITVL_MIN,
+        .itvl_max = BLE_GAP_INITIAL_CONN_ITVL_MAX,
+        .latency = 0,
+        .supervision_timeout = 0x0C80,
+        .min_ce_len = 0x0010,
+        .max_ce_len = 0x0300,
+    };
```
sjanc, line 860: are those connections maintained in test for at least this timeout after last one is created?

sjanc:
please rebase

apps/blestress/syscfg.yml
```diff
@@ -75,3 +75,6 @@ syscfg.vals:
 
     # Whether to save data to sys/config, or just keep it in RAM.
     BLE_STORE_CONFIG_PERSIST: 0
+
+    # ACL buf size must be able to contain CoC MPS plus ACL header
+    BLE_ACL_BUF_SIZE: MYNEWT_VAL_BLE_L2CAP_COC_MPS+4
```
rymanluk, line 80: shouldn't it be (MYNEWT_VAL(MYNEWT_VAL_BLE_L2CAP_COC_MPS) + 4 ) ?

apps/blestress/src/stress.h
```diff
@@ -39,7 +39,7 @@ extern "C" {
 #define TX_PHY_MASK 0
 #define RX_PHY_MASK 0
 /* L2CAP SDU */
-#define STRESS_COC_MTU (64000)
```
rymanluk, line 42: why we need to decrease it ?

apps/blestress/src/tx_stress.c
rymanluk: @KKopyscinski ?

apps/blestress/src/rx_stress.c
```diff
@@ -1444,14 +1444,11 @@ rx_stress_main_task_fn(void *arg)
     os_sem_pend(&rx_stress_main_sem, OS_TIMEOUT_NEVER);
 
     /* Standard tests perform */
-    for (i = 11; i < STRESS_UUIDS_NUM; ++i) {
-        if (i == 7 || i == 8 || i == 13) {
-            /* 7,8: PHY update tests cause that the device during the next test
-             * will stuck somewhere and will reset. Skip them for now.
-             * 13: Should work after fixing ble_gattc_notify_custom (nimble issue on GitHub)*/
+    for (i = 2; i < STRESS_UUIDS_NUM; ++i) {
```
rymanluk, line 1447: is this patch still needeD?

apps/blestress/syscfg.yml
KKopyscinski: This incorrectly expands to
``` c
#define MYNEWT_VAL_BLE_TRANSPORT_ACL_SIZE (MYNEWT_VAL_MSYS_1_BLOCK_SIZE-8)
```
and ignores +4. So we need to use `MYNEWT_VAL_` variant. I think it's the same as for `BLE_L2CAP_COC_MPS` definition in `nimble/host/syscfg.yml`

apps/blestress/src/stress.h
KKopyscinski: I'm not sure what was wrong there, reading commit message I mentioned some issue with `ble_l2cap_coc_rx_fn`. Whatever this was, was fixed and this commit is not needed - I retested it. I suspect that https://github.com/apache/mynewt-nimble/commit/c1a2c9949805148520afb852d86afe588061d64e was the fix

apps/blestress/src/tx_stress.c
KKopyscinski: There are total of 15 successful concurrent connections. For example run, last succesfull connection occurs ~164343728us timestamp and first disconnect at 196078120us. This gives us 31.734392s. 0x0C80 is 3200, x10ms = 32s, which considering print delays seems pretty close IMO.

apps/blestress/src/rx_stress.c
KKopyscinski: Yes, but modified - looks like all tests are working now and neither should be skipped.

rymanluk (approved):
LGTM

PR #987, opened 2021-06-08
l2cap: implement echo request-response procedure (l2cap ping)

haukepetersen:
For doing run-time link quality assessment I am in the need of the echo request-response procedure provided by the L2CAP signalling layer. As of now, this functionality is not exposed through any NimBLE APIs. This PR exposes this functionality by adding the `ble_l2cap_ping()` function to NimBLE's l2cap API.

The usage is straight forward, once a BLE connection is established, one has simply to call the `ble_l2cap_ping()`. By passing an optional callback function to that function, this callback is executed one the `ECHO_RSP` packet is received. Additionally, an optional user payload can be appended to the `ECHO_REQ` packet.

Regarding testing of this feature: so far I only integrated and tested it with RIOT by exposing it as a `ping` subcommand for the `nimble_netif` shell command (-> https://github.com/haukepetersen/RIOT/tree/add_nimble_scnetifping, https://github.com/RIOT-OS/RIOT/pull/16539):
```
2021-06-08 12:14:35,823 # ble ping 0
> 2021-06-08 12:14:35,920 # ECHO_RSP, rtt:95 size:0
2021-06-08 12:14:35,936 # ble ping 0
> 2021-06-08 12:14:36,070 # ECHO_RSP, rtt:132 size:0
ble ping 0 moinsen
2021-06-08 12:14:44,246 # ble ping 0 moinsen
> 2021-06-08 12:14:44,321 # ECHO_RSP, rtt:73 size:7
```

Should we also add this to some of the NimBLE examples/test applications to ensure the feature is properly picked up by the NimBLE testing?

### DIFF
--- a/nimble/host/src/ble_l2cap_sig.c
+++ b/nimble/host/src/ble_l2cap_sig.c
@@ -60,6 +60,7 @@
 #define BLE_L2CAP_SIG_PROC_OP_RECONFIG          2
 #define BLE_L2CAP_SIG_PROC_OP_DISCONNECT        3
 #define BLE_L2CAP_SIG_PROC_OP_MAX               4
+#define BLE_L2CAP_SIG_PROC_OP_PING              5
 
 #if MYNEWT_VAL(BLE_L2CAP_ENHANCED_COC)
 #define BLE_L2CAP_ECOC_MIN_MTU  (64)
@@ -97,6 +98,10 @@ struct ble_l2cap_sig_proc {
             uint16_t new_mtu;
         } reconfig;
 #endif
+        struct {
+            ble_l2cap_ping_fn *cb;
+            ble_npl_time_t time_sent;
+        } ping;
     };
 };
 
@@ -112,6 +117,10 @@ static ble_l2cap_sig_rx_fn ble_l2cap_sig_rx_noop;
 static ble_l2cap_sig_rx_fn ble_l2cap_sig_update_req_rx;
 static ble_l2cap_sig_rx_fn ble_l2cap_sig_update_rsp_rx;
 static ble_l2cap_sig_rx_fn ble_l2cap_sig_rx_reject;
+#if MYNEWT_VAL(BLE_L2CAP_COC_MAX_NUM) != 0
+static ble_l2cap_sig_rx_fn ble_l2cap_sig_echo_req;
+static ble_l2cap_sig_rx_fn ble_l2cap_sig_echo_rsp;
+#endif
 
 #if MYNEWT_VAL(BLE_L2CAP_COC_MAX_NUM) != 0
 static ble_l2cap_sig_rx_fn ble_l2cap_sig_coc_req_rx;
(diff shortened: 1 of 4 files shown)


haukepetersen:
rebased and fixed a bug where the signalling command's mbuf was not released in case the command failed.

nimble/host/src/ble_l2cap_sig.c
```diff
@@ -1979,6 +1979,9 @@ ble_l2cap_sig_ping(uint16_t conn_handle, ble_l2cap_ping_fn cb,
 
 
     rc = ble_l2cap_sig_tx(proc->conn_handle, txom);
+    if (rc != 0) {
```
sjanc, line 1982: hmm I think ble_l2cap_sig_tx is (or should) consume txom regardless of result

haukepetersen: Nope, unfortunately it does not. I have been debugging some of my nodes stopping to scan properly, until I found out that the node's msys buffer simply run out of mbufs, caused by this function not freeing `txom` on error...

haukepetersen: For further verification, here the code of `ble_l2cap_sig_tx()`:
```c
int
ble_l2cap_sig_tx(uint16_t conn_handle, struct os_mbuf *txom)
{
    struct ble_l2cap_chan *chan;
    struct ble_hs_conn *conn;
    int rc;

    ble_hs_lock();
    rc = ble_hs_misc_conn_chan_find_reqd(conn_handle, BLE_L2CAP_CID_SIG,
                                         &conn, &chan);
    if (rc == 0) {
        rc = ble_l2cap_tx(conn, chan, txom);
    }
    ble_hs_unlock();

    return rc;
}
```
As one can see, no freeing of `txom` in case `ble_hs_misc_conn_chan_find_reqd()` fails...

sjanc: hmm but then, ble_l2cap_tx will consume txom even on error
maybe ble_l2cap_sig_tx() should be updated to always consume txom,  btw I think ble_sm_tx() has same issue

I'll try to lo review those code parts this week and propose some unified solution

haukepetersen: I completely agree, this makes sense to be fixed coherently. Debugging this kind of inconsistency is no fun :-)

Let me know where I can help!

haukepetersen:
- removed the quick fix for releasing the used mbuf on error in favor of #999
- added precompiler guards around the ping function mapping to include them only if `BLE_L2CAP_COC_MAX_NUM != 0` to prevent compile errors if L2CAP COCs are not used

haukepetersen:
Any chance someone finds the time to look at this feature? It would be great to have this in master :-)

andrzej-kaczmarek (approved):

sjanc (approved):
just squash fixup commits :)

haukepetersen:
wow, that was quick, thanks! :-) Squashed as requested.

haukepetersen:
all green -> go

PR #998, opened 2021-07-08
nimble/porting: Make os_mbuf/os_mpool work on 64-bit.

jimmo:
We're using NimBLE in [MicroPython](https://github.com/micropython/micropython) and do much of our testing on (64-bit) Linux. @dpgeorge

Similar to #96 (@turon). Please see commit messages for details.

@KKopyscinski / @sjanc should I try and get this fixed in mynewt-core instead?

### DIFF
--- a/porting/nimble/src/os_mempool.c
+++ b/porting/nimble/src/os_mempool.c
@@ -139,7 +139,7 @@ os_mempool_init_internal(struct os_mempool *mp, uint16_t blocks,
         /* Blocks need to be sized properly and memory buffer should be
          * aligned
          */
-        if (((uint32_t)(uintptr_t)membuf & (OS_ALIGNMENT - 1)) != 0) {
+        if (((uintptr_t)membuf & (OS_ALIGNMENT - 1)) != 0) {
             return OS_MEM_NOT_ALIGNED;
         }
     }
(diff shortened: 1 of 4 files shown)


porting/nimble/src/os_mempool.c
```diff
@@ -38,7 +38,8 @@
 #define OS_MEMPOOL_TRUE_BLOCK_SIZE(mp) OS_MEM_TRUE_BLOCK_SIZE(mp->mp_block_size)
 #endif
 
-STAILQ_HEAD(, os_mempool) g_os_mempool_list;
+STAILQ_HEAD(, os_mempool) g_os_mempool_list =
+    STAILQ_HEAD_INITIALIZER(g_os_mempool_list);
```
sjanc, line 42: Could you explain why is this needed?  (AFAIR initially it was like that to allow multiple initialization while executing tests)

sjanc (commented):
Yes, lets fix it in core and then resync here.
In general I think we should simply start using uintptr where needed (although mynewt was/is 32bit os only).

jimmo:
> Yes, lets fix it in core and then resync here.
> In general I think we should simply start using uintptr where needed (although mynewt was/is 32bit os only).

Thanks -- I'm not too familiar with mynewt (outside of NimBLE). Do you want me to follow this up or is this best left up to the maintainers?

> Could you explain why is this needed? (AFAIR initially it was like that to allow multiple initialization while executing tests)

I'm not too fussed either way, but I noticed that nothing in NimBLE calls os_mempool_module_init and there's no declaration for it in any header files. So this change seems like the safest way to prevent implementors missing it.

PR #1001, opened 2021-07-19
host/l2cap: send command reject in L2CAP disc req for invalid CID

KKopyscinski:
If ble_hs_conn_chan_find_by_scid fails to find channel, it means
that destination CID in L2CAP Disconnection Request is invalid.
Send L2CAP_COMMAND_REJECT_RSP with BLE_L2CAP_SIG_ERR_INVALID_CID
reason.

This is affecting L2CAP/LE/CFC/BV-23-C

### DIFF
--- a/nimble/host/src/ble_l2cap_sig.c
+++ b/nimble/host/src/ble_l2cap_sig.c
@@ -1540,7 +1540,13 @@ ble_l2cap_sig_disc_req_rx(uint16_t conn_handle, struct ble_l2cap_sig_hdr *hdr,
      * is from peer perspective. It is source CID from nimble perspective
      */
     chan = ble_hs_conn_chan_find_by_scid(conn, le16toh(req->dcid));
-    if (!chan || (le16toh(req->scid) != chan->dcid)) {
+    if (!chan) {
+        os_mbuf_free_chain(txom);
+        ble_hs_unlock();
+        ble_l2cap_sig_reject_invalid_cid_tx(conn_handle, hdr->identifier, req->dcid, req->scid);
+        return 0;
+    }
+    if (le16toh(req->scid) != chan->dcid) {
         os_mbuf_free_chain(txom);
         ble_hs_unlock();
         return 0;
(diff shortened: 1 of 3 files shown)


nimble/host/src/ble_l2cap_sig.c
```diff
@@ -1543,6 +1543,7 @@ ble_l2cap_sig_disc_req_rx(uint16_t conn_handle, struct ble_l2cap_sig_hdr *hdr,
     if (!chan || (le16toh(req->scid) != chan->dcid)) {
         os_mbuf_free_chain(txom);
         ble_hs_unlock();
+        ble_l2cap_sig_reject_invalid_cid_tx(conn_handle, hdr->identifier, req->dcid, req->scid);
```
rymanluk, line 1546: actually we are doing here wrong.

as per Bluetooth specification "If the DCID is not recognized by the
receiver of this message, an L2CAP_COMMAND_REJECT_RSP
packet with ‘invalid CID’ result code shall be sent in response. If the
receiver finds a DCID match but the SCID fails to find the same match, the
request should be silently discarded"

That means - if (!chan) { we should send REJECT}
if ((le16toh(req->scid) != chan->dcid))  -> we should silently discard a message

rymanluk (changes requested):

nimble/host/src/ble_l2cap_sig.c
KKopyscinski: I pushed the change but I checked it out again and turns out that this requires another one PTS issue ticket: TS says that lower tester should send invalid DCID, but PTS sends invalid SCID and valid DCID and expects disconnection. Of course what we got here is OK, it just allowed test to pass when it was wrong (before the push)

rymanluk (changes requested):
can we have unit test for that?

rymanluk (approved):
looks good to me. Having unit test for that would also be good.

KKopyscinski:
Added requested unit test

PR #1002, opened 2021-07-20
host/l2cap: disconnect channel if received more than expected

KKopyscinski:
If receive data lenght exceeds what was defined in first packet
disconnect with peer.

This is affecting L2CAP/LE/CFC/BV-28-C

### DIFF
--- a/nimble/host/src/ble_l2cap_coc.c
+++ b/nimble/host/src/ble_l2cap_coc.c
@@ -198,6 +198,17 @@ ble_l2cap_coc_rx_fn(struct ble_l2cap_chan *chan)
         }
 
         sdu_len = get_le16((*om)->om_data);
+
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_l2cap_coc.c
```diff
@@ -255,6 +255,11 @@ ble_l2cap_coc_rx_fn(struct ble_l2cap_chan *chan)
         return 0;
     }
 
+    if (OS_MBUF_PKTLEN(rx->sdu) > rx->data_offset) {
```
rymanluk, line 258: This seems to be too late.
I think we should add checks somewhere around line 201 and in case of fragmentation we should add additional checks  around 229.

In addition, when we decide to disconnect, we probably need to clear already received data.

rymanluk (changes requested):

sjanc (approved):

PR #1003, opened 2021-07-20
host/l2cap: disconnect if received packet is larger than MPS

KKopyscinski:
Peer sending packet larger than MPS is invalid, and should be met
with L2CAP channel disconnection.

This affects L2CAP/LE/CFC/BV-27-C


nimble/host/src/ble_l2cap.c
```diff
@@ -387,6 +387,15 @@ ble_l2cap_rx(struct ble_hs_conn *conn,
             goto err;
         }
 
+        if (l2cap_hdr.len > chan->my_coc_mps) {
```
rymanluk, line 390: what about ATT ? in such a case chan->my_coc_mps is 0 I believe. Or not?

rymanluk (changes requested):

nimble/host/src/ble_l2cap.c
KKopyscinski: Yes, I think you are right. What's more, I think only dynamic range CIDs use `my_coc_mps` field, and for others MPS=MTU of channel, which varies from channel to channel. I added 3 cases for `BLE_L2CAP_CID_ATT`, `BLE_L2CAP_CID_SIG` and `BLE_L2CAP_CID_SM`.

nimble/host/src/ble_l2cap.c
```diff
@@ -387,6 +387,23 @@ ble_l2cap_rx(struct ble_hs_conn *conn,
             goto err;
         }
 
+        if ((chan->dcid >= 0x0040 && chan->dcid <= 0x007F && l2cap_hdr.len > chan->my_coc_mps) ||
```
rymanluk, line 390: @KKopyscinski  don't you want just add `ble_l2cap_disconnect(chan);` after line 413 ?

KKopyscinski:
@sjanc as discussed offline, only CIDs from dynamic range are checked now

sjanc (approved):

sjanc (approved):

PR #1006, opened 2021-07-28
nimble/host : Check stack status before executing stack command

rahult-github:
Added change of checking stack status in the APIs exposed to user. This check exists in few functions, extending them to all.


rahult-github:
Hi @sjanc  , @rymanluk  , can you please review and share your comments, if any.

PR #1007, opened 2021-07-30
nimble/ll: Use callout for scan duration and period

sjanc:
hal_timer executes from ISR which was preempting LL task
and could result in HCI flow issues if unlucky.


vrahane:
Tested this out, nrf5340 seems quite stable with periodic scanning.

vrahane (approved):

PR #1008, opened 2021-07-30
transport/nrf5340: Fix assignment in assert

sjanc:
Fix typo in assert which make it never catch intended error.


andrzej-kaczmarek (approved):
:1st_place_medal:

PR #1012, opened 2021-08-25
nimble/host: Sync sending Host Number of Completed Packets command

sjanc:
Host Number of Completed Packets command is special as it doesn't
generate status event and thus is not send with ble_hs_hci_cmd_tx.

We should stil sync sending it with ble_hs_hci_cmd_tx() though to
avoid requesting more than one command buffer from HCI transport
(which may not be able to provide more).


rymanluk (approved):
LGTM

PR #1013, opened 2021-08-29
Change GATT notify/indicate from gattc to gatts

blueMoods:
GATT notify/indicate are functions of GATTS, so we should change `ble_gattc_notify`/`ble_gattc_indicate` to `ble_gatts_notify`/`ble_gatts_indicate`

### DIFF
--- a/apps/blecsc/src/gatt_svr.c
+++ b/apps/blecsc/src/gatt_svr.c
@@ -264,7 +264,7 @@ gatt_svr_chr_access_sc_control_point(uint16_t conn_handle,
     }
 #endif
 
-    rc = ble_gattc_indicate_custom(conn_handle, csc_control_point_handle,
+    rc = ble_gatts_indicate_custom(conn_handle, csc_control_point_handle,
                                    om_indication);
 
     return rc;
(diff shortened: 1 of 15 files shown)


sjanc (changes requested):
Hi,

I'm fine with rename but we should try to keep API compatible.

I suggest we add wrappers for old API with note that it is deprecated and new one should be used instead eg:

/**
 * Deprecated. Should not be used.  Use ble_gatts_notify_custom instead.
 */
static inline int ble_gattc_notify_custom(uint16_t conn_handle, uint16_t att_handle, struct os_mbuf *om)
{
    return ble_gatts_notify_custom(conn_handle, att_handle, om);
}


This should be done for all public API (ie from gatt.h that was renamed)

blueMoods:
> Hi,
> 
> I'm fine with rename but we should try to keep API compatible.
> 
> I suggest we add wrappers for old API with note that it is deprecated and new one should be used instead eg:
> 
> /**
> 
> * Deprecated. Should not be used.  Use ble_gatts_notify_custom instead.
>   */
>   static inline int ble_gattc_notify_custom(uint16_t conn_handle, uint16_t att_handle, struct os_mbuf *om)
>   {
>   return ble_gatts_notify_custom(conn_handle, att_handle, om);
>   }
> 
> This should be done for all public API (ie from gatt.h that was renamed)

@sjanc issues fixed, thanks

sjanc (approved):

blueMoods:
> Hi, sorry for late reply, could you please rebase PR so that it can be merged?

@sjanc Having rebased.

sjanc:
Hi,

Looks like in latest rebase you dropped backward compatibility wrappers as discussed before (hence the build failure)

blueMoods:
Hi @sjanc , Having updated, but the CI is still failed.

It seem like the error is nothing to do with my change, such as:

`nimble/host/test/src/ble_hs_test_util.c:1852:1: error: redefinition of ‘ble_transport_to_ll_acl’`

PR #1024, opened 2021-09-21
apps/bttester: add command to force pairing

KKopyscinski:
Normally, we let to choose security procedure by
ble_gap_security_initiate(). Some tests require though to pair again
with the same peer without breakign the link - gap_unpair cannot
be used. Let's ad gap_force_pair to expicitly call
ble_gap_pair_initiate.

### DIFF
--- a/apps/bttester/src/gap.c
+++ b/apps/bttester/src/gap.c
@@ -1301,6 +1301,37 @@ rsp:
 	tester_rsp(BTP_SERVICE_ID_GAP, GAP_PAIR, CONTROLLER_INDEX, status);
 }
 
+static void force_pair(const uint8_t *data, uint16_t len)
+{
+    /**
+     * This method is used to force pairing instead of giving the option
+     * of automatically choosing pairing or encryption by ble_gap_security_initiate()
+     */
+    struct ble_gap_conn_desc desc;
+    uint8_t status;
+    int rc;
+
+    SYS_LOG_DBG("");
+
+    rc = gap_conn_find_by_addr((ble_addr_t *)data, &desc);
+    if (rc) {
+        status = BTP_STATUS_FAILED;
+        goto rsp;
+    }
+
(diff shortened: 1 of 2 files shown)


KKopyscinski:
This requires https://github.com/apache/mynewt-nimble/pull/1032 to pass the tests

KKopyscinski:
Removed gap_force_enc, because it is equivalent of gap_pair when peers are already bonded, and they always are when PTS requests encryption.

apps/bttester/src/gap.c
```diff
@@ -1301,6 +1301,37 @@ static void pair(const uint8_t *data, uint16_t len)
...
+    rc = ble_gap_pair_initiate(desc.conn_handle);
+    SYS_LOG_DBG("ble_gap_pair_initiate finished with rc=%d", rc);
+    if (rc) {
+        status = BTP_STATUS_FAILED;
+        goto rsp;
+    }
+
+    status = BTP_STATUS_SUCCESS;
+
+    rsp:
```
rymanluk, line 1331: indentation issue.

apps/bttester/src/gap.c
```diff
@@ -1608,6 +1639,9 @@ void tester_handle_gap(uint8_t opcode, uint8_t index, uint8_t *data,
 	case GAP_PAIR:
 		pair(data, len);
 		return;
+    case GAP_FORCE_PAIR:
+        force_pair(data, len);
```
rymanluk, line 1643: indentation issue?

apps/bttester/src/gap.c
KKopyscinski: solved

apps/bttester/src/gap.c
KKopyscinski: solved

rymanluk (approved):
looks good to me

KKopyscinski:
No longer needed as of PTS 8.2.2

PR #1032, opened 2021-09-28
host/ble_sm: delete old security information before saving new

KKopyscinski:
Pairing might occur again with the same peer, for example on call of
test function ble_gap_pair_initiate(). This will lead to adding new
keys to storage for the same peer, but old ones will still exist.
In the future encryption might be initiated, and old keys will be
chosen, which will lead to encryption failure and disconnection with
MIC failure reason code, despite that the correct LTK is saved. Before
saving new keys, check if others aren't already saved for the peer,
and if they do, delete them.


michal-narajowski (approved):

michal-narajowski:
@sjanc ok to merge?

andrzej-kaczmarek:
https://github.com/apache/mynewt-nimble/pull/1473

PR #1033, opened 2021-09-29
apps/bttester: implement NRPA rotation

KKopyscinski:
As per CORE v5.3, Vol 3, Part C 10.7.1 NRPA should also rotate, same
as RPA is doing right now. For time beeing let's implement this in
tester application, before it's implemented in host.

This affects GAP/BROB/BCST/BV-04-C

### DIFF
--- a/apps/bttester/src/gap.c
+++ b/apps/bttester/src/gap.c
@@ -63,6 +63,10 @@ static struct os_callout connected_ev_co;
 static struct gap_device_connected_ev connected_ev;
 #define CONNECTED_EV_DELAY_MS(itvl) 8 * BLE_HCI_CONN_ITVL * itvl / 1000
 static int connection_attempts;
+#if MYNEWT_VAL(BTTESTER_PRIVACY_MODE) && MYNEWT_VAL(BTTESTER_USE_NRPA)
+static int64_t advertising_start;
+static struct os_callout bttester_nrpa_rotate_timer;
+#endif
 
 static const struct ble_gap_conn_params dflt_conn_params = {
 	.scan_itvl = 0x0010,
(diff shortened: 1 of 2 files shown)


apps/bttester/src/gap.c
```diff
@@ -63,6 +63,10 @@ static struct os_callout connected_ev_co;
 static struct gap_device_connected_ev connected_ev;
 #define CONNECTED_EV_DELAY_MS(itvl) 8 * BLE_HCI_CONN_ITVL * itvl / 1000
 static int connection_attempts;
+#if MYNEWT_VAL(BTTESTER_PRIVACY_MODE) && MYNEWT_VAL(BTTESTER_USE_NRPA)
+static ble_npl_time_t advertising_start;
```
andrzej-kaczmarek, line 67: bttester does not use npl

KKopyscinski: switched to os_time and fixed logic

apps/bttester/src/gap.c
```diff
@@ -248,6 +252,40 @@ static struct ble_gap_adv_params adv_params = {
 	.disc_mode = BLE_GAP_DISC_MODE_NON,
 };
 
+#if MYNEWT_VAL(BTTESTER_PRIVACY_MODE) && MYNEWT_VAL(BTTESTER_USE_NRPA)
+static void rotate_nrpa_cb(struct os_event *ev)
+{
+    assert(ev != NULL);
```
rymanluk, line 258: this assert seems to be not needed. Otherwise looks good. Please remove it and feel free to merge

rymanluk (approved):
One comment only

PR #1041, opened 2021-09-30
nimble/ll: Refactor item scheduling

andrzej-kaczmarek:
This refactors all scheduling APIs (i.e. insertions to scheduler) to
use common code instead of reimplementing the same or similar flow
over and over again. The single 'ble_ll_sched_insert' call covers all
currently possible scenarios with minimal overhead so all existing
insertion APIs use it and add some more extra processing if needed.
Also it adds some flexibility via preemption callback so it's possible
to better define priorities for each item.


andrzej-kaczmarek:
fyi, this passes all LL test cases

wes3 (approved):
I read through all the code and I did not see anything that I thought was wrong. Seems like excellent changes and this was due for re-factor for a long time. Good job! I think it could use a few more header comments to describe return values and such but not a big deal.

andrzej-kaczmarek:
rebased + added comment

PR #1044, opened 2021-10-04
Linux blemesh port fixes

sjanc:


sjanc:
This is a bit of and RFC...  maybe word aligment is not needed on all platforms and check in create_free_list() could be relaxed instead?

KKopyscinski (approved):
Imo this is OK, because it's done in the similar way that in Zephyr: there, `block_size` is rounded up to next word boundary to be word alligned. That way of initializing it will help if in the future memslab functions change and we won't need to accommodate them to our initializer.

PR #1051, opened 2021-10-06
nimble/ll: Fix scanning with LL Privacy disabled

andrzej-kaczmarek:
This is regression after eed4389851fcf96922a2e429376808f2dfddc912.
Without this fix only SCAN_RSP are reported back to host...


andrzej-kaczmarek:
@haukepetersen fixed

PR #1073, opened 2021-11-04
targets/nordic_pca10095-blehci: Enable coredump support

kasjer:
This enables (by default) coredump generation on net core.
It also enables net core flash server that can be used
to get coredumps from application core.


vrahane:
@kasjer `img_mgmt.h` needs to be included conditionally based on `LOG_SOFT_RESET` in `repos/apache-mynewt-mcumgr/cmd/os_mgmt/port/mynewt/src/mynewt_os_mgmt.c`, that's why the build is failing.

PR #1074, opened 2021-11-04
nimble/phy/nrf: Remove workrounds for ifs on LE Coded

andrzej-kaczmarek:
BLE_PHY_CODED_RX_IFS_EXTRA_MARGIN was added with initial implementation
for LE Coded to workaround problems with some early releases of devices
that could not keep ifs timing properly. We do not want it anymore, all
devices should have ifs implemented properly in 2021 already.


rymanluk:
since this is debugging option - why to remove it?
Imho even in 2021 Coded has not reach its momentum - I would keep this nice debug feature.

sjanc:
Hi,

I'm fine with removing this, I don't think we should have syscfgs that allows to behave against the spec

sjanc (approved):

PR #1075, opened 2021-11-05
nimble/ll: Add support for vs hci commands registration

andrzej-kaczmarek:
This adds API to allow registration of custom HCI VS commands.

### DIFF
--- a/nimble/controller/src/ble_ll_hci_vs.c
+++ b/nimble/controller/src/ble_ll_hci_vs.c
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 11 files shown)


nimble/controller/src/ble_ll_hci_vs.c
```diff
@@ -0,0 +1,117 @@
...
+}
+
+void
+ble_ll_hci_vs_register(struct ble_ll_hci_vs_cmd *cmds, uint8_t num_cmds)
+{
+    /* Assume all cmds are registered early on init, so just assert in case of
+     * invalid request since it means something is wrong with the code itself.
+     */
+
+    for (uint8_t i = 0; i < num_cmds; i++, cmds++) {
```
sjanc, line 100: nitpick:  i should be declared at the top of function

also, not sure why this is uint8_t and not just unsigned int

sjanc (approved):
I guess we should start documenting mynewt specific VS HCI commands too :-)

Issue #1076, opened 2021-11-05
ll: assertion fail in ble_ll_scan_aux.c


haukepetersen:
We currently run some RIOT nodes that use extended advertisements to communicate with each other. In particular, the advertisements are either `directed` or `undirected`, and all of them are `non-connectable` and `non-scannable`. In our configuration, the nodes are scanning with a 100% duty cycle (scan itvl := scan win) and are advertising data in short bursts every second or so.

Now when running a couple of nodes, they crash after random amounts of time, all showing the same assertion failure, e.g.:
```
1636113458.598309;nrf52dk-10;build/pkg/nimble/nimble/controller/src/ble_ll_scan_aux.c:1718 => 0x779                    
1636113458.600407;nrf52dk-10;*** RIOT kernel panic:                                                                    
1636113458.602457;nrf52dk-10;FAILED ASSERTION.  
```

I am pretty sure we were able to run a similar setup without any trouble some months ago, so could this be related to the latest restructuring of code in the controller?

I have tried to go through the controller code, but I can't find an obvious reason for `aux_data_current` being set `NULL` before `ble_ll_scan_aux_halt()` is called. Could it be a concurrency issue? If I see it correctly, `ble_ll_scan_aux_halt` is only used from interrupt context. And also all functions that change the value of `aux_data_current` are executed in the same context, right? 

Any idea what could be causing our nodes to crash?

For completeness, we use the RIOT default configuration (as in `porting/npl/riot/include/syscfg/syscfg.h`), with the following changes:
```
MYNEWT_VAL_MSYS_1_BLOCK_SIZE=294
MYNEWT_VAL_MSYS_1_BLOCK_COUNT=32
MYNEWT_VAL_BLE_MULTI_ADV_INSTANCES=2
MYNEWT_VAL_BLE_EXT_ADV=1
MYNEWT_VAL_BLE_HCI_EVT_HI_BUF_COUNT=12
MYNEWT_VAL_BLE_LL_EXT_ADV_AUX_PTR_CNT=10
MYNEWT_VAL_BLE_LL_CFG_FEAT_LL_EXT_ADV=1
MYNEWT_VAL_BLE_HCI_EVT_BUF_SIZE=257
MYNEWT_VAL_BLE_EXT_ADV_MAX_SIZE=1650
```

andrzej-kaczmarek:
[...]

PR #1077, opened 2021-11-05
nimble/ll: Some fixes for ll_scan_aux

andrzej-kaczmarek:


nimble/controller/src/ble_ll_scan.c
```diff
@@ -1928,6 +1928,22 @@ ble_ll_scan_rx_pkt_in(uint8_t ptype, struct os_mbuf *om, struct ble_mbuf_hdr *hd
     struct ble_ll_scan_sm *scansm;
     struct ble_ll_scan_addr_data addrd;
     uint8_t *targeta;
+    uint8_t max_pdu_type;
+
+    scansm = &g_ble_ll_scan_sm;
+    rxinfo = &hdr->rxinfo;
+
+    /* Ignore PDUs we do not expect here */
+    max_pdu_type = BLE_ADV_PDU_TYPE_ADV_SCAN_IND;
```
haukepetersen, line 1937: If I didn't miss anything, this filter is incomplete: in case extended advertisements are not enabled and an incoming packet is of type `BLE_ADV_PDU_TYPE_AUX_CONNECT_REQ`, it passes this filter (`BLE_ADV_PDU_TYPE_AUX_CONNECT_REQ := 5` which is <= `BLE_ADV_PDU_TYPE_ADV_SCAN_IND := 6`). 
The same goes for packets with type `BLE_ADV_PDU_TYPE_AUX_SCAN_REQ`...

haukepetersen (changes requested):
My first tests look very promising, after I rebased my working branch on top of this PR (including the important https://github.com/apache/mynewt-nimble/pull/1077/commits/8bcd2db1105352a99fe575e2ae4490178970d8ef), none of the nodes crash anymore with the failed assertion as described in #1076.

I did however spot a minor issue in the code that we might want to address before merging this.

nimble/controller/src/ble_ll_scan.c
andrzej-kaczmarek: AUX_CONNECT_REQ and AUX_SCAN_REQ are the same PDUs as CONNECT_IND and SCAN_REQ and thus they have the same PDU type. The only difference is the channel on which they are sent - ble_ll_scan only scans on primary advertising channels.

sjanc (approved):

nimble/controller/src/ble_ll_scan.c
haukepetersen: I see, makes sense. That is where my current knowledge about extended advertisements ends :-)

haukepetersen (approved):

PR #1079, opened 2021-11-05
nimble/ll: Add more convenient way to set pub dev addr

andrzej-kaczmarek:
This adds BLE_LL_PUBLIC_DEV_ADDR which allows to conveniently set
public device address as 48-bit number. It has priority over existing
BLE_PUBLIC_DEV_ADDR which allows to do the same but in a very nasty
way via code injection so is not very intuitive to use.

### DIFF
--- a/nimble/controller/src/ble_ll.c
+++ b/nimble/controller/src/ble_ll.c
@@ -1573,6 +1573,10 @@ ble_ll_init(void)
 {
     int rc;
     uint64_t features;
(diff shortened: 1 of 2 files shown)


sjanc (changes requested):
Lets make old way deprecated in same PR (since host has standard means to read it anyway),  also document that this is for test purposes and shall not be used production build)

sjanc (approved):

PR #1080, opened 2021-11-08
 nimble/ll: Fix encrypted payload length 

andrzej-kaczmarek:
ble_ll_pdu_max_tx_octets_get() returns number of bytes that can be added between LL header and CRC so the complete PDU is no longer than maxTxTime allowed. However, if link is encrypted there will be also MIC added after the payload so effectively we need to stript that extra 4 bytes from allowed payload length.


rymanluk (approved):
LGTM - works fine.

PR #1093, opened 2021-11-16
Revert "l2cap: implement echo request-response procedure"

sjanc:
This reverts commit ddf8043d1b97f1600aa88c53bd935d0fc9b22c9c.

Unfortunatelly this procedure is not allowed on LE transport. This
is affecting L2CAP/LE/REJ/BI-02-C [Command Reject – Reserved PDU Codes]
qualification test case.

### DIFF
--- a/nimble/host/include/host/ble_l2cap.h
+++ b/nimble/host/include/host/ble_l2cap.h
@@ -246,8 +246,6 @@ struct ble_l2cap_chan_info {
 
 typedef int ble_l2cap_event_fn(struct ble_l2cap_event *event, void *arg);
 
-typedef void ble_l2cap_ping_fn(uint16_t conn_handle, uint32_t rtt_ms,
-                               struct os_mbuf *om);
 
 uint16_t ble_l2cap_get_conn_handle(struct ble_l2cap_chan *chan);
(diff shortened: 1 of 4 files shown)


KKopyscinski (approved):
Test is passing after this revert. Spec reference: BLUETOOTH CORE SPECIFICATION Version 5.3 | Vol 3, Part A: 4. SIGNALING PACKET FORMATS Table 4.2

sjanc:
@haukepetersen I think that if you need echo style for testing you probably should implement this on top of L2CAP CoC

haukepetersen:
Too bad :-)

I have definitely missed that part in `BLUETOOTH CORE SPECIFICATION Version 5.3 | Vol 3, Part A: 4. SIGNALING PACKET FORMATS Table 4.2`, no further remarks needed.

When adding that feature, I simply looked at the LL Control PDU description in `Vol 6, Part B, 2.4.2.19` and more important at the `LE Ping procedure` description in `Vol 6, Part B, 5.1.8`. Here it is described, that the LE link layer might indeed send `LL_PING_REQ` packets. Is it possible, that the standard is inconsistent, or is there something else that I have missed?

PR #1098, opened 2021-11-18
host/l2cap_sig: don't allow to set MPS between existing MPS values

KKopyscinski:
If reconfiguration request contains >1 CID and MPS value is reduced for
only one of them this request should still fail. In other words, it's
not allowed to reconfigure MPS to a value between the existing MPS
values of different channels.

This is affecting L2CAP/ECFC/BI-04-C


sjanc (approved):
looks OK but please change commit message to indicate that "MPS on any channel cannot be decreased" as this is what is it all about :)

PR #1099, opened 2021-11-19
host/l2cap: disconnect peer that sends L2CAP packets with hdr len > mtu

KKopyscinski:
Rest of the checks disconnects peer if SDU size exceeds MTU, but value
in header is OK. We should also disconnect if PDU lenght in L2CAP packet
header exceeds it, regardles of how much data it actually contains, not
just return error.

This is affecting L2CAP/LE/CFC/BV-26-C and L2CAP/ECFC/BV-33-C


KKopyscinski:
fixed typo

sjanc (approved):

PR #1101, opened 2021-11-22
transport: Add bridge transport

kasjer:
blehci application is for controller only build with external
interface. In case of dual-core MCUs (NRF5340, DA1469x) controller
runs on second core that may not be able to operate on external
interface. To be able to run blehci like application on dual-core
systems, application side has to route traffic from external interface
transport to internal transport.
So far only one transport package was used, for bridge configuration
application must include two transport packages.
To facilitate this:
- transport package has BLE_HCI_BRIDGE that is set to 1 for dual-core
  bridge configuration
- common syscfg values defined for in all transports are moved
  to transport package (BLE_HCI_EVT_HI_BUF_COUNT..)
- syscfg definitions present in all transport packages are converted to
  syscfg values
- DA1469x and NRF5340 transports in bridge configuration will NOT create
  buffers for commands and events they relay on external transport provided
  buffers


sjanc (approved):

sjanc:
but please fix buidl errors before merging :)

PR #1106, opened 2021-11-25
nimble/sm: Ensure bonded flag is set in sm_result on new pair & bond.

andrewleech:
The bonded flag was not set on new connection in ble_gap_enc_event()

Without this change, the `if (bonded)` was always `false` on re-connection of a bonded peer in: 
https://github.com/apache/mynewt-nimble/blob/0ef8d590db8bc741f3e25610c6e3e6d11dba2c6c/nimble/host/src/ble_gap.c#L5799


sjanc:
please rebase and send PR against master branch

andrewleech:
Thanks for the pointer @sjanc done now.

sjanc (approved):

PR #1110, opened 2021-11-25
Fix HCI_VS guards and refresh RIOT syscfg

andrzej-kaczmarek:


sjanc (approved):

sjanc:
was nimble/controller/include/controller/ble_ll_hci.h left on purpose? (it also uses BLE_HCI_VS instead of _LL_ variant)

sjanc (approved):

PR #1111, opened 2021-11-26
Fix Coverity issues in mesh

KKopyscinski:


nimble/host/mesh/src/proxy_srv.c
```diff
@@ -429,7 +431,7 @@ static int node_id_adv(struct bt_mesh_subnet *sub, int32_t duration)
 	memcpy(proxy_svc_data + 3, tmp + 8, 8);
 
 	err = bt_mesh_adv_start(&fast_adv_param, duration, node_id_ad,
-			      ARRAY_SIZE(node_id_ad), sd, 0);
+                            ARRAY_SIZE(node_id_ad), sd, sd_len);
```
sjanc, line 434: couldn't this just accept NULL instead?

nimble/host/mesh/src/testing.c
```diff
@@ -123,7 +123,7 @@ void bt_test_print_credentials(void)
 	console_printf("IV Index: %08lx\n", (long) bt_mesh.iv_index);
 	console_printf("Dev key: %s\n", bt_hex(bt_mesh.dev_key, 16));
 
-	for (i = 0; i < ARRAY_SIZE(bt_mesh_cdb.app_keys); ++i)
+	for (i = 0; i < ARRAY_SIZE(bt_mesh_cdb.subnets); ++i)
```
sjanc, line 126: this looks kinda odd,   subnets here, but in loop app_key arrays is accessed

are lines 128-131 correct?

nimble/host/mesh/include/mesh/glue.h
```diff
@@ -566,8 +566,12 @@ static inline void k_sem_init(struct k_sem *sem, unsigned int initial_count,
 static inline int k_sem_take(struct k_sem *sem, int32_t timeout)
 {
 	uint32_t ticks;
+	int rc;
 
-	ble_npl_time_ms_to_ticks(timeout, &ticks);
+	rc = ble_npl_time_ms_to_ticks(timeout, &ticks);
+	if (rc) {
+		assert(0);
```
sjanc, line 573: can't we just return error here?

nimble/host/mesh/src/proxy_srv.c
KKopyscinski: Now that I look at it, I see that it just passes `sd` into `bt_le_adv_start` - so I think yes, it already accepts NULL and this is false positive

nimble/host/mesh/src/testing.c
michal-narajowski: I think this is now resolved.

michal-narajowski:
Please rebase.

nimble/host/mesh/src/cfg_cli.c
```diff
@@ -815,17 +815,13 @@ static int node_identity_status(struct bt_mesh_model *model,
 	}
 
 	status = net_buf_simple_pull_u8(buf);
-	net_idx = net_buf_simple_pull_le16(buf) & 0xfff;
+	net_buf_simple_pull_le16(buf);
```
rymanluk, line 818: would be good to explain in commit message why this action is not necessary

KKopyscinski: done

PR #1113, opened 2021-11-29
host/ble_att_svr: Reset handles index on attributes reset.

MarceauFillon:
Resets handles index when server's attributes are reset.
Prevents mix-up when attributes are cached.

### DIFF
--- a/nimble/host/src/ble_att_svr.c
+++ b/nimble/host/src/ble_att_svr.c
@@ -2671,6 +2671,8 @@ ble_att_svr_reset(void)
         ble_att_svr_entry_free(entry);
     }
 
+    ble_att_svr_id = 0;
+    
     /* Note: prep entries do not get freed here because it is assumed there are
      * no established connections.
      */


sjanc:
you may need to rebase to make CI happy

MarceauFillon:
@sjanc is the target branch wrong?

sjanc:
ahh yes, please make PR against master branch

MarceauFillon:
Done, thanks.

andrewleech:
As some background to this change, we're using nimble as the primary BLE stack in micropython.

During the initial startup of the stack we register the standard gap/gatt services: https://github.com/micropython/micropython/blob/851ecb2da178fff0b60aefdb5af502f28787a7ec/extmod/nimble/modbluetooth_nimble.c#L653

The micropython end user application would later typically define all the desired services/characteristics which is managed dynamically by resetting the registration and building up the required structs to be registered: https://github.com/micropython/micropython/blob/851ecb2da178fff0b60aefdb5af502f28787a7ec/extmod/nimble/modbluetooth_nimble.c#L907

This has been working very well, however we recently discoevered that the handles would continue incrementing over each registration reset, causing difficulties with any paired devices which have cached services/characteristics.

sjanc (approved):

PR #1143, opened 2022-01-14
apps/btshell: fix build on 'native' bsp

KKopyscinski:
strl* methods are not always available, let's use snprintf()
for string copy/concat actions in parse_dev_addr, as it allows
us to keep buffer overflow control and NULL termination of string.


sjanc (commented):
OK for now (but fix whitespaces according to coding style), in long term we should switch to latest ubuntu on travis this will allow to use libbsd functions (https://github.com/apache/mynewt-core/pull/2751)

KKopyscinski:
Fixed whitespaces

PR #1145, opened 2022-01-17
nimble/phy/nrf: Add support for reading Ublox BMD-345 public address

sjanc:
The BMD-345 modules are preprogrammed from the factory with a unique
public Bluetooth device address stored in the CUSTOMER[0] and
CUSTOMER[1] registers of the User Information Configuration Registers
(UICR).


kasjer (commented):
+1

PR #1146, opened 2022-01-18
nimble/ll: Adjust start time of the first anchor point of connection

sjanc:
Set the start_time of the first anchor point of new connection to
entry->end_time+1.

For the 32768 Hz crystal in nrf chip, 1 tick is 30.517us.
The connection state machine use anchor point to store the cpu
ticks and anchor_point_usec to store the remainder. Therefore,
to compensate the inaccuracy of the crystal, the ticks of
anchor_point will be add with 1 once the value of
anchor_point_usec exceed 31. If two connections have same
connection interval, the time difference between the two start
of schedule item will decreased 1, which lead to an overlap.
To prevent this from happenning, we set the start_time of sch
to 1 cpu tick after the end_time of entry.

### DIFF
--- a/nimble/controller/src/ble_ll_sched.c
+++ b/nimble/controller/src/ble_ll_sched.c
@@ -237,7 +237,19 @@ ble_ll_sched_insert(struct ble_ll_sched_item *sch, uint32_t max_delay,
                 }
             } else {
                 preempt_first = NULL;
-                sch->start_time = entry->end_time;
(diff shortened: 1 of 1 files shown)


sjanc:
@sada45 sorry, I've miss clicked on your original PR, will merge fix from here

andrzej-kaczmarek (approved):

PR #1155, opened 2022-01-30
nimble/phy/cmac: Use 31250 tps timer

andrzej-kaczmarek:
This adjusts phy to 31250 tps timer used in LL on CMAC.


andrzej-kaczmarek:
see https://github.com/apache/mynewt-core/pull/2761

kasjer (approved):
With those changes throughput tests do not crash cmac.

PR #1160, opened 2022-02-01
nimble/transport/cmac: Flush UART on CMAC error

andrzej-kaczmarek:
When CMAC is used with HCI bridge over UART we should wait for all
data to be flushed from UART before asserting on error, otherwise we
will lose debug events.

### DIFF
--- a/nimble/transport/dialog_cmac/cmac_driver/src/cmac_host.c
+++ b/nimble/transport/dialog_cmac/cmac_driver/src/cmac_host.c
@@ -65,6 +65,13 @@ static struct os_event g_cmac_host_rand_ev = {
 
 static void cmac_host_rand_chk_fill(void);
 
+#if MYNEWT_VAL(BLE_HCI_BRIDGE) && MYNEWT_VAL_CHOICE(BLE_HCI_TRANSPORT, uart)
+static void cmac_host_error_w4flush(struct os_event *ev);
+static struct os_event g_cmac_host_error_ev = {
+        .ev_cb = cmac_host_error_w4flush
+};
+#endif
+
 static void
 cmac2sys_isr(void)
 {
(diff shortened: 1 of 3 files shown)


nimble/transport/dialog_cmac/cmac_driver/src/cmac_host.c
```diff
@@ -147,6 +163,34 @@ cmac_host_rand_chk_fill(void)
...
 }
 
+#if MYNEWT_VAL(BLE_HCI_BRIDGE) && MYNEWT_VAL_CHOICE(BLE_HCI_TRANSPORT, uart)
+#if MYNEWT_VAL(BLE_HCI_UART_PORT) < 0 || MYNEWT_VAL(BLE_HCI_UART_PORT) > 2
+#error Invalid BLE_HCI_UART_PORT
+#endif
+static void
+cmac_host_error_w4flush(struct os_event *ev)
+{
+    static const UART_Type *regs[] = {
```
kasjer, line 173: I guess `const` should be before `regs`

nimble/transport/dialog_cmac/cmac_driver/src/cmac_host.c
```diff
@@ -147,6 +163,34 @@ cmac_host_rand_chk_fill(void)
...
+    if (!ev) {
+        /* Move to task context, we do not want to spin in interrupt */
+        os_eventq_put(os_eventq_dflt_get(), &g_cmac_host_error_ev);
+        return;
+    }
+
+    while ((regs[MYNEWT_VAL(BLE_HCI_UART_PORT)]->UART_LSR_REG &
+            UART_UART_LSR_REG_UART_TEMT_Msk) == 0) {
+        /* Wait until both FIFO and shift registers are empty */
+    };
```
kasjer, line 188: semicolon after `}` is not really needed

kasjer (approved):

andrzej-kaczmarek:
As it turns out, resetting CMAC in cmac2sys interrupt is not a good idea since it probably will be reset before debug events are written to mbox from fault handler. So we better only disable cmac2sys interrupt (so we don't keep spinning in that interrupt since we cannot clear CMAC error) and then wait in task context until both mbox and UART are flushed.

kasjer (approved):
LGTM

PR #1164, opened 2022-02-04
nimble/ll: New connection strict scheduling

andrzej-kaczmarek:
This replaces existing implementation of connection strict scheduling with a new one that allows host to modify scheduling in runtime.

In the simplest case a slot duration and number of slots per period have to be configured. This determines scheduling interval (slot duration) and connection interval for each connection (slot duration * slots per period). Each new central connection will now be scheduled at slot boundary and have the same connection interval, this guaranteed that two connections won't overlap.

By default, for each new connection 1st unused slot is allocated. However, it's possible to allocate specific slot to each connection using hci_vs command in 2 ways:
- set slot for new connection prior to sending connection create command,
- set slot for existing connection at any time, this will move anchor point.

### DIFF
--- a/nimble/controller/include/controller/ble_ll_conn.h
+++ b/nimble/controller/include/controller/ble_ll_conn.h
@@ -230,10 +230,6 @@ struct ble_ll_conn_sm
     uint8_t last_unmapped_chan;
     uint8_t num_used_chans;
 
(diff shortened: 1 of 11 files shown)


sjanc (approved):
looks good, just update commit messages and maybe add some (even basic) description of VS HCI

andrzej-kaczmarek:
updated commit messages

PR #1173, opened 2022-02-16
babblesim: Use pthreads instead of setlongjmp for tasks

andrzej-kaczmarek:
This changes tasks handling to native threads instead of setlongjmp()
which resolves issue with calling the setlongjmp() from nested signal
handlers but also simplifies code, makes debugging much easier and can
work nicely with e.g. Valgrind.

Each task is wrapped in a thread and all threads are synchronized on
a global mutex to make sure only one task executes at any time. If
context switch is requested (this is always done via os_sched() in
critical section), a flag is set to indicate pending context switch
which will be handled after exiting from critical setion and handling
all other pending interrupts. This mimics the way it's done on a real
hardware.


andrzej-kaczmarek:
@mkasenberg @jrotkiewicz

PR #1181, opened 2022-02-17
apps/blehci: Add full console option

kasjer:
So far blehci always used console/stub.
For debugging purpose it may be useful to enable console without
modifying blehci application pkg.deps.

Default value is still stub.

### DIFF
--- a/apps/blehci/syscfg.yml
+++ b/apps/blehci/syscfg.yml
@@ -21,3 +21,5 @@ syscfg.vals:
     OS_MAIN_STACK_SIZE: 64
     # Use UART transport by default
     BLE_HCI_TRANSPORT: uart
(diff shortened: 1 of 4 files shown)


apps/blehci/syscfg.yml
```diff
@@ -16,6 +16,15 @@
...
 #
 
+syscfg.defs:
+    CONSOLE_MODE:
+        description: Which console to use
+        value: stub
+        choices:
+            - full
+            - minimal
+            - stub
```
andrzej-kaczmarek, line 26: I'd add this setting to sys/console instead so we can include sys/console instead of specific package in app and select via syscfg, similar as we did it for hci transport some time ago

PR #1182, opened 2022-02-17
nimble/ll: Fix scan proc handling

andrzej-kaczmarek:
Fall-through for ll_state was not handled properly after some changes
and this means scan proc could disable phy mid-event.


nimble/controller/src/ble_ll_scan.c
```diff
@@ -1091,6 +1091,10 @@ ble_ll_scan_event_proc(struct ble_npl_event *ev)
      */
     start_scan = inside_window;
     switch (ble_ll_state_get()) {
+#if MYNEWT_VAL(BLE_LL_ROLE_BROADCASTER) || MYNEWT_VAL(BLE_LL_ROLE_PERIPHERAL) || \
```
sjanc, line 1094: I'd set start_scan=false explicitly in each case instead of this combined #if

sjanc (approved):

PR #1185, opened 2022-02-21
babblesim/edtt: Refactor queue handling

andrzej-kaczmarek:
This reworks events handling in EDTT.

The "service_events" routine is now handled in separate task which
blocks until new event is put in queue. This is possible since we use
os_eventq to pass events and this means any get from queue will put
task to sleep properly and trigger context switch if necessary.

The blocking edtt_read() will now sleep if not data is available instead
of advancing time machine forward so it does not interfere with timing.
OS_TICKS_PER_SEC was increased to 1024 to have better granularity and
allow ~5ms sleep time for edtt poller task as required.

Also some non-LL code is simplified to use simple calloc/free as it does
not really seem to be cessary to use mempools there, i.e. we just need
some memory block to pass data and it does not really matter how it's
allocated. Also using calloc/free means we will never run out of memory
so no need to check for that.

### DIFF
--- a/babblesim/edtt/hci_transport/src/ble_hci_edtt.c
+++ b/babblesim/edtt/hci_transport/src/ble_hci_edtt.c
@@ -11,8 +11,7 @@
 #include <bs_tracing.h>
 #include "sysinit/sysinit.h"
 #include "syscfg/syscfg.h"
-#include "os/os_cputime.h"
-#include "os/os.h"
+#include "os/os_eventq.h"
 
 /* BLE */
 #include "nimble/ble.h"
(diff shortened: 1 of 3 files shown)


andrzej-kaczmarek:
```
edtt: @00:39:44.260000  Summary:
edtt: @00:39:44.260000  
edtt: @00:39:44.260000  Status   Count
edtt: @00:39:44.260000  ==============
edtt: @00:39:44.260000  PASS        86
edtt: @00:39:44.260000  FAIL        34
edtt: @00:39:44.260000  ==============
edtt: @00:39:44.260000  Total      120
```

sjanc (approved):

PR #1186, opened 2022-02-21
nimble/ll: Fix InitA handling in connectio requests

andrzej-kaczmarek:
As per Core 5.3, Vol 6, Part B, 6.4:
> The Link Layer should not set the InitA field to the same value as
> the TargetA field in the received advertising PDU.

This was apparently omitted during init refactor, so need to bring it
back.

Also simplify generating InitA in case we use RPA and TargetA is not
present - we do not need to lookup for an item in RL since this is
already done by filtering code and we can just use rpa_index.


sjanc (approved):

andrzej-kaczmarek:
merged via https://github.com/apache/mynewt-nimble/pull/1188

PR #1187, opened 2022-02-23
nimble/ll: Add LE Enhanced Connection Update support

andrzej-kaczmarek:
This adds support for LE Enhanced Connection Update feature (Core 5.3) a.k.a. connection subrating.

### DIFF
--- a/nimble/controller/src/ble_ll_conn_hci.c
+++ b/nimble/controller/src/ble_ll_conn_hci.c
@@ -1476,7 +1476,7 @@ ble_ll_conn_hci_set_data_len(const uint8_t *cmdbuf, uint8_t len,
     connsm->host_req_max_tx_time = txtime;
 
     /* If peer does not support coded, we cannot use value larger than 2120us */
-    if (!(connsm->remote_features[0] & (BLE_LL_FEAT_LE_CODED_PHY >> 8))) {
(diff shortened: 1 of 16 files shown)


nimble/controller/src/ble_ll_conn_hci.c
```diff
@@ -1692,6 +1693,102 @@ ble_ll_conn_req_peer_sca(const uint8_t *cmdbuf, uint8_t len,
...
+        } else if (rc == -EBUSY) {
+            return BLE_ERR_CTLR_BUSY;
+        } else if (rc == -ENOTSUP) {
+            return BLE_ERR_UNSUPP_REM_FEATURE;
+        } else {
+            return BLE_ERR_UNSPECIFIED;
+        }
+    }
+
+#if MYNEWT_VAL(BLE_LL_ROLE_CENTRAL)
```
sjanc, line 1778: we could simply make whole command conditional

sjanc (approved):

nimble/controller/src/ble_ll_conn_hci.c
andrzej-kaczmarek: this also can be used for peripheral. for central we need to store those parameters in connsm since they are used to check future requests from peripheral.

PR #1189, opened 2022-02-24
    nimble/hs: fix the LE Supported feature mask

supperthomas:
when set the features we set the htole64 in
    ble_ll_hci.c.It will match the value about it


nimble/host/src/ble_hs_startup.c
```diff
@@ -40,7 +40,7 @@ ble_hs_startup_read_sup_f_tx(void)
     /* for now we don't use it outside of init sequence so check this here
      * LE Supported (Controller) byte 4, bit 6
      */
-    if (!(rsp.features & 0x0000006000000000)) {
+    if (!(rsp.features &  htole64(0x0000006000000000))) {
```
sjanc, line 43: Hi,

I think it would be cleaner to do other way around:
if (!(le64toh(rsp.features) & 0x0000006000000000))) {


that way constant is used in host order, while LE value is converted to host before use

sjanc (changes requested):

nimble/host/src/ble_hs_startup.c
supperthomas: complete .

sjanc (approved):

PR #1191, opened 2022-02-24
apps: enable UART Flow Control in bttester

KKopyscinski:
This prevents data loss when RTT logs collection is enabled un
Upper Tester side.


KKopyscinski:
@mkasenberg FYI

michal-narajowski (approved):

PR #1204, opened 2022-03-09
nimble/ll: Add fake dual mode option

andrzej-kaczmarek:
This adds option to enable some hacks to allow controller to be
initialized by dual-mode host. It basically fakes some HCI command that
we do not support since they are not relevant for BR/EDR, but it allows
for NimBLE controller to be used as a Bluetooth controller e.g. in
Windows 10, at least to some extent.

This option is not intended to be used for production setup, it should
be only used for testing purposes.


kasjer (approved):
Works as advertised, requires public BT address

PR #1210, opened 2022-03-19
apps/btshell: Fix peer_addr_type argument parsing

mkasenberg:
The cmd parser expected a variable named peer_type instead of peer_addr_type.

### DIFF
--- a/apps/btshell/src/cmd.c
+++ b/apps/btshell/src/cmd.c
@@ -103,24 +103,27 @@ parse_dev_addr(const char *prefix, const struct kv_pair *addr_types,
     if (!prefix) {
         name[0] = '\0';
     } else {
-        written = snprintf(name, sizeof(name) - 1, "%s", prefix);
+        written = snprintf(name, sizeof(name), "%s", prefix);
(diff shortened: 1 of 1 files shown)


apps/btshell/src/cmd.c
```diff
@@ -103,23 +103,25 @@ parse_dev_addr(const char *prefix, const struct kv_pair *addr_types,
...
+        written = snprintf(name, sizeof(name), "%s", prefix);
         if (written >= sizeof(name) || written < 0) {
             return EINVAL;
         }
     }
 
-    written = snprintf(name + written, sizeof(name) - written - 1, "%s", "addr");
-    if (written >= sizeof(name) || written < 0) {
+    rc = snprintf(name + written, sizeof(name) - written, "%s", "addr");
+    if (rc >= sizeof(name) || rc < 0) {
```
sjanc, line 113: hmm shouldn't this be   rc >= (sizeof(name) - written)) ?

(same below, but that would require extra val to hold tmp value)

sjanc (approved):

PR #1218, opened 2022-03-23
nimble/ll: Fix missing Read Remote Version Info in supported commands

sjanc:
Read Remote Version Information command bit was not set in supported
HCI commands bitmask.


jrotkiewicz:
tested, works

andrzej-kaczmarek (approved):

PR #1224, opened 2022-03-28
nimble/host: Add HCI commands / events for LE Power control

rahult-github:
LE Power control is BT 5.2 feature. Added support for HCI commands / events for host to controller communication

### DIFF
--- a/nimble/host/include/host/ble_gap.h
+++ b/nimble/host/include/host/ble_gap.h
@@ -135,6 +135,8 @@ struct hci_conn_update;
 #define BLE_GAP_EVENT_PERIODIC_SYNC_LOST    22
 #define BLE_GAP_EVENT_SCAN_REQ_RCVD         23
 #define BLE_GAP_EVENT_PERIODIC_TRANSFER     24
+#define BLE_GAP_EVENT_PATHLOSS_THRESHOLD    25
+#define BLE_GAP_EVENT_TRANSMIT_POWER        26
 
 /*** Reason codes for the subscribe GAP event. */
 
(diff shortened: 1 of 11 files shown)


rahult-github:
Hi @sjanc  , this PR attempts to add host side support for LE Power Control . can you please review the same?

nimble/host/include/host/ble_gap.h
```diff
@@ -973,6 +975,21 @@ struct ble_gap_event {
             uint8_t adv_clk_accuracy;
         } periodic_transfer;
 #endif
+	struct {
```
sjanc, line 978: please comments those events and its members as we have for other events

nimble/host/include/host/ble_gap.h
```diff
@@ -2097,6 +2114,20 @@ int ble_gap_event_listener_register(struct ble_gap_event_listener *listener,
  */
 int ble_gap_event_listener_unregister(struct ble_gap_event_listener *listener);
 
+int ble_gap_set_path_loss_reporting_enable(uint16_t conn_handle, uint8_t enable);
```
sjanc, line 2117: public API should be documented with doxygen

nimble/host/include/host/ble_gap.h
```diff
@@ -2097,6 +2114,20 @@ int ble_gap_event_listener_register(struct ble_gap_event_listener *listener,
...
+
+int ble_gap_set_transmit_power_reporting_enable(uint16_t conn_handle,
+                                            uint8_t local_enable,
```
sjanc, line 2120: some indentation issues here (and functions below too)

sjanc (changes requested):
Overall code looks good (I added some comment thought) but I may have some more once I get through this feature in the spec..

I think it is OK to have support for feature in host that is not yet supported by out controller (I assume you run this with other controller, right?) but it should be optional feature to enable via syscfg (probably with restrictions to host only build until controller support is in place).

rahult-github:
> Overall code looks good (I added some comment thought) but I may have some more once I get through this feature in the spec..
> 
> I think it is OK to have support for feature in host that is not yet supported by out controller (I assume you run this with other controller, right?) but it should be optional feature to enable via syscfg (probably with restrictions to host only build until controller support is in place).

Agreed, i will add all relevant code under feature flag and upload new changes by addressing other comments too .

nimble/host/include/host/ble_gap.h
sjanc: you also need to comment what this event is about and to which event type it match,  see other events (it is for user to know which member in union should it use for type)

nimble/host/include/host/ble_gap.h
```diff
@@ -973,6 +975,42 @@ struct ble_gap_event {
             uint8_t adv_clk_accuracy;
         } periodic_transfer;
 #endif
+
+#if MYNEWT_VAL(BLE_POWER_CTRL)
```
sjanc, line 979: hmm syscfg.yml addition is missing (I suggest to use nimble/syscfg.yml instead of host one as eventually we will get support for this in controller too)

sjanc (changes requested):
couple more comments

nimble/host/include/host/ble_gap.h
rahult-github: Thanks for detailed comments. I have tried to address all.

PR #1225, opened 2022-03-28
nimble/ll: Fix internal flow control for conn

andrzej-kaczmarek:
We need free credit if:
in isr: credit is allocated but ll->hs flow control blocked allocation
in task: credit was allocated for mbuf being freed (e.g. we allocate
         credits only for non-empty PDUs so we should not free a credit
         if this is an empty PDU)


kasjer (approved):
Seems to fix problem with put/get mismatch

sjanc (approved):

PR #1231, opened 2022-04-06
nimble/ll: Optimize BLE_LL_ASSERT

andrzej-kaczmarek:
BLE_LL_ASSERT() now calls ble_ll_assert() which decides whether to break
or send hci_vs event. This significantly reduces code size since there's
no need to include all the branches and calls for each BLE_LL_ASSERT.

Also for hci_vs we only need base file name (i.e. without path) so we
can strip it - __buildiin_strrchr is evaluated during compilation.

This reduces code size for debug build with all features enabled by 5K.

### DIFF
--- a/nimble/controller/include/controller/ble_ll.h
+++ b/nimble/controller/include/controller/ble_ll.h
@@ -41,14 +41,12 @@ extern "C" {
 #ifdef NDEBUG
 #define BLE_LL_ASSERT(cond) (void(0))
 #else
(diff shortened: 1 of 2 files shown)


sjanc (approved):
I assume there is no __FILE_NAME__ or similar provided by compiler ?:)

andrzej-kaczmarek:
`__FILE_NAME__` will be available in GCC 12

PR #1236, opened 2022-04-14
apps/btshell: Add scan filtering based on device name

sjanc:
This allows to specify name prefix that is used to filter out devices
when scanning. Useful in busy environment.

### DIFF
--- a/apps/btshell/src/btshell.h
+++ b/apps/btshell/src/btshell.h
@@ -83,10 +83,14 @@ struct btshell_conn {
     struct btshell_l2cap_coc_list coc_list;
 };
 
+#define NAME_FILTER_LEN_MAX 20
+
 struct btshell_scan_opts {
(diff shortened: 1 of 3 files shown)


apps/btshell/src/btshell.h
```diff
@@ -83,10 +83,14 @@ struct btshell_conn {
...
 };
 
+#define NAME_FILTER_LEN_MAX 20
+
 struct btshell_scan_opts {
     uint16_t limit;
     uint8_t ignore_legacy:1;
     uint8_t periodic_only:1;
+    uint8_t name_filter;
+    char name[NAME_FILTER_LEN_MAX];
```
andrzej-kaczmarek, line 93: `name_filter` and `name_filter_len` perhaps...?

andrzej-kaczmarek (approved):

PR #1254, opened 2022-05-05
nimble/phy/nrf53: Fix PA/LNA enable

andrzej-kaczmarek:
READY/DISABLED events were not published if PA/LNA was enabled so it was
not working - it only worked if GPIO debugging is also enabled which
published those events.


utzig:
## Style check summary

sjanc (approved):

PR #1265, opened 2022-05-19
host/ble_att_srv: fix ble_att_svr_check_perms for SC Only mode

KKopyscinski:
If security is not good enough, return BLE_ATT_ERR_INSUFFICIENT_AUTHEN.
If security is enabled, but key is to short, return
BLE_ATT_ERR_INSUFFICIENT_KEY_SZ


utzig:
## Style check summary

sjanc (approved):

PR #1266, opened 2022-05-19
nimble/host/ble_sm: add restriciton for SC Only

KKopyscinski:
This mode requires MITM to be on.

### DIFF
--- a/nimble/host/syscfg.yml
+++ b/nimble/host/syscfg.yml
@@ -104,6 +104,9 @@ syscfg.defs:
         value: 0
         restrictions:
             - 'BLE_SM_SC_LVL == 4 if 1'
+            - BLE_SM_MITM
+            - 'BLE_SM_SC if 1'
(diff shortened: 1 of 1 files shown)


utzig:
## Style check summary

nimble/host/syscfg.yml
```diff
@@ -104,6 +104,7 @@ syscfg.defs:
         value: 0
         restrictions:
             - 'BLE_SM_SC_LVL == 4 if 1'
+            - BLE_SM_MITM
```
sjanc, line 107: should this also force BLE_SM_SC:1  and BLE_SM_LEGACY:0 ?

KKopyscinski: I haven't though about it, but yes, it makes sense - how could we force SC level on legacy? I'll update this PR

sjanc (approved):

PR #1267, opened 2022-05-23
tools/hci_throughput: Initial version of the HCI throughput tester

jrotkiewicz:
Python tool for measuring throughput over BLE on Linux. User provides HCI indexes of connected controllers, the tool configures them and starts throughput test. More in README file. 

@sjanc @kasjer


jrotkiewicz:
sjanc

PR #1272, opened 2022-06-07
controller: use OS specific PRNG for RIOT builds

haukepetersen:
Currently the controller is using jrand48() to generate pseudo
random numbers. For RIOT builds this commit switches to the RIOT
build-in PRNG API, as this has two benefits: i) it saves 500bytes
of flash and ii) jrand48() internally uses malloc, which has been
source to runtime failures in the past.


andrzej-kaczmarek (approved):

haukepetersen:
All green -> go

PR #1278, opened 2022-06-20
ble_conn: fix auth_pyld_timer memleak

bvdberg:
Every time a connection is made, the  ble_npl_callout_init() is done.
This causes a timer creation in some NPL cases (eg. FreeRTOS).
The alloc behaviour of auth_pyld_timer is now in sync with ctrl_proc_rsp_timer
that lives in the same struct.


sjanc:
I'm fine with having this hotfixed as is (although the leak will stil be present on each HCI reset call, right?)

in long term I think we need to expand npl api so that every init function has cleanup/deinit couterpart  (or assume that some accounting will be handled by NPL layer eg freertos NPL could have internal list of timers and know if one was already created or not..)

bvdberg:
I'm not sure about that. I see that at start, all connections you define (Max-Connections) get created and the timer also.
They are never freed, since the Max number of connections are compile-time defined. When a connection is reset (eg failed to connect) the state is reset, but the timer is not alloc'ed/freed, so that is fine.

bvdberg:
rebased on update master

sjanc (approved):

PR #1280, opened 2022-06-20
nimble/ll:: Make LL task stack size configurable

sjanc:
This commit makes LL stack size configurable. Depending on compiler and
platform default value may not be enough. It was reported that on nRF52840
90 words is not enough even if Ext Adv is disabled. To keep things simple
lets just increase default stack size to 120 word for all usecases (with
exception of nRF51 where defults to 96 bytes).


andrzej-kaczmarek:
why this is defined in phy instead of controller pkg?

andrzej-kaczmarek (approved):

PR #1305, opened 2022-07-25
use true block size to calculate net buf id

rinehaidvon:


sjanc:
Hi,

could you update commit message with a bit of explanation on what is being fixed by this?

PR #1321, opened 2022-08-12
nimble/ll/css: Add vs event to indicate slot change

andrzej-kaczmarek:
If hci_vs support is enabled for CSS, an event will be sent every time
slot chanes for a connection. Event is also sent after connection is
created (after conn complete and csa events).


andrzej-kaczmarek:
added in https://github.com/apache/mynewt-nimble/pull/1334

PR #1327, opened 2022-08-22
host/ble_gap.c: Fix failure to run callback on finished scan

m-gorecki:
When both roles - peripheral and central where turned off, removed lines where causing
not calling callback function after completing the scan.


rymanluk (approved):
Looks good to me.

@KKopyscinski , any comments?

KKopyscinski:
Nice catch! LGTM also

KKopyscinski (approved):

PR #1328, opened 2022-08-22
nimble: Fix hci defs order

andrzej-kaczmarek:


nimble/include/nimble/hci_common.h
```diff
@@ -1082,6 +1082,48 @@ struct ble_hci_le_subrate_req_cp {
     uint16_t supervision_tmo;
 } __attribute__((packed));
 
+#define BLE_HCI_OCF_LE_ENH_READ_TRANSMIT_POWER_LEVEL     (0x0076)
```
sjanc, line 1085: hmm shouldn't those be before 0x007D ?

andrzej-kaczmarek: let me think about it :P

sjanc (approved):

PR #1332, opened 2022-08-22
nimble/ll: Fix aux scan timing

andrzej-kaczmarek:
We should include clock drift due to sleep clock accuracy and jitter
when calculating aux scan start time (Core 5.3, Vol 6, Part B, 4.2.2).

Slot duration calculation should also include the above and possible
drift due to offset accuracy.


andrzej-kaczmarek:
rebased on top of https://github.com/apache/mynewt-nimble/pull/1333

sjanc (approved):

PR #1342, opened 2022-08-31
nimble/ll: Fix not setting TX power in HCI VS command

sjanc:
PHY was not configured to use new TX power immediatelly. This resulted
in new TX power being used only after advertising event ended.


andrzej-kaczmarek (approved):

andrzej-kaczmarek (changes requested):
hmmm actually I am not sure this is ok - it's possible that you will change power mid-tx if this is done during other activity...

andrzej-kaczmarek (approved):

PR #1344, opened 2022-08-31
nimble/phy: Fix fem enable timings

andrzej-kaczmarek:


andrzej-kaczmarek:
fixed in https://github.com/apache/mynewt-nimble/pull/1354

PR #1354, opened 2022-09-08
nimble/phy: Introduce common phy for nRF52 and nRF53

andrzej-kaczmarek:
This is first cut of common nrf5x phy which works for both nRF52 and nRF53. The goal is to have the same code used on both series in order to make maintenance easier and have the same fixes and features supported.


andrzej-kaczmarek:
rebased on https://github.com/apache/mynewt-nimble/pull/1345

andrzej-kaczmarek:
@haukepetersen fyi, this needs to be updated in RIOT

sjanc (approved):

PR #1359, opened 2022-09-15
nimble/ll: Add support for custom TX power configuration

sjanc:
This allows to provide custom TX power setting and rounding functions which can take into account FEM configuration. Since this is highly dependend on HW configuration (FEM type and mode, BPFs, PA power etc) there is no point in providing "default" implementation as this should be typically provided by BSP or application.

### DIFF
--- a/nimble/controller/include/controller/ble_ll_fem.h
+++ b/nimble/controller/include/controller/ble_ll_fem.h
@@ -30,6 +30,11 @@ extern "C" {
 void ble_ll_fem_pa_init(void);
 void ble_ll_fem_pa_enable(void);
 void ble_ll_fem_pa_disable(void);
+
+#if MYNEWT_VAL(BLE_LL_FEM_PA_TXPOWER_TUNE)
+void ble_ll_fem_txpower_set(int dbm);
+int ble_ll_fem_txpower_round(int dbm);
+#endif
 #endif
 
 #if MYNEWT_VAL(BLE_LL_FEM_LNA)
--- a/nimble/controller/src/ble_ll.c
+++ b/nimble/controller/src/ble_ll.c
@@ -65,7 +65,7 @@
  * right thing to do.
  */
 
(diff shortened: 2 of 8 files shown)


andrzej-kaczmarek:
so I thought it over and I think it should work a bit different:
- LL should always set TX power instead of relying on FEM to do this
- we should have API to get PA/LNA gain, LL should call it and then calculate actual TX power or RSSI using that value and also RF path compensation, if set by host
- optionally we can have syscfg to define constant gain for PA/LNA to have some build time optimizations possible

proposed names for APIs would be ble_ll_fem_(pa|lna)_gain_get()
syscfg BLE_LL_FEM_(PA|LNA)_GAIN - if set to 0 means LL has to call API to get actual gain, otherwise use that value

sjanc:
The thing is that FEM might not have fixed gain,  ie in SKY66112 TX gain depends on SKY configuration _and_ input power, so LL will have to iterate (?) to match expected output power with whatever can be set by phy, alternatively we can add some output parameters where FEM returns also expected PHY tx power

andrzej-kaczmarek:
The "current" solution will call FEM API which will set tx power via phy. Proposed solution will call FEM API to obtain tx power and set it via phy - I suspect this will be the same value, so what is the problem here?

sjanc:
"LL should call it and then calculate actual TX power or RSSI using that value and also RF path compensation, if set by host"

this is other way around, only FEM can calculate actual TX power since it (may) depends on input power and FEM (possibly runtime) configuration.

Also to cover all scenarios, like rounding, reading supported power ranges etc we need ble_ll_fem_pa_gain_set() which will configure FEM as needed (as oppose to get() which would just return values for rounding)


Than, ble_ll_fem_pa_gain_get() would return actually configured gain (compared to PHY tx power that was last "set")

sjanc:
this will be handled in https://github.com/apache/mynewt-nimble/pull/1374

PR #1372, opened 2022-09-26
apps: Add privacy samples

m-gorecki:
Sample applications privacy_central and privacy_periph, showing how to configure privacy.


br101:
Those examples would be very useful

PR #1378, opened 2022-09-27
nimble/ll: Fix MIC failure during encryption start

andrzej-kaczmarek:
This fixes unexpected MIC failure when retransmission happens during encryption start procedure as follows:
- peripheral sends LL_START_ENC_REQ unencrypted, central acks
- central sends LL_START_ENC_RSP encrypted, peripheral acks
- central retransmits LL_START_ENC_RSP for whatever reason

The problem is that peripheral increments rx packet counter after 1st LL_START_ENC_RSP is received, so retransmission is decrypted with different rx packet counter and thus is not valid. We properly ignore MIC failure for retransmission, but then code checks if received PDU is valid in currect state, i.e. encryption start procedure. Since it was not properly decrypted, the PDU type is likely garbage and thus considered as not allowed so we terminate connection with MIC failure.

The "ultimate" fix for such issues is to simply ignore any retransmitted PDU with MIC failure since basically contents of such PDUs are garbage and not really useful for any checks.


sjanc (approved):

jrotkiewicz:
Tested few hundred of times and I could not reproduce mic failure.

PR #1379, opened 2022-09-27
nimble/transport: Add generic IPC transport with flow control

andrzej-kaczmarek:
This adds generic IPC transport that should be used as a base for any transport with LL on separate core. It's similar to HCI H4, but has a bit different header to simplify processing and supports internal flow control to make sure LL does not send data that other core does not have free buffers for.

IPC transport requires memory shared between cores to keep track of number of available buffers on application core. Application core will initialize counters to total number of available buffers. Then on each alloc LL will atomically test and decrease relevan counter and will allocate buffer only if that counter was non-zero. The counter will be increased by application core if relevant buffer was freed.

### DIFF
--- a/nimble/controller/src/ble_ll_conn.c
+++ b/nimble/controller/src/ble_ll_conn.c
@@ -43,6 +43,9 @@
 #include "controller/ble_ll_utils.h"
 #include "ble_ll_conn_priv.h"
 #include "ble_ll_ctrl_priv.h"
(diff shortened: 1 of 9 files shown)


andrzej-kaczmarek:
requires also some core changes: https://github.com/apache/mynewt-core/pull/2889

sjanc (approved):

PR #1390, opened 2022-10-06
Fix missing includes and missing return (compiler gives error)

bvdberg:
This fixes the build after the recent HCI_IPC changes.

Also these changes now add a dependency from Mynewt/drivers/ipc5430 to the Nimble stack, which IMHO is a no go. This is not fixes in this request.

### DIFF
--- a/nimble/transport/nrf5340/src/nrf5340_ble_hci.c
+++ b/nimble/transport/nrf5340/src/nrf5340_ble_hci.c
@@ -19,6 +19,7 @@
 
 #include <assert.h>
 #include <string.h>
+#include <syscfg/syscfg.h>
 #include <sysinit/sysinit.h>
 #include <nimble/ble.h>
 #include <ipc_nrf5340/ipc_nrf5340.h>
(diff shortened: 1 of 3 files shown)


nimble/transport/nrf5340/src/nrf5340_ble_hci.c
```diff
@@ -19,11 +19,13 @@
 
 #include <assert.h>
 #include <string.h>
+#include <syscfg/syscfg.h>
 #include <sysinit/sysinit.h>
 #include <nimble/ble.h>
 #include <ipc_nrf5340/ipc_nrf5340.h>
 #include <nimble/transport.h>
 #include <nimble/transport/hci_ipc.h>
+#include <nimble/transport/transport_ipc.h>
```
andrzej-kaczmarek, line 28: this is included from nimble/transport if hci_ipc package is included in build and it should be, because it's added as dependency to nrf5340 transport package - you likely do not have recent newt which generates required defines in syscfg.h

andrzej-kaczmarek:
> Also these changes now add a dependency from Mynewt/drivers/ipc5430 to the Nimble stack, which IMHO is a no go. This is not fixes in this request.

dependency is to add nimble package which only has generic defs, it does not add dependency to host or controller

bvdberg:
We use FreeRTOS and dont use the Newt build sysetm. I have removed the include. The build also works without.

andrzej-kaczmarek (approved):

sjanc (approved):

PR #1392, opened 2022-10-07
nimble/ll: Fix not re-initializing CSS on HCI reset

sjanc:
Reset CSS configuration to defaults on HCI reset. Otherwise we could end up with enabled CSS after HCI reset.


sjanc:
@jrotkiewicz FYI

andrzej-kaczmarek (approved):

PR #1394, opened 2022-10-10
nimble/ll: Fix ifdefs for BLE_LL_EXT

andrzej-kaczmarek:


sjanc (approved):
maybe this should be name BLE_LL_EXTERNAL to avoid confusions with ext adv?

PR #1396, opened 2022-10-10
nimble/host: Add host callback to provide security keys

andrzej-kaczmarek:
Application can use new callback to generate keys for new pairing prior to keys distribution. This allows keys to be derived from some other source (e.g. as in Vol 3, Part H, appendix B) instead of just being randomized.

### DIFF
--- a/nimble/host/include/host/ble_store.h
+++ b/nimble/host/include/host/ble_store.h
@@ -178,6 +178,42 @@ struct ble_store_status_event {
     };
 };
 
+/* Generate LTK, EDIT and Rand */
+#define BLE_STORE_GEN_KEY_LTK       0x01
+/* Generate IRK */
+#define BLE_STORE_GEN_KEY_IRK       0x02
+/* Generate CSRK */
+#define BLE_STORE_GEN_KEY_CSRK      0x03
+
+struct ble_store_gen_key {
+    union {
+        uint8_t ltk_periph[16];
+        uint8_t irk[16];
+        uint8_t csrk[16];
+    };
+    uint16_t ediv;
+    uint64_t rand;
+};
+
+/**
+ * Generates key required by security module.
+ * This can be used to use custom routines to generate keys instead of simply
+ * randomizing them.
+ *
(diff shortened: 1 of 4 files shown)


nimble/host/include/host/ble_store.h
```diff
@@ -178,6 +178,36 @@ struct ble_store_status_event {
...
 
+/* blestore.ltk_periph, ediv and rand are valid */
+#define BLE_STORE_KEYS_MASK_LTK     0x01
+/* ble_store_keys.irk is valid */
+#define BLE_STORE_KEYS_MASK_IRK     0x02
+/* ble_store_keys.csrk is valid */
+#define BLE_STORE_KEYS_MASK_CSRK    0x04
+
+struct ble_store_keys {
+    uint8_t ltk_periph[16];
```
sjanc, line 189: make this union

nimble/host/include/host/ble_store.h
```diff
@@ -178,6 +178,36 @@ struct ble_store_status_event {
...
+ * randomizing them.
+ *
+ * @param keys_mask             Keys that shall be generated.
+ * @param keys                  Storage for generated keys. Only keys indicated
+ *                              by keys mask should be updated.
+ *
+ * @return                      0 if keys were generated successfully
+ *                              Other nonzero on error.
+ */
+typedef int ble_store_gen_keys_fn(uint8_t keys_mask,
```
sjanc, line 208: key instead of keys_mask, since we always ask for single key

nimble/host/src/ble_hs_startup.c
```diff
@@ -372,7 +373,18 @@ ble_hs_startup_go(void)
         return rc;
     }
 
-    ble_hs_pvcy_set_our_irk(NULL);
+    if (ble_hs_cfg.store_gen_keys_cb) {
+        rc = ble_hs_cfg.store_gen_keys_cb(BLE_STORE_KEYS_MASK_IRK, &keys);
```
sjanc, line 377: maybe memset keys before calling callback

sjanc (approved):

PR #1399, opened 2022-10-14
nimble/transport: Add UART transport for external controller

andrzej-kaczmarek:
This is the same as standard "uart" transport except it's for LL side which means it can be used for NimBLE host to talk to external controller. Enable it by "BLE_TRANSPORT_LL: uart_ll".

### DIFF
--- a/nimble/transport/syscfg.yml
+++ b/nimble/transport/syscfg.yml
@@ -46,6 +46,7 @@ syscfg.defs:
             - nrf5340
             - socket
             - apollo3
+            - uart_ll
             - custom
 
     BLE_TRANSPORT_ACL_COUNT:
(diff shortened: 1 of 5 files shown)


nimble/transport/syscfg.yml
```diff
@@ -46,6 +46,7 @@ syscfg.defs:
             - nrf5340
             - socket
             - apollo3
+            - uart_hs
```
jocelynmass, line 49: Shouldn't this value be uart_ll instead?

jocelynmass:
Tested and working! Thanks

AxelLin:
Hi,
Will this PR be merged in next release?

andrzej-kaczmarek:
Yes, although I am thinking of merging this into existing UART H4 transport so we don't duplicate code unnecessarily. But this feature will be available in next release.

AxelLin:
Hi @andrzej-kaczmarek 
Thanks for the prompt reply.
This feature was working in NimBLE-1.4.0, I'm surprised that it was missed in NimBLE-1.5.0. (So please consider it as a bug fix)
What's the target release timeline for next release?

AxelLin:
> Yes, although I am thinking of merging this into existing UART H4 transport so we don't duplicate code unnecessarily. But this feature will be available in next release.

Any update for merging this? @andrzej-kaczmarek

sjanc (approved):

PR #1403, opened 2022-11-02
nimble/host : Added new method of marking an api as deprecated.

IshaESP:
In the current code, using deprecated api's doesn't give any warnings. So, added a new way in which if these api's are used, compiler will give a warning associated with it.


IshaESP:
Hi @sjanc ,
Can you please take a look at this? 
The CI checks are failing as the files where these api's are used are giving warnings which are treated as errors.

sjanc:
we build mynewt with -Werror so this will break build

maybe this should be a macro configurable via syscfg? eg  MYNEWT_VAL(WARN_ON_DEPRECATED) or similar...

IshaESP:
> we build mynewt with -Werror so this will break build
> 
> maybe this should be a macro configurable via syscfg? eg MYNEWT_VAL(WARN_ON_DEPRECATED) or similar...

Added a configurable macro.

nimble/host/include/host/ble_gatt.h
```diff
@@ -476,8 +477,15 @@ int ble_gatts_notify_custom(uint16_t conn_handle, uint16_t att_handle,
 /**
  * Deprecated. Should not be used. Use ble_gatts_notify_custom instead.
  */
+#if MYNEWT_VAL(BLE_SUPPRESS_DEPRECATE_WARN)
```
sjanc, line 480: I'd rather have this as an OPT-IN    MYNEWT_VAL(BLE_WARN_DEPRECATED)
and also define makro (eg in ble_hs.h) for this and not ifdef every single function declaration

#if MYNEWT_VAL(BLE_WARN_DEPRECATED)
#define NIMBLE_DEPRECATED(_msg) __attribute__((deprecated("Deprecated. Should not be used. " _msg))) 
#else
#define NIMBLE_DEPRECATED(_msg)
#endif

and than just do this to deprecate
NIMBLE_DEPRECATED("use bar") void foo(void);

sjanc (changes requested):

nimble/host/include/host/ble_hs.h
```diff
@@ -184,6 +184,22 @@ extern "C" {
...
+ */
+
+/**
+ * @brief LE deprecate warning
+ *
+ * @{
+ */
+
+#if MYNEWT_VAL(BLE_SUPPRESS_DEPRECATE_WARN)
+#define NIMBLE_DEPRECATED(_msg) __attribute__((deprecated("Deprecated. Should not be used. " _msg)))
```
sjanc, line 198: This should be the other way around... but I really think that this feature (warn on deprecate) should be opt in, ie  name it BLE_WARN_ON_DEPRECATE and default to 0 and warn only if it is set to 1

nimble/host/include/host/ble_hs.h
```diff
@@ -184,6 +184,22 @@ extern "C" {
...
+ * @}
+ */
+
+/**
+ * @brief LE deprecate warning
+ *
+ * @{
+ */
+
+#if MYNEWT_VAL(BLE_SUPPRESS_DEPRECATE_WARN)
```
sjanc, line 197: this header should include #include "syscfg/syscfg.h if it is using MYNEWT_VAL() makro

nimble/host/include/host/ble_hs.h
IshaESP: It's default value is set to 0, if someone wants to opt in they can set it to 1.

nimble/host/include/host/ble_gatt.h
IshaESP: For this - NIMBLE_DEPRECATED("use bar") void foo(void);   CI always fails. Tried the other way around too.

nimble/host/include/host/ble_hs.h
IshaESP: Done

nimble/host/include/host/ble_hs.h
sjanc: BLE_SUPPRESS_DEPRECATE_WARN:
        description: >
            This suppresses deprecated warning
        value: 0


Enabling warning should be opt-int, not disabling those  ie:

BLE_WARN_ON_DEPRECATE:
        description: >
            This enabling warning on used deprecated API.
        value: 0

PR #1405, opened 2022-11-07
Compile fix nimble

timoxd7:
## Removes global min/max definition causing problems with other libraries
Some libraries define min/max macros too (like STL) by using templates. With these makros already defined, the compilation might fail or have an unexprected behaviour, especially while using `using namespace std;` together with anything in STL.

## Removes `using namespace std` from header file
Inlcuding `using namespace std` in a header file results in a `using namespace std` be part of any header/cpp file using this header. This might lead to unexpected behaviour and false selection of methods/functions by setting the std namespace to a global scope.

### DIFF
--- a/nimble/host/src/ble_att.c
+++ b/nimble/host/src/ble_att.c
@@ -23,6 +23,15 @@
 #include "host/ble_att.h"
 
 #if NIMBLE_BLE_CONNECT
+
+#ifndef min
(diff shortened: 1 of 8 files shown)


sjanc (approved):
As a temp solution this can go in but in long run I'd rather go with minimizing number those macros definitions.

I think we should just replace those with MIN/MAX and start using those in codebase

PR #1406, opened 2022-11-08
nimble/ll: Fix compile errors

novumdun:
Disallow unused functions when BLE_LL_ROLE_CENTRAL is not set


sjanc:
While I'm fine in general with such patches, I wonder how you setup your compiler to warn about unused public functions?

novumdun:
I port nimble to our system and it's compiled by Keil ac6, not GCC.

sjanc:
alright,  but please fix CI failures (some static functions are left unused if public are ifdeffed)

PR #1408, opened 2022-11-18
nimble/phy/nrf5x: Allow to use named pins for radio debug

sjanc:
This allows to use LED_1 etc for GPIO debug.


andrzej-kaczmarek (changes requested):
I don't really think we should include bsp.h other only to use LED_1 instead of pin number for debug syscfg val...

PR #1410, opened 2022-11-23
nimble/host: Enabled LE Power control event mask if feature is enabled

rahult-github:
If LE Power Control feature is enabled, then host should get relevant events. This PR enables event mask associated with Power control feature.


rahult-github:
@sjanc  , PTAL .

sjanc (approved):

PR #1414, opened 2022-11-30
nimble/ll: Simplify DTM RX test

sjanc:
DTM test uses fixed frequency and PHY mode so there is no need to re-configure those after every packet reception. It is enough to just restart scanning.


andrzej-kaczmarek (commented):
we also don't need to schedule 1st rx, just start it from LL context the same way scan is started.

andrzej-kaczmarek (approved):

PR #1419, opened 2022-12-09
porting/examples: Fix build error on linux

sjanc:
Since 'using namespace std' was removed from public header it is needed to explicitly use it when using std::list


sjanc:
@timoxd7 FYI

andrzej-kaczmarek (approved):

PR #1425, opened 2023-01-04
nimble/ll: Allow to tune time needed for scheduling next conn event

sjanc:
This works by allowing to shorten connection event end time by specified number of us. This value should be tuned only after measuring performance as depends on various factors like build optimisation, cache, clock speed etc.

### DIFF
--- a/nimble/controller/syscfg.yml
+++ b/nimble/controller/syscfg.yml
@@ -537,6 +537,15 @@ syscfg.defs:
         range: 1..257
         value: 32
 
+    BLE_LL_CONN_EVENT_END_MARGIN:
+        description: >
+            Extra time needed for scheduling next connection event. Setting this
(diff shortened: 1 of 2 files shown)


nimble/controller/syscfg.yml
```diff
@@ -537,6 +537,15 @@ syscfg.defs:
         range: 1..257
         value: 32
 
+    BLE_LL_CONN_EVENT_END_MARGIN:
+        description: >
+            Extra time needed for scheduling next connection event. Setting this
+            value results in ending connection event sonner (in us) which gives
```
tannewt, line 543:
```suggestion
            value results in ending connection event sooner (in microseconds) which gives
```

andrzej-kaczmarek (approved):

PR #1428, opened 2023-01-18
Fixed a check in ble_att_clt_tx_prep_write() api

IshaESP:
If application is going to send data more than MTU size using ble_gattc_write_long(), after sending first batch of data of MTU size the offset gets incremented by (MTU - BLE_ATT_PREP_WRITE_CMD_BASE_SZ). This check will always fail in that case, as offset will be equal to (MTU - BLE_ATT_PREP_WRITE_CMD_BASE_SZ), and (offset + OS_MBUF_PKTLEN(txom)) will always be greater than BLE_ATT_ATTR_MAX_LEN. Thus failing to send the whole data.


IshaESP:
Hi @sjanc ,
Can you please take a look?

IshaESP:
Hi @sjanc , can you please take a look at this?

sjanc:
#AutoPTS run mynewt GATT/CL/GAW

codecoup-tester:
Scheduled PR #https://github.com/apache/mynewt-nimble/pull/1428#issuecomment-1410996839 after 10:25:44.

codecoup-tester:
AutoPTS Bot results:
No failed test found.

sjanc:
OK, so I'm not sure about this patch... spec clearly states that maximum attribute value shall be 512 and that data interpretation is responsibility of the higher layer.   And offset is used for reading long attribute value but long here doesn't mean bigger than 512, but bigger than (ATT_MTU -1)..

PR #1432, opened 2023-01-19
Initial Github Action integration

sjanc:
This adds:
 - building all test targets on Linux
 - runing 'newt test all'
 - building dummy and Linux ports


sjanc:
verified locally with act

michal-narajowski (approved):

PR #1436, opened 2023-01-27
nimble/ll: Some updates before ISO BIG

andrzej-kaczmarek:


sjanc (approved):
would be good to later add some unit tests for crypto stuff (h7, h8, etc)

PR #1443, opened 2023-01-30
targets: Restore example targets

sjanc:
Those were accidentally moved to CI targets but are expected to be used by users.


sjanc:
@mkasenberg FYI

andrzej-kaczmarek (approved):

PR #1448, opened 2023-02-02
nimble/ll: Use MIN/MAX macros from ble_ll_utils.h

sjanc:
Use macros provided by utils instead of relying on system providing min and max macros.

### DIFF
--- a/nimble/controller/include/controller/ble_ll_utils.h
+++ b/nimble/controller/include/controller/ble_ll_utils.h
@@ -24,7 +24,7 @@
 #define INT16_LTE(_a, _b) ((int16_t)((_a) - (_b)) <= 0)
 
 #define MIN(_a, _b) ((_a) < (_b) ? (_a) : (_b))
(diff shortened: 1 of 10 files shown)


andrzej-kaczmarek (approved):

AxelLin:
Hi,

There are many min()/max() usage in nimble/host, nimble/transport, etc. (Not just in nimble/controller)
And there are so many duplicate code as :
https://github.com/apache/mynewt-nimble/blob/master/nimble/host/src/ble_att.c#L27-L33
https://github.com/apache/mynewt-nimble/blob/master/nimble/host/src/ble_gap.c#L30-L36
https://github.com/apache/mynewt-nimble/blob/master/porting/nimble/src/os_mbuf.c#L44-L50

Do you consider to put MIN/MAX macros at a common header file and replace all the min()/max() calls?

sjanc:
yes, we will clean it up eventually (this PR was to handle LL code only)

PR #1450, opened 2023-02-02
ci: Build RIOT port

sjanc:
Building against RIOT master branch and thus this is allowed to fail.


.github/workflows/build_ports.yml
```diff
@@ -27,12 +27,25 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/checkout@v3
```
andrzej-kaczmarek, line 30: why need this 2x?

andrzej-kaczmarek (approved):

.github/workflows/build_ports.yml
sjanc: cnp error :)

PR #1451, opened 2023-02-02
nimble/ll: Initial ISO Broadcaster support

andrzej-kaczmarek:
This adds initial implementation for ISO Broadcaster to LL. It's functional, but has some limitations:
- only unframed PDUs are supported
- BIG events will preempt any other event and thus can starve scheduler
- LE Create BIG command is dumb, i.e. does not calculate optimal BIG parameters
- GSK is derived using TinyCrypt, i.e. does not use HW accelerators

Other than the above, any valid BIG configuration should work including encryption.


nimble/controller/src/ble_ll_iso_big.c
```diff
@@ -160,6 +164,58 @@ static uint8_t bis_pool_free = BIS_POOL_SIZE;
...
+int
+ble_ll_iso_big_last_tx_timestamp_get(struct ble_ll_iso_bis *bis,
+                                     uint16_t *packet_seq_num, uint32_t *timestamp)
+{
+    struct ble_ll_iso_big *big;
+
+    big = bis->big;
+
+    *packet_seq_num = big->bis_counter;
+    *timestamp = (uint64_t)big->event_start * 1000000 / 32768 +
```
sjanc, line 213: some comment would be helpful here (unless we have some macro that would cover this)

nimble/controller/src/ble_ll_iso_big.c
```diff
@@ -1159,9 +1159,63 @@ ble_ll_iso_big_hci_evt_complete(void)
...
+    }
 
-    return BLE_ERR_UNSPECIFIED;
+    if (!IN_RANGE(cmd->big_handle, 0x00, 0xef) ||
+        !IN_RANGE(cmd->adv_handle, 0x00, 0xef) ||
+        !IN_RANGE(cmd->num_bis, 0x01, 0x1f) ||
+        !IN_RANGE(get_le24(cmd->sdu_interval), 0x0000ff, 0x0fffff) ||
+        !IN_RANGE(le16toh(cmd->max_sdu), 0x0001, 0x0fff) ||
+        !IN_RANGE(le16toh(cmd->max_transport_latency), 0x0005, 0x0fa0) ||
+        !IN_RANGE(cmd->rtn, 0x00, 0x1e) ||
```
sjanc, line 1176: this may warn about checking unsigned value for >=0 (which is always true)

nimble/controller/src/ble_ll_iso_big.c
```diff
@@ -0,0 +1,1087 @@
...
+    big->control_active = 0;
+    big->control_instant = 0;
+    big->chan_map_new_pending = 0;
+    big->term_pending = 0;
+    big->term_reason = 0;
+
+    /* Calculate number of additional events for pre-transmissions */
+    /* Core 5.3, Vol 6, Part B, 4.4.6.6 */
+    gc = bp->nse / bp->bn;
+    (void)pte;
```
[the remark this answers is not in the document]
andrzej-kaczmarek: it's used by code that will follow, I didn't want to remove calculation from this patchset so I don't need to mess with the code again later

PR #1452, opened 2023-02-03
nimble/ll: Refactor ble_ll_supp_cmd

andrzej-kaczmarek:


nimble/controller/src/ble_ll_hci_supp_cmd.c
```diff
@@ -0,0 +1,328 @@
...
+    octet_43,
+    octet_44,
+    0,
+    octet_46,
+};
+
+void
+ble_ll_hci_supp_cmd_get(uint8_t *buf)
+{
+    memcpy(buf, g_ble_ll_hci_supp_cmds, sizeof(g_ble_ll_hci_supp_cmds));
```
sjanc, line 327: memset of remaining byes is not needed anymore?

andrzej-kaczmarek: g_ble_ll_hci_supp_cmds array is extended to 64 bytes so it's already padded with zeroes

sjanc (approved):

PR #1461, opened 2023-02-10
nimble/ll: Add option to control initial DLE update

sjanc:
New option allows to select if controller should initiate DLE update automatically for connection. By default it is enabled to not change existing controller behavior.


andrzej-kaczmarek (approved):
cfbit in connsm can also be #ifdefd since it's only used by two statememts that are now #ifdefd

PR #1462, opened 2023-02-12
porting/nuttx; Fix error: unrecognized command-line option '-m32'

xiaoxiang781216:
-m32 is supported by x64 toolchain, not arm-none-eabi-gcc.
Report here: https://github.com/apache/nuttx-apps/actions/runs/4131763426/jobs/7139771820

### DIFF
--- a/porting/examples/nuttx/Make.defs
+++ b/porting/examples/nuttx/Make.defs
@@ -48,8 +48,7 @@ INC = \
 
 INCLUDES := $(addprefix -I, $(INC))
 
-CFLAGS +=                    \
-    $(NIMBLE_CFLAGS)        \
+CFLAGS +=                   \
     $(INCLUDES)             \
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@sjanc could you review this patch?

xiaoxiang781216:
ping @sjanc again.

sjanc (approved):

sjanc:
note that if you also support x86 you may need to use -m32 since nimble was not designed for 64bit (although host side may work, but should probably be reviewed if there is any code that assumes 32bit pointers),  for controller side 64bit is not supported

xiaoxiang781216:
> note that if you also support x86 you may need to use -m32 since nimble was not designed for 64bit (although host side may work, but should probably be reviewed if there is any code that assumes 32bit pointers), for controller side 64bit is not supported

Sure.

PR #1469, opened 2023-02-28
nimble/host: Add connection handle related GAP APIs

m-gorecki:
This adds two new GAP APIs:

void ble_gap_conn_foreach_handle(ble_gap_conn_foreach_handle_fn &ast;cb, void &ast;arg)
API can be used to call function declared by the user for every connection that is currently established. As arguments, it takes a pointer to the function, that will be called and a void pointer, which is used to pass the optional argument of any type to the called function. Function from passed pointer should return an int value and take two arguments - uint16_t for conn_handle and void&ast; for the optional argument (see ble_gap_conn_foreach_handle_fn typedef).

int ble_gap_conn_func_handle_by_addr(const ble_addr_t &ast;addr, uint16_t &ast;out_conn_handle)
API will look for the connection with address specified in the first argument and if it succeds, it will write the connection handle to the variable pointed by out_conn_handle. Returns 0 on success or BLE_HS_ENOTCONN if connection with specified address was not found.

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -509,6 +509,57 @@ ble_gap_conn_find_by_addr(const ble_addr_t *addr,
 #endif
 }
 
+int
+ble_gap_conn_find_handle_by_addr(const ble_addr_t *addr, uint16_t *out_conn_handle)
+{
+#if NIMBLE_BLE_CONNECT
+    struct ble_hs_conn *conn;
+
+    ble_hs_lock();
+
+    conn = ble_hs_conn_find_by_addr(addr);
+    if (conn != NULL) {
+        *out_conn_handle = conn->bhc_handle;
+    } else {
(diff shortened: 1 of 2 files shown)


nimble/host/src/ble_gap.c
```diff
@@ -509,6 +509,54 @@ ble_gap_conn_find_by_addr(const ble_addr_t *addr,
...
+ble_gap_conn_find_handle_by_addr(const ble_addr_t *addr, uint16_t *out_conn_handle)
+{
+#if NIMBLE_BLE_CONNECT
+    struct ble_hs_conn *conn;
+
+    ble_hs_lock();
+
+    conn = ble_hs_conn_find_by_addr(addr);
+
+    *out_conn_handle = conn->bhc_handle;
```
andrzej-kaczmarek, line 522: need to check if conn!=NULL first

andrzej-kaczmarek (changes requested):

andrzej-kaczmarek (approved):

PR #1470, opened 2023-02-28
nimble/host: Add peer id addr to identity resolved event

andrzej-kaczmarek:
This way app does not need to query for new ID address.


rymanluk (approved):
Looks good to me

sjanc (approved):

PR #1476, opened 2023-03-09
apps: Add lvgl scanner

m-gorecki:
Application performs extended scan and displays the results on the Adafruit 2.8" TFT LCD display.


apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+{
+    console_printf("Resetting state; reason=%d\n", reason);
+}
+
+static void
+btn_event_cb(lv_event_t *e)
+{
+    static uint8_t scanning = 0;
+    lv_event_code_t code = lv_event_get_code(e);
+    lv_obj_t * btn = lv_event_get_target(e);
```
andrzej-kaczmarek, line 250: `lv_obj_t *btn`

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+
+static void
+start_scan(void)
+{
+    struct ble_gap_ext_disc_params coded;
+    struct ble_gap_ext_disc_params uncoded;
+    uint8_t filter_duplicates = 0;
+    uint8_t filter_policy = 0;
+    uint8_t limited = 0;
+    coded.passive = 1;
```
andrzej-kaczmarek, line 207: use active scanning, some devices may have local name in scan_rsp only

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+    coded.itvl = 500;
```
andrzej-kaczmarek, line 208: use `BLE_GAP_SCAN_ITVL_MS` and `BLE_GAP_SCAN_WIN_MS` macros and configure for 100% duty cycle

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+#include "lvgl.h"
+
+#define MAX_ENTRIES         30
+#define ENTRY_ID_NONE       (MAX_ENTRIES + 1)
+
+lv_obj_t *entries_list;
+
+struct list_entry_data {
+    int8_t last_rssi;
+    uint8_t addr_val[6];
```
andrzej-kaczmarek, line 34: `ble_addr_t`

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+}
+
+static int8_t
+get_entry_id_by_addr(uint8_t *addr)
+{
+    uint8_t n = lv_obj_get_child_cnt(entries_list);
+
+    assert (n <= MAX_ENTRIES);
+
+    for (uint8_t id = 0; id < n; id++) {
```
andrzej-kaczmarek, line 70: `id` should be a local variable

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+        if (!memcmp(addr, entry_data[id].addr_val, 6)) {
+            return id;
+        }
+    }
+    return ENTRY_ID_NONE;
```
andrzej-kaczmarek, line 75: better use `int` as return value type and simply return -1

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+    uint8_t n = lv_obj_get_child_cnt(entries_list);
+
+    for (i = 0; i < n - 1; i++) {
+        for (j = 0; j < n - i - 1; j++) {
+            rssi1 = entry_data[j].last_rssi;
+            rssi2 = entry_data[j + 1].last_rssi;
+
+            entry = lv_obj_get_child(entries_list, j + 1);
+            if (rssi1 < rssi2) {
+                struct list_entry_data help = entry_data[j + 1];
```
andrzej-kaczmarek, line 54: local variable

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+    uint8_t i;
+    uint8_t j;
+    uint8_t n = lv_obj_get_child_cnt(entries_list);
+
+    for (i = 0; i < n - 1; i++) {
+        for (j = 0; j < n - i - 1; j++) {
+            rssi1 = entry_data[j].last_rssi;
+            rssi2 = entry_data[j + 1].last_rssi;
+
+            entry = lv_obj_get_child(entries_list, j + 1);
```
andrzej-kaczmarek, line 52: this can be called inside if below

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+{
+    lv_obj_t *entry;
+    int8_t rssi1;
+    int8_t rssi2;
+    uint8_t i;
+    uint8_t j;
+    uint8_t n = lv_obj_get_child_cnt(entries_list);
+
+    for (i = 0; i < n - 1; i++) {
+        for (j = 0; j < n - i - 1; j++) {
```
andrzej-kaczmarek, line 48: use `qsort()`

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+            rssi1 = entry_data[j].last_rssi;
+            rssi2 = entry_data[j + 1].last_rssi;
+
+            entry = lv_obj_get_child(entries_list, j + 1);
+            if (rssi1 < rssi2) {
+                struct list_entry_data help = entry_data[j + 1];
+                entry_data[j + 1] = entry_data[j];
+                entry_data[j] = help;
+                lv_obj_move_to_index(entry, j);
```
andrzej-kaczmarek, line 57: you can add pointer to object to `entry_data` struct, then simply sort entries and move objects to appropriate positions after sorting is done. this way you don't need to move them multiple times during sorting

andrzej-kaczmarek (changes requested):

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,302 @@
...
+ * be updated */
+static void
+fill_entry_labels(lv_obj_t *entry, struct ble_gap_ext_disc_desc *desc, uint8_t is_new)
+{
+    lv_obj_t *label;
+
+    if (is_new) {
+        label = lv_label_create(entry);
+
+        lv_label_set_text_fmt(label, "%02x:%02x:%02x:%02x:%02x:%02x", desc->addr.val[5],
```
andrzej-kaczmarek, line 89: also need address type

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+    ble_addr_t addr;
+    lv_obj_t *list_entry;
+};
+
+static struct device_data devices[MAX_DEVICES];
+
+int
+rssi_compare(const void *p1, const void *p2)
+{
+    struct device_data *d1 = (struct device_data *) p1;
```
andrzej-kaczmarek, line 44: do not use cast to remove const
`const struct device_data *d1 = p1;`

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+#include "console/console.h"
+#include "host/ble_hs.h"
+#include "log/log.h"
+#include "lvgl.h"
+
+#define MAX_DEVICES         30
+#define DEVICE_ID_NONE      -1
+
+lv_style_t style;
+lv_obj_t *entries_list;
```
andrzej-kaczmarek, line 31: both should be static

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+
+#include "sysinit/sysinit.h"
+#include "os/os.h"
+#include "console/console.h"
+#include "host/ble_hs.h"
+#include "log/log.h"
+#include "lvgl.h"
+
+#define MAX_DEVICES         30
+#define DEVICE_ID_NONE      -1
```
andrzej-kaczmarek, line 28: no need to define this, just return -1 in code and then compare <0 where applicable

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+ */
+
+#include "sysinit/sysinit.h"
+#include "os/os.h"
+#include "console/console.h"
+#include "host/ble_hs.h"
+#include "log/log.h"
+#include "lvgl.h"
+
+#define MAX_DEVICES         30
```
andrzej-kaczmarek, line 27: make this configurable via syscfg

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+    lv_obj_align(entries_list, LV_ALIGN_TOP_MID, 0, 5);
+    lv_obj_set_flex_flow(entries_list, LV_FLEX_FLOW_COLUMN);
+}
+
+static void
+handle_disc_report(struct ble_gap_ext_disc_desc *desc)
+{
+    int id;
+
+    if (desc->rssi < -70 || !is_event_connectable(desc)) {
```
andrzej-kaczmarek, line 185: rssi threshold should be configurable via syscfg or preferable via ui

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+        return;
+    }
+    if (lv_obj_get_child_cnt(entries_list) >= MAX_DEVICES) {
+        lv_obj_del(entries_list);
+        list_init();
```
andrzej-kaczmarek, line 190: not sure if I understand correctly, but does this code just recreate list from scratch if max number of devices are scanned?

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+    return DEVICE_ID_NONE;
+}
+
+/* This fills entry labels which are going to be displayed. Takes is_new flag as argument.
+ * If flag is set, then new labels will be created, otherwise existing labels will
+ * be updated */
+static void
+fill_device_labels(struct ble_gap_ext_disc_desc *desc, uint8_t id, uint8_t is_new)
```
andrzej-kaczmarek, line 82: probably it makes more sense to have separate function to initialize empty object for new device (i.e. `add_new_device`) and then call update all labels which can be used for both existing device and the one that we just created (i.e. `update_device`)

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+
+static int
+get_device_id_by_addr(uint8_t *addr)
+{
+    uint8_t id;
+    uint8_t n = lv_obj_get_child_cnt(entries_list);
+    assert (n <= MAX_DEVICES);
+
+    for (id = 0; id < n; id++) {
+        if (!memcmp(addr, devices[id].addr.val, 6)) {
```
andrzej-kaczmarek, line 71: need to also include address type, use `ble_addr_cmp`

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+/* This fills entry labels which are going to be displayed. Takes is_new flag as argument.
+ * If flag is set, then new labels will be created, otherwise existing labels will
+ * be updated */
+static void
+fill_device_labels(struct ble_gap_ext_disc_desc *desc, uint8_t id, uint8_t is_new)
+{
+    lv_obj_t *entry = devices[id].list_entry;
+    lv_obj_t *label;
+    struct ble_hs_adv_fields fields;
+    char name_buf[30];
```
andrzej-kaczmarek, line 87: name can be up to 255 chars

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+scan_event(struct ble_gap_event *event, void *arg)
+{
+    switch (event->type) {
+    /* advertising report has been received during discovery procedure */
+    case BLE_GAP_EVENT_EXT_DISC:
+        handle_disc_report(&event->ext_disc);
+        return 0;
+
+    default:
+        MODLOG_DFLT(ERROR, "Discovery event not handled\n");
```
andrzej-kaczmarek, line 213: just ignore without message

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+
+static void
+start_scan(void)
+{
+    struct ble_gap_ext_disc_params coded;
+    struct ble_gap_ext_disc_params uncoded;
+    uint8_t filter_duplicates = 0;
+    uint8_t filter_policy = 0;
+    uint8_t limited = 0;
+    coded.passive = 0;
```
andrzej-kaczmarek, line 226: empty line between declarations and code

apps/lvgl_scanner/src/main.c
```diff
@@ -0,0 +1,333 @@
...
+}
+
+static void
+start_scan(void)
+{
+    struct ble_gap_ext_disc_params coded;
+    struct ble_gap_ext_disc_params uncoded;
+    uint8_t filter_duplicates = 0;
+    uint8_t filter_policy = 0;
+    uint8_t limited = 0;
```
andrzej-kaczmarek, line 225: not need to define variables which are in this case only used as parameter to function

andrzej-kaczmarek (changes requested):

PR #1480, opened 2023-03-31
nimble/host: Expose the API to enable / disable privacy outside of stack.

rahult-github:
This PR exposes the API to enable/ disable controller privacy outside stack. This way applications can enable / disable controller privacy as per use case.

### DIFF
--- a/nimble/host/include/ble_hs_pvcy.h
+++ b/nimble/host/include/ble_hs_pvcy.h
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


rahult-github:
Hi @andrzej-kaczmarek  ,  

The use case is that controller should not resolve a resolvable address and end user should get this option to control . Disabling address resolution in controller , eventually implies Link layer will not resolve the private address using the resolving list's peer IRK, thereby bypassing privacy. Please let me know if the commit message is confusing and hence can modify the commit message itself.  

Please let us know if you have any other suggestion to achieve this , would be happy to modify the commit accordingly .

andrzej-kaczmarek:
but why would user want that? what is the specific use case? "bypassing privacy" sounds like a hack to me...

rahult-github:
Hi @andrzej-kaczmarek ,

As per our understanding, address resolution ( and hence privacy) is NOT a mandatory/ must have feature. End user can have the option to not enable it by default in its implementation.  We can have it enabled by default in nimble, but providing an option to user to disable it for its use case would give the nimble stack more flexibility.

So basically, applications that don't need controller to do address resolution can disable it and then enable back when needed. Controller address resolution is good, but hosts providing support for host based privacy ( i.e. address resolution in host) , would need address resolution disabled in controller. 

Having said above, if you have any other comment on how to achieve this, please do let me know.

andrzej-kaczmarek:
NimBLE host does not support host-based privacy. If you need to use it with controller that does not support LL Privacy, you need to add support for that.

Exposing an internal API to simply toggle address resolution makes sense only if there's a specific case where this could be useful. Disabling LL Privacy is not such case simply because disabling address resolution does only disable address resolution, not an entire LL Privacy feature.

PR #1487, opened 2023-04-20
nimble/ll: Allow to locally update connection min/max CE length

sjanc:
Updating Min/Max CE Length can be done without triggering LL procedure. We can do that if current connection parameters fits parameters parameters provided in HCI command. Note that this is also valid for Connections Strict Scheduling.

### DIFF
--- a/nimble/controller/include/controller/ble_ll_hci.h
+++ b/nimble/controller/include/controller/ble_ll_hci.h
@@ -38,7 +38,7 @@ extern "C" {
  */
 #define BLE_LL_CFG_NUM_HCI_CMD_PKTS     (1)
 
-typedef void (*ble_ll_hci_post_cmd_complete_cb)(void);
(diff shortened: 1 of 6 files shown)


nimble/controller/include/controller/ble_ll_hci.h
```diff
@@ -81,6 +81,9 @@ int ble_ll_hci_check_dle(uint16_t max_octets, uint16_t max_time);
 
 void ble_ll_hci_supp_cmd_get(uint8_t *buf);
 
+/* Used to set post HCI command hook */
+void ble_ll_hci_set_post_cmd_cb(ble_ll_hci_post_cmd_complete_cb cb, void *user_data);
```
andrzej-kaczmarek, line 85: `ble_ll_hci_post_cmd_cb_set`

andrzej-kaczmarek (approved):

PR #1494, opened 2023-05-08
mynewt: fix assert

bvdberg:
The assert only works on MyNewt since the ev field does not exist in other versions of ble_npl_event


bvdberg:
renamed commit

sjanc (commented):
Actually,  there is ble_npl_event_is_queued() which should be used instead

bvdberg:
does that also work on MyNewt?
so 
    BLE_LL_ASSERT(!ble_npl_event_is_queued(&sm->sync_ev_end));
or put in #else of #if MYNEWT
?

bvdberg:
changed so your suggestion

PR #1497, opened 2023-05-10
porting/FreeRTOS: fix IRQ priority

bvdberg:
FreeRTOS assumes a max irq priority for IRQs that are allowed to do system calls. This is set to 5 by default, so we use the highest allowed for the Radio.


bvdberg:
Sure, changed as you suggested.

sjanc (approved):

andrzej-kaczmarek (approved):

PR #1504, opened 2023-05-23
nimble: Added support of encrypted advertising data

IshaESP:


sjanc:
Hi,

This requires some explanation of design, functionality etc in commit message (and preferably split into multiple functional commits).  Otherwise it might be hard to review 1.2k lines of code...

Also, while NimBLE supports ports our main platform is Mynewt and thus you should at least validate compilation and testsuite passing (newt test) when upstreaming code

IshaESP:
Hi,
Pipeline has passed.
Regarding different commits, it's very hard to split the code into multiple commits as everything is dependent on each other.

Description:

- Added key material characteristic. This characteristic returns the session key and IV when read. This session key and IV are important for client to decrypt the advertising data.
- Similarly, on server side, encryption of advertising data is done using session key and IV.
- A new structure `enc_adv_data` is being added to store encrypted advertising data.
- If an advertising packet has code 0x31, it means it contains encrypted advertising data. (Macro used : `BLE_GAP_ENC_ADV_DATA`)
- The file `ble_ead.h` has API's for encrypting and decrypting data, ble_ead_encrypt and ble_ead_decrypt respectfully. These API's are called by application layer.
- The file `ble_aes_ccm.h` is a helper file for `ble_ead.h`. It actually calls the mbedtls API's for encrypting/decrypting data.
-  New structure `ble_store_value_ead` has beein added in ble_store_config files to store the ead data across reboots.

PR #1510, opened 2023-05-31
nimble: Allow to configure BLE version 5.4

KKopyscinski:
Adds configuration for Core Specification 5.4


andrzej-kaczmarek:
why 5.3 is not enough for bttester? afaik we do not support any 5.4 feature.

KKopyscinski:
@andrzej-kaczmarek I just set the highest possible, but it makes sense to set at least 5.3, because for example periodic advertising belongs there

andrzej-kaczmarek:
I'd stick to 5.3 to make sure we are compliant, then check if 5.4 is just a matter of bumping version or there are some erratas that we are missing for 5.4 compliance.

andrzej-kaczmarek (approved):

sjanc (approved):

PR #1525, opened 2023-06-13
nimble/ll: Fix connection update instant calculation

andrzej-kaczmarek:
The instant for connection update was calculated at the time PDU was
enqueued in connsm. This caused new conn params to be always applied at
instant regardless if PDU was event dequeued for tx, e.g. in case there
were lots of data PDUs to be send before control PDU.

Currently we calculate instant when PDU is dequeued to make sure this is
the next PDU to be sent and thus instant is valid.


andrzej-kaczmarek:
@jrotkiewicz

jrotkiewicz:
Tested, works ; )

PR #1526, opened 2023-06-14
nimble: Fixed incrementing tx credits in l2cap coc layer after a packet has been sent.

IshaESP:
In function `ble_l2cap_coc_continue_tx` after sending a packet `tx->credits` was not getting incremented. Due to this, l2cap was not able to send more packets than initial credits. 
For example, after sending one packet of 512 bytes, l2cap tx credits became 0 which didn't allow sending any more packet.
Fixed it by keeping a count `num_tx_credits` which will keep track of credits used for a packet and after that packet has been sent `tx->credit` will be incremented by `num_tx_credits`.

### DIFF
--- a/nimble/host/src/ble_l2cap_coc.c
+++ b/nimble/host/src/ble_l2cap_coc.c
@@ -443,6 +443,7 @@ ble_l2cap_coc_continue_tx(struct ble_l2cap_chan *chan)
     struct ble_hs_conn *conn;
     uint16_t sdu_size_offset;
     int rc;
+    uint16_t num_tx_credits = 0;
 
     /* If there is no data to send, just return success */
     tx = &chan->coc_tx;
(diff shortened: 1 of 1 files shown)


IshaESP:
Hi @sjanc, can you please take a look.

andrzej-kaczmarek:
perhaps I'm missing something here, but basically your fix will return credits *locally* immediately after l2cap sdu was transmitted. this kind of defeats the purpose of coc where credits should be returned by peer once it can receive more data.

IshaESP:
So ble_l2cap_chan has two separate variables for handling tx and rx credits, coc_tx and coc_rx respectively. As you said, "credits should be returned by peer once it can receive more data" is taken care by coc_rx when it's ready to receive more data. It gets assigned initial credits in ble_l2cap_coc_recv_ready once the peer is ready. In this PR, credits of coc_tx is getting incremented for sending more data once one SDU is completely sent.

PR #1530, opened 2023-06-19
host/l2cap: add doxygen comments for the header file

wpiet:


nimble/host/include/host/ble_l2cap.h
```diff
@@ -234,33 +470,166 @@ struct ble_l2cap_event {
...
+ * @brief Send an SDU (Service Data Unit) over an L2CAP channel.
+ *
+ * This function sends an SDU over the specified L2CAP channel. The SDU is encapsulated
+ * in L2CAP frames and transmitted to the remote device.
+ *
+ * @param chan          Pointer to the L2CAP channel structure representing the channel to send the SDU on.
+ * @param sdu_tx        Pointer to the os_mbuf structure containing the SDU (Service Data Unit) to send.
+ *
+ * @return              0 on success;
+ *                      BLE_HS_ESTALLED: if the transmit buffer contained an SDU that has not been entirely sent,
```
sjanc, line 587: this is not correct,    host will return ESTALLED when there was not enough credits available to send whole SDU,  in such case application needs to wait for BLE_L2CAP_EVENT_COC_TX_UNSTALLED event before being able to transmit more data


(and BLE_L2CAP_COC_FLAG_STALLED is implementation internal detail)

PR #1531, opened 2023-06-20
nimble/phy/cmac: Move chip variant detection to core

andrzej-kaczmarek:
Chip variant detection does not work properly in ble_rf_init since it may be called when PD_SYS is disabled. We better do this in core on CMAC startup as PD_SYS is always enabled at that time.


andrzej-kaczmarek:
requires https://github.com/apache/mynewt-core/pull/3026

sjanc (approved):

PR #1539, opened 2023-06-29
   nimble/host: Set length correctly in perdioic_set_adv_data

rahult-github:
This change sets length field only if data is non NULL, else sets  length to zero.
   
   Core 5.4, Vol 4, Part E, section 7.8.62 states 
   
  "If Operation is not 0x03 or 0x04 and Advertising_Data_Length is zero, then the Controller shall return the error code Invalid HCI Command Parameters (0x12).."

   Current code sets length to size of data. However, to validate above, need to send 0 length . So calculate length only if data is present.


sjanc (changes requested):
currently ble_gap_periodic_adv_set() assumes that data is always valid and I'm not sure on how this patch would work since *data is dereferenced mutliple times later on in the function

rahult-github:
> currently ble_gap_periodic_adv_set() assumes that data is always valid and I'm not sure on how this patch would work since *data is dereferenced mutliple times later on in the function

Thanks for pointing this out. I have updated the patch to copy data only if len is valid . ( in the cases where length can be zero )

rahult-github:
Hi @sjanc  , please let me know if there are any more comments.

sjanc (approved):

PR #1540, opened 2023-06-30
host/dtm: add doxygen comments for the header file

wpiet:


nimble/host/include/host/ble_dtm.h
```diff
@@ -25,25 +25,120 @@
...
 
+/**
+ * @brief Start a Direct Test Mode (DTM) receiver test.
+ *
+ * This function starts a DTM RX test with the provided parameters.
+ *
+ * @param params                The parameters for the DTM RX test.
+ *
+ * @return                      0 on success;
+ *                              A Bluetooth Host Error Code on failure:
```
sjanc, line 68: looks good but I don't think we need to comment on every possible error code here,  just note that "0 on success, nonzero on failure."  (otherwise this will get out of sync eventually)

sjanc (approved):

PR #1548, opened 2023-07-11
 nimble/host: Added return type handling for npl_mutex_init

rahult-github:
npl_mutex_init may return a non-zero value in case of failure. At certain places, the check is missing to capture the return value. Added the same.


rahult-github:
@andrzej-kaczmarek   , please take a look

sjanc (approved):

PR #773, opened 2020-03-12
nimble: [WIP] Eatt initial implementation

rymanluk:
This is initial implementation and not yet completed. Idea is to hide EATT from the application, but instead allow to Host to choose bearer to use. Code pass some PTS test cases related to EATT and new opcodes.

Posted for early comments: TODO list
* number of TODOs in the code (error paths)
* handling read/write Client/Server Supported Features  - now it is done in EATT code but should be generic
* cache client/server supported features values for bonded devices
* add/fix unit tests
* how we want to expose to app sending multiple notifications.
* expose test API to choose different channels to send data (required by PTS)

PR #1549, opened 2023-07-13
nimble: Eatt initial implementation

KKopyscinski:
This is rebased https://github.com/apache/mynewt-nimble/pull/773

Before merging this needs to be checked against specification and tested


KKopyscinski:
TODO: it may be worth checking `client_att_busy` flag reset after disconnection. Because `ble_gatt_conn_test_disconnect` tests only GATT API for broken connection (connection instance still exists) it probably should be done as separate test, maybe in GAP.

nimble/host/src/ble_hs.c
```diff
@@ -734,6 +734,9 @@ ble_hs_init(void)
     rc = ble_l2cap_init();
     SYSINIT_PANIC_ASSERT(rc == 0);
 
+    rc = ble_gap_init();
```
rymanluk, line 737: we need to do something with line 752 I guess

KKopyscinski: Also, `ble_gap_init` should be outside `#if NIMBLE_BLE_CONNECT` - it includes code for features without connection

KKopyscinski: fixed

KKopyscinski:
Found bug in Read Multiple Variable Length handling: implementation is OK, but btshell must be changed.
Code seems copied from other read/write callbacks, but here format is different. Response is single message with multiple tuples of {length, value} pairs, that is up to MTU in size. If more should be returned, it's trimmed: so we don't use `BLE_HS_EDONE` to mark final message, we will receive only one, with success code of 0. `struct ble_gatt_attr *attr` points to the first attreibute in list, and `num_handles` must be used to access the rest

KKopyscinski:
This was tested against PTS and implemented features are passing. After rebase some commits were added and some modified. Changes between rebase and current version can be seen here: https://gist.github.com/KKopyscinski/5ed12fe57bb457f21a8d98d2a84e0406
rebased version without these changes is here https://github.com/KKopyscinski/mynewt-nimble/tree/eatt_initial_rebase

TODO: Add support for sending Multiple Handle Variable Length Notification.

rymanluk:
Looks like it was merged accidentally but if all is working then it is fine. I was about to approve it anyways.

PR #1559, opened 2023-07-20
porting/npl/linux: Replace pthread_yield() with sched_yield()

rojer:
Per `pthread_yield(3)`:
```
  This call is nonstandard, but present on several other systems.
  Use the standardized sched_yield(2) instead.
```


sjanc (commented):
code looks ok, but could you prefix commit message with "porting: " (to keep convention used in project))

andrzej-kaczmarek (approved):

PR #1562, opened 2023-07-20
Implement ble_hs_locked_by_cur_task() on Linux

rojer:
Otherwise BLE_HS_DEBUG does not work


rojer:
it's been a while. ping?

PR #1563, opened 2023-07-20
Make mempool 64-bit compatible

rojer:


rojer:
Note: This is by no means complete 64-bit support, for example tracing API still assumes pointers are 32 bit, but with this (as well as [other](https://github.com/apache/mynewt-nimble/pull/1559) [related](https://github.com/apache/mynewt-nimble/pull/1560) [changes](https://github.com/apache/mynewt-nimble/pull/1562) i've posted) i have Linux host + kernel HCI building and working with clang and address sanitizer enabled (which is very useful in itself).

andrzej-kaczmarek:
we recently pushed PRs that fix mynewt code to work on x64, see here for mempool changes: https://github.com/apache/mynewt-core/pull/3042

I suggest that you add your comments there in case something else has to be fixed as we'd prefer to fix in core then copy to nimble directly instead of fixing both separately.

rojer:
ok, great, feel free to close this one then. but please do take a look at others, they are fairly minor.

PR #1580, opened 2023-08-01
host/ble_att_svr.c: do not read own characteristics without read perm…

KKopyscinski:
…ission

Characteristics without read permission most likely do not expect being read by their access callback. This especially relates to control point characteristics, that do not hold any particular value, but are used to access procedures, like GATT_PROXY_DATA_IN characteristic in Mesh. Reading them will lead to triggering procedures resulting with undefined behavior.


KKopyscinski:
So this breaks unittests as it is expected that permission is not checked for internal access by att server to it's own characteristics. Imo either unittest expectation should be changed (but not checking permission of own database makes sense) or other method of verifying what operation is supported on given attribute  must be developed.

PR #1586, opened 2023-08-08
nimble/host: Add connection handle realated GAP APIs to header file

m-gorecki:
This adds the declarations of two GAP APIs that were previously defined in the ble_gap.c. Now both APIs can be used in other source files.


KKopyscinski (approved):
LGTM, I think that was missing from https://github.com/apache/mynewt-nimble/pull/1469. Is that correct @andrzej-kaczmarek ?

andrzej-kaczmarek (approved):

PR #1591, opened 2023-08-11
porting/npl/freertos: Add way of including HW specific header

pfyra:
As mentioned in the in_isr function, there is HW specific code in this file but there is no way to include a header file that defines the SCB etc. This PR adds a way to include a header file so it is possible to make this file compile without local changes.

### DIFF
--- a/porting/npl/freertos/src/npl_os_freertos.c
+++ b/porting/npl/freertos/src/npl_os_freertos.c
@@ -22,6 +22,7 @@
 #include <string.h>
 #include "nimble/nimble_npl.h"
 
+/* Include the file that defines the SCB for your HW. */
 #ifdef NIMBLE_NPL_OS_EXTRA_INCLUDE
 #include NIMBLE_NPL_OS_EXTRA_INCLUDE
 #endif
--- a/porting/npl/freertos/src/npl_os_freertos.c
+++ b/porting/npl/freertos/src/npl_os_freertos.c
@@ -22,6 +22,10 @@
 #include <string.h>
 #include "nimble/nimble_npl.h"
 
+#ifdef NIMBLE_NPL_OS_EXTRA_INCLUDE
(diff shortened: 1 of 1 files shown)


porting/npl/freertos/src/npl_os_freertos.c
```diff
@@ -22,6 +22,10 @@
 #include <string.h>
 #include "nimble/nimble_npl.h"
 
+#ifdef NIMBLE_NPL_OS_EXTRA_INCLUDE
+#include NIMBLE_NPL_OS_EXTRA_INCLUDE
+#endif
+
```
sjanc, line 28: since in_isr() is HW specific anyway,  why not just include required file here explicitly?

pfyra: I suppose all ARM MCUs have the SCB so it should be enough to include a header file for the HW specific stuff. At least in my case (stm32) it is. I suppose everyone who use this code needs to add an include statement to this file (or modify it further).
It is a bit cumbersome to maintain changes to the nimble repo at our end since we want to use the nimble repo as a submodule and it is very close to work without changes. Having a possibility to specify a file to include here makes it possible for us to use nimble without any changes.

Perhaps there is a CMSIS file that can be included that specifies the SCB? That might introduce some unwanted dependencies though...

sjanc: fair enough, but could you put some comment above this include? explaining why it is needed etc

pfyra: Certainly. I have added a comment now and I tried to keep it short but descriptive. To me, it is borderline obvious that you use this to avoid changes in the code and if other changes are required anyway then I guess people simply won't bother with this. Let me know what you think and I'll update.

sjanc (approved):

PR #1599, opened 2023-08-18
host: Add support for LE Set Data Related Address change command

rahult-github:
Add API to send LE Set Data related Address change HCI command


rahult-github:
@sjanc  , please take a look

sjanc:
Hi,

Could you rebase so that CI tests kick in?

rahult-github:
> Hi,
> 
> Could you rebase so that CI tests kick in?

Yes . done now.

rahult-github:
@sjanc  , can you please take a look as have rebased the PR.

sjanc:
Hi,

Why not put this configuration in advertising parameters and handle it there? Instead of adding dedicated API that just wraps HCI command

rahult-github:
> Hi,
> 
> Why not put this configuration in advertising parameters and handle it there? Instead of adding dedicated API that just wraps HCI command

Thanks for the review. There is no harm in setting the reasons as part of advertising parameters and issuing this command. 

But spec says "This command may be used while advertising is enabled".  This gives a thought , that this HCI command can be invoked later on also with a reason code as per use case which may not be provided when advertising was enabled? A standalone usage of sending the HCI command is possible. That is why created a separate API.  But please let me know if this understanding of mine is incorrect.

nimble/host/include/host/ble_gap.h
```diff
@@ -2682,6 +2682,17 @@ int ble_gap_set_path_loss_reporting_param(uint16_t conn_handle, uint8_t high_thr
...
+
+/**
+ * Set Data Related Address Changes Param
+ *
+ * @param adv_handle        Advertising handle
+ * @param change_reason     Reasons for refreshing addresses
+ *
+ * @return                  0 on success; nonzero on failure.
+ */
+int ble_gap_set_data_related_addr_change_param(uint8_t adv_handle, uint8_t change_reason);
```
sjanc, line 2694: since behavior is different if legacy commands are used I'd split this into two separate APIs and make parameters name follow other adv api conventions:
```
struct ble_gap_adv_addr_change_params {
      unsigned int on_adv_data_change : 1;
      unsigned int on_scan_rsp_change : 1;
}

int ble_gap_ext_adv_set_data_related_addr_change(uint8_t instance,
                                                 const struct ble_gap_adv_addr_change_param *params);

int ble_gap_adv_set_data_related_addr_change(const struct ble_gap_adv_addr_change_param *params);
```

and put them under proper ifdef sections in headers and source code.
Legacy variant can skipped as it is unlikely to be used anyway with legacy HCI.

sjanc:
Hi,

sorry for late reply,  fair enough, so this can be separated API but I'd make some changes to it to follow NimBLE API convention more closely (and bit more future proof)

PR #1603, opened 2023-08-31
apps/bttester: disable EATT in default config

KKopyscinski:
EATT causes issues with tests GAP/SEC/SEM/BV-{65-67}-C. It is experimental feature and should not influence config for current test plan.


KKopyscinski:
FYI @mkasenberg

rymanluk (approved):
But we need to add it back soon :)

PR #1607, opened 2023-09-10
porting/linux: fix deadlock in ble_npl_hw_enter_critical

Fanhang32:
Fix wrong mutex attributes in pthread_mutexattr_settype. It can cause deadlock after locked.


Fanhang32:
side effect of this change https://github.com/apache/mynewt-nimble/pull/1031 @sjanc

sjanc:
@gavin-yy FYI

sjanc (approved):

PR #1611, opened 2023-09-15
host/ble_gap.c : Check allowed random address

rahult-github:
Add a check to validate whether the random address being set is allowed as per spec


rahult-github:
Hi @sjanc , can you please review this change.

sjanc:
I think this should be in some common code eg ble_hs_id.c ?  Some helpers for generating and internal helpers for validating type of addresses.
There are similar checks for static random when calling ble_hs_id_set_rnd() etc

rahult-github:
> I think this should be in some common code eg ble_hs_id.c ? Some helpers for generating and internal helpers for validating type of addresses. There are similar checks for static random when calling ble_hs_id_set_rnd() etc

Thanks for the review . I have made a generic function to check if passed address is as per spec or not.

rahult-github:
@sjanc , Please let me know if you have any more suggestions / comments.

rymanluk:
@sjanc ping

rahult-github:
Hi @sjanc  , please take a look

PR #1612, opened 2023-09-15
nimble/ll: Drop received PDU if it is handled after scan was stopped

sjanc:
When doing continuous scan and channel is changed close to scan stop it is possible that PDU is received before interval timer is cancelled and RX is disabled. This may lead to processing received PDU only after scanner is already stopped. While this case was already handled in LL task it was not correctly handled in ISR. If this happen simply mark PDU to be ignored by LL and don't process it in ISR.

This is more likely to happen when small scan window is used.

Mostly affects initiator state as this may lead to assert when happen after CONNECTION_IND was already sent.


sjanc:
compliance check failure is due to STATS macros...

andrzej-kaczmarek (approved):

PR #1613, opened 2023-09-18
Fix C++ reserved identifiers in header guards

sjanc:
C++ doesn't allow underscores at the begining of identifier. This changes all such identifiers to FOO_H_ format.


elfring (commented):
* :eyes: [The name space rules are more detailed](https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier#answer-228797 "What are the rules about using an underscore in a C++ identifier?").
* :thought_balloon: Were source code places adjusted from any bundled software components?

sjanc:
I didn't update tinycrypt headers, and headers that are imported from mynewt core

PR #1618, opened 2023-09-22
ll/ble_iso_big: set default values for BIG params in BIG Create

KKopyscinski:
Big Create has less arguments than BIG Create Test, and missing params shall be calculated by controller. Added sane defaults that maximise BN assuming ISO_interval could be the same as MAX_Trasport_Latency.

Because this feature will be developed further for now there is no pre-transmissions (PTO = 0), and IRC is set to RTN.


KKopyscinski:
This is not working as intended and probably is not needed

PR #1619, opened 2023-09-22
[dnm] nimble/host: auracast API proposal

KKopyscinski:
This is rough outline of API that could be used to create BASE configuration and start broadcast (setup extended advertising, periodic advertising and BASE advertisements)


nimble/host/services/leaudio/broadcaster/include/leaudio/broadcaster/broadcaster.h
```diff
@@ -0,0 +1,57 @@
...
+struct broadcast_bis *
+broadcast_alloc_bis(struct broadcast_subgroup *sub);
+
+int
+broadcast_free_bis(struct broadcast_subgroup *big_sub,
+                   struct broadcast_bis *bis);
+
+int
+broadcast_start_advertising(struct broadcast_big *big, uint8_t instance,
+                            const uint8_t *name, ble_addr_t *addr,
```
rymanluk, line 49: struct with params please

nimble/host/services/leaudio/broadcaster/include/leaudio/broadcaster/broadcaster.h
```diff
@@ -0,0 +1,57 @@
...
+
+struct broadcast_bis *
+broadcast_alloc_bis(struct broadcast_subgroup *sub);
+
+int
+broadcast_free_bis(struct broadcast_subgroup *big_sub,
+                   struct broadcast_bis *bis);
+
+int
+broadcast_start_advertising(struct broadcast_big *big, uint8_t instance,
```
rymanluk, line 48: broadcast_create()
broadcast_start()
broadcast_stop()

nimble/host/services/leaudio/broadcaster/include/leaudio/broadcaster/broadcaster.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include <stdint.h>
+#include "host/ble_gap.h"
+#include "host/ble_iso.h"
+#include "leaudio/leaudio_common.h"
+
+#define BROADCAST_AUDIO_ANNOUNCEMENT_SVC_UUID               0x1852
+
+struct broadcast_big *
+broadcast_alloc_big(void);
```
rymanluk, line 28: as agreed memory will be managed by the application

nimble/host/services/leaudio/include/leaudio/leaudio_common.h
```diff
@@ -0,0 +1,61 @@
...
+
+/** TODO: these lengths may need to be changed to fit more data */
+#define METADATA_MAX_LEN (32)
+#define CODEC_SPEC_MAX_LEN (32)
+
+struct codec_id {
+    uint8_t format;
+    uint16_t company_id;
+    uint16_t vendor_specific;
+}__attribute__((packed));
```
rymanluk, line 33: remote packed from public header

nimble/host/services/leaudio/broadcaster/include/leaudio/broadcaster/broadcaster.h
rymanluk: broadcast_destroy()
broadcast_update()

nimble/host/services/leaudio/include/leaudio/leaudio_common.h
```diff
@@ -0,0 +1,61 @@
...
+
+struct broadcast_bis {
```
rymanluk, line 35: add helper functions to create / build the BIG structure.

nimble/host/services/leaudio/include/leaudio/leaudio_common.h
```diff
@@ -0,0 +1,61 @@
...
+    STAILQ_ENTRY(broadcast_bis) next;
+    uint8_t idx;
+    uint8_t codec_spec_config_len;
+    uint8_t codec_spec_config[CODEC_SPEC_MAX_LEN];
```
rymanluk, line 39: change to pointers maintained by app

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+/** BASE configuration describing broadcast advertisement */
+struct ble_broadcast_base_config {
+    /** Advertising instance used by broadcast */
+    uint8_t adv_instance;
+
+    /** Pointer to BIG configuration */
+    struct ble_audio_big *big;
+
+    /** Broadcast ID */
+    uint32_t broadcast_id;
```
rymanluk, line 83: ins't broadcast_id already in *big?

nimble/host/include/host/ble_iso.h
```diff
@@ -0,0 +1,51 @@
...
+};
+
+struct ble_iso_create_big_params {
+    uint16_t adv_handle;
+    uint8_t bis_cnt;
+};
+
+int ble_iso_create_big(struct ble_iso_create_big_params *create_params,
+                       struct ble_iso_big_params *big_params,
+                       uint8_t *big_id);
```
rymanluk, line 47: why pointer?

KKopyscinski: const pointer for params, big_id -> big_id_out

nimble/host/include/host/ble_iso.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef H_BLE_ISO_
+#define H_BLE_ISO_
+#include "syscfg/syscfg.h"
+
+#if MYNEWT_VAL(BLE_ISO)
+
+#include <inttypes.h>
+
+struct ble_iso_big_params {
+    uint8_t sdu_interval[3];
```
KKopyscinski, line 29: uint32

nimble/host/include/host/ble_iso.h
```diff
@@ -0,0 +1,51 @@
...
+    uint16_t max_sdu;
+    uint16_t max_transport_latency;
+    uint8_t rtn;
+    uint8_t phy;
+    uint8_t packing;
+    uint8_t framing;
+    uint8_t encryption;
+    uint8_t broadcast_code[16];
```
KKopyscinski, line 37: byte order hci

KKopyscinski: const pointer char
app must provide 16 bytes, 0 filled

nimble/host/include/host/ble_audio_common.h
```diff
@@ -0,0 +1,139 @@
...
+    uint8_t idx;
+
+    /** BIS level Codec Specific Configuration */
+    uint8_t codec_spec_config_len;
+
+    /** BIS level Codec Specific Configuration length */
+    uint8_t *codec_spec_config;
+};
+
+typedef int ble_audio_broadcast_sub_free_fn(void *cb_args);
```
KKopyscinski, line 53: leftover

nimble/host/include/host/ble_audio_common.h
```diff
@@ -0,0 +1,139 @@
...
+    /** Number of subgroups in BIG */
+    uint8_t num_subgroups;
+
+    /** Link list of subgroups */
+    STAILQ_HEAD(, ble_audio_subgroup) subs;
+};
+
+struct ble_audio_codec_config_params {
+    /** Mandatory field, codec sampling frequency. */
+    uint8_t sampling_freq;
```
KKopyscinski, line 97: add defines

nimble/host/include/host/ble_audio_common.h
```diff
@@ -0,0 +1,139 @@
...
+ *                              expected LTV structure size.
+ * @param codec_config_out_len  Total length of constructed Codec Specific
+ *                              Configuration.
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_build_codec_config(struct ble_audio_codec_config_params *params,
+                                 uint8_t *codec_config_out,
+                                 uint8_t codec_config_out_len);
```
KKopyscinski, line 138: pointer to len

nimble/host/include/host/ble_audio_common.h
```diff
@@ -0,0 +1,139 @@
...
+ *                              Codec Specific Configuration LTV structure.
+ *                              Memory size must be sufficient to fit
+ *                              expected LTV structure size.
+ * @param codec_config_out_len  Total length of constructed Codec Specific
+ *                              Configuration.
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_build_codec_config(struct ble_audio_codec_config_params *params,
```
KKopyscinski, line 136: const pointer

nimble/host/include/host/ble_audio_common.h
```diff
@@ -0,0 +1,139 @@
...
+     */
+    uint8_t codec_frame_blocks_per_sdu;
+};
+
+/**
+ * @brief Build Codec Specific Configuration LTV structure
+ *
+ * This function packs Codec Specific Configuration settings into LTV format.
+ *
+ * @param params                Pointer to a `ble_audio_codec_config_params`
```
KKopyscinski, line 123: param in/param out (here and everywhere)

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+    struct ble_gap_ext_adv_params *extended_params;
+
+    /** Parameters used to configure Periodic advertising */
+    struct ble_gap_periodic_adv_params *periodic_params;
+
+    /** BIG parameters */
+    struct ble_iso_big_params *big_params;
+
+    /** Broadcast name */
+    const uint8_t *name;
```
KKopyscinski, line 45: char

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+
+    /** Broadcast name length */
+    uint8_t name_len;
```
KKopyscinski, line 48: null terminated, remove

nimble/host/include/host/ble_audio_broadcast.h
KKopyscinski: uint8_t *svc_data; // additional advertising data for extended

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+#include "host/ble_gap.h"
+#include "host/ble_iso.h"
+#include "host/ble_audio_common.h"
+
+struct ble_broadcast_create_params {
+    /** BIG to broadcast */
+    struct ble_audio_big *big;
+
+    /** BLE address to use for advertising */
+    ble_addr_t *addr;
```
KKopyscinski, line 33: Consider ble_own_addr_t

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+    uint8_t adv_instance;
+
+    /** Pointer to BIG configuration */
+    struct ble_audio_big *big;
+
+    /** Broadcast ID */
+    uint32_t broadcast_id;
+
+    /** Broadcast ID */
+    uint8_t big_id;
```
KKopyscinski, line 86: = adv instance, to remove

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+    /** Broadcast name length */
+    uint8_t name_len;
+
+    /** Advertising instance */
+    uint8_t adv_instance;
+};
+
+struct ble_broadcast_update_params {
+    /** New BLE address to use for advertising */
+    ble_addr_t *addr;
```
KKopyscinski, line 56: remove

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+    const uint8_t *name;
+
+    /** Broadcast name length */
+    uint8_t name_len;
+
+    /** Advertising instance */
+    uint8_t adv_instance;
+};
+
+struct ble_broadcast_update_params {
```
KKopyscinski, line 54: remove

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+ * changed during it's lifetime. To modify BASE configuration destroy it using
+ * `ble_audio_broadcast_destroy` and recreate with `ble_audio_broadcast_create`
+ *
+ * @param params                Pointer to a `ble_broadcast_update_params`
+ *                              structure.
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_update(struct ble_broadcast_update_params *params);
```
KKopyscinski, line 180: (uint8* svc_metadata)

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+ * advertising and terminates them. After return advertising instance is free
+ * and must be configured again for future advertisements.
+ *
+ * @param base_config           Pointer to a `ble_broadcast_base_config`
+ *                              struct that broadcast shall be terminated for.
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_destroy(struct ble_broadcast_base_config *base_config);
```
KKopyscinski, line 163: void * usr_data

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_start(struct ble_broadcast_base_config *base_config);
+
+/**
+ * @brief Stop advertisements for given BASE configuration
+ *
+ * This function stops BASE advertisement by disabling extended and periodic
+ * advertising. Advertising instance is still configured and ready for resume.
```
KKopyscinski, line 140: terminate big

nimble/host/services/leaudio/auracast/include/leaudio/auracast/auracast.h
```diff
@@ -0,0 +1,97 @@
...
+ *                              advertisement is destroyed.
+ * @param gap_cb                GAP event callback to be associated with
+ *                              Auracast advertisement.
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_svc_auracast_create(struct ble_svc_auracast_create_params *params,
+                            struct ble_broadcast_base_config *config_out,
+                            ble_audio_broadcast_destroy_fn *destroy_cb,
```
KKopyscinski, line 53: void * arg

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+ *                              advertisement is destroyed.
+ * @param gap_cb                GAP event callback to be associated with BASE
+ *                              advertisement.
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_create(struct ble_broadcast_create_params *params,
+                               struct ble_broadcast_base_config *config_out,
+                               ble_audio_broadcast_destroy_fn *destroy_cb,
```
KKopyscinski, line 119: void *args

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+ * @param big                   Pointer to a `ble_audio_big` structure,
+ *                              that will be extended by the new subgroup
+ * @param sub                   Pointer to a `ble_broadcast_subgroup`
+ *                              structure, that will be filled out with
+ *                              supplied configuration
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_build_sub(struct ble_broadcast_subgroup_params params,
```
KKopyscinski, line 219: *

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,257 @@
...
+ * This function starts BASE advertisement on by enabling extended and periodic
+ * advertising.
+ *
+ * @param base_config           Pointer to a `ble_broadcast_base_config`
+ *                              struct that shall be used for broadcast.
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_start(struct ble_broadcast_base_config *base_config);
```
KKopyscinski, line 134: move big here

nimble/host/include/host/ble_audio_broadcast.h
KKopyscinski: AFAIK this is not needed (own_addr_type in ble_gap_ext_adv_params defines address, set separately)

KKopyscinski:
This code is still being tested. Auracast package (that is simplified broadcast) implementation is pending

KKopyscinski:
@kolodgrz please review

nimble/host/services/auracast/include/services/auracast/ble_svc_auracast.h
```diff
@@ -0,0 +1,100 @@
...
+ * This function configures advertising instance for extended and periodic
+ * advertisements to be ready for Auracast broadcast.
+ *
+ * @param[in] params            Pointer to a `ble_svc_auracast_create_params`
+ *                              structure that defines BIG and broadcast name.
+ * @param[out] config_out       Pointer to a `ble_broadcast_base_config`
+ *                              structure to return configuration of created
+ *                              Auracast advertisement.
+ * @param[in] destroy_cb        Optional callback to be called when Auracast
+ *                              advertisement is destroyed.
```
kolodgrz, line 44: broadcast is still valid even without advertisements BAP 1.0 `Table 6.2: Broadcast Audio Stream states` , `Continuous transmission of the EA is not required.` and `Continuous transmission of the PA is not required.`

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+
+    /** Parameters used to configure Periodic advertising */
+    struct ble_gap_periodic_adv_params *periodic_params;
+
+    /** Broadcast name - null terminated.
+     * Set NULL to not include in advertising
+     */
+    const char *name;
+
+    /** Advertising instance */
```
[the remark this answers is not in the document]
KKopyscinski: in `host/ble_gap.h` it's just ` Advertising instance`. Maybe keep it consistent?

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+    /** Parameters used to configure Extended advertising */
+    struct ble_gap_ext_adv_params *extended_params;
+
+    /** Parameters used to configure Periodic advertising */
+    struct ble_gap_periodic_adv_params *periodic_params;
+
+    /** Broadcast name - null terminated.
+     * Set NULL to not include in advertising
+     */
+    const char *name;
```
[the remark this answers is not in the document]
KKopyscinski: Hmm I don't see in spec this length requirement. Could you give a paragraph? `LOCAL NAME` AD type doesn't seem to have length requirement, neither complete or shortened.  GATT characteristic can have lenght in range 0 to 248. Is it somewhere in BAP maybe? If so I'd add this to documentation and added check in implementation

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ * advertisements to be ready for broadcast with BASE configuration.
+ *
+ * @param[in] params            Pointer to a `ble_broadcast_base_params`
+ *                              structure that defines BASE, extended
+ *                              advertising and periodic advertising
+ *                              configuration.
+ * @param[out] config_out       Pointer to a `ble_broadcast_base_config`
+ *                              structure to return configuration of created
+ *                              BASE advertisement.
+ * @param[in] destroy_cb        Optional callback to be called when BASE
```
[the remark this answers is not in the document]
KKopyscinski: I'm not sure either but this is how we discussed this offline. AFAIK it's the similar logic as with extended and periodic: periodic is auxilary of extended, and similary BIGInfo is auxilary of periodic. But when extended is stopped periodic is still being advertised. So when we want to stop all advertisements, we need to stop both extended, periodic and terminate BIG. Without it, extended and periodic would stop but BIGInfo would still advertise. Am I correct here @andrzej-kaczmarek ?

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ *                              advertisement is destroyed.
+ * @param[in] args              Optional arguments to be passed to `destroy_cb`
```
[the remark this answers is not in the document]
KKopyscinski: I'm planning to add sample that acts as usb soundcard for pc, where all of this API will be used. But that requires some additional code, outside of scope of this PR, like LC3. For now I added shell commands to test this code. I think it would be nice to add unittests too, when this API is semi-established I'll add them I think

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ * @param[in] gap_cb            GAP event callback to be associated with BASE
```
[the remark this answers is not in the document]
KKopyscinski: Yes, this is generic one. If you mean callback for events related to big it's declared in BIG, here: https://github.com/apache/mynewt-nimble/pull/1619/files#diff-b1517448655ef48c5df75a3f4575a12dd70843b36608468fbc001057d3f5f1efR103

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ *                              advertisement.
+ *
+ * @return                      0 on success;
```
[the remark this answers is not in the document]
KKopyscinski: As far as I can see it's either that or what I used:
```c
 /*
 * @return                          0 on success;
 *                                  A non-zero value on failure.
 */
```
I can see it in L2CAP, I can change it obviously

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+                               struct ble_broadcast_base_config *config_out,
+                               ble_audio_broadcast_destroy_fn *destroy_cb,
+                               void *args,
+                               ble_gap_event_fn *gap_cb);
+
+/**
+ * @brief Start advertisements for given BASE configuration
+ *
+ * This function starts BASE advertisement on by enabling extended and periodic
+ * advertising and creates BASE for this instance.
```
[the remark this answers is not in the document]
KKopyscinski: You're right, I'll remove it from here, it's a leftover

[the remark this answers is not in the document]
KKopyscinski: Hmm the wording I used here seems confusing. So `ble_audio_broadcast_create` created BASE in host, but here we start advertising it. I think proper wording would be
```
 This function starts BASE advertisement by enabling extended,  periodic and BIGInfo advertisements for this instance.
```

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+
+    /** Broadcast ID */
+    uint32_t broadcast_id;
+};
+
+typedef int ble_audio_broadcast_destroy_fn(struct ble_audio_base *base,
+                                           void *args);
+
+/** BASE configuration describing broadcast advertisement */
+struct ble_broadcast_base_config {
```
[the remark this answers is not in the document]
KKopyscinski: So my idea for this was that `ble_broadcast_create_params` is `in` parameter. So it's used to configure instances and so on. If function is successful, `ble_broadcast_base_config` will be filled with `out` values, some copied from `ble_broadcast_create_params`, some generated, like `base->broadcast_id`. On second thought I see that this is not the best, I think I'll remove this parameter and save what's needed in host, so the app doesn't need to save `ble_broadcast_base_config` on it's own. And BASE can be identified just by `adv_instance`

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+    uint16_t svc_data_len;
+
+    /** Broadcast ID */
+    uint32_t broadcast_id;
+};
+
+typedef int ble_audio_broadcast_destroy_fn(struct ble_audio_base *base,
+                                           void *args);
+
+/** BASE configuration describing broadcast advertisement */
```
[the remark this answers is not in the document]
KKopyscinski: as in https://github.com/apache/mynewt-nimble/pull/1619#discussion_r1371242163, I'll remove it entirely

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ */
+int ble_audio_broadcast_start(const struct ble_broadcast_base_config *base_config,
+                              ble_iso_event_fn *cb, void *cb_arg);
+
+/**
+ * @brief Stop advertisements for given BASE configuration
+ *
+ * This function stops BASE advertisement by disabling extended and periodic
+ * advertising. Advertising instance is still configured and ready for resume.
+ * BASE will be terminated.
```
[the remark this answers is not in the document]
KKopyscinski: I think it's the same as what I wrote in https://github.com/apache/mynewt-nimble/pull/1619#discussion_r1371212124

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_stop(const struct ble_broadcast_base_config
+                             *base_config);
+
+/**
+ * @brief Destroy advertisements for given BASE configuration
+ *
+ * This function terminates BASE advertisement instance.
```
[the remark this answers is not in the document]
KKopyscinski: Yes

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_destroy(const struct ble_broadcast_base_config
+                                *base_config);
+
+/**
+ * @brief Update advertisements for given BASE configuration
+ *
+ * This function can be used to update extended advertisements. This function
```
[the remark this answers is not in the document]
KKopyscinski: Maybe I'm mistaken, but only EXT Adv can have additional data here? PA has basically BIG structure, with subgroups and BISes, and BIGInfo is something totally else. EXT has Broadcast Name and `service_data` of BLE_BROADCAST_AUDIO_ANNOUNCEMENT_SVC_UUID, and additional data that can be updated here

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ * cannot be used to update BASE configuration itself, because BIG cannot be
+ * changed during it's lifetime. To modify BASE configuration destroy it using
```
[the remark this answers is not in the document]
KKopyscinski: I agree

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ *                              supplied configuration
+ * @param[in] params            Pointer to a `ble_broadcast_subgroup_params`
+ *                              structure, containing information about new
+ *                              subgroup
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_audio_broadcast_build_sub(struct ble_audio_base *base,
+                                  struct ble_audio_subgroup *sub,
```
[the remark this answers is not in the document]
KKopyscinski: I agree, it's more descriptive

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,271 @@
...
+ * @brief Build BIS structure
+ *
+ * This is a helper function can be used to fill out `ble_broadcast_bis`
+ * structure. Created BIS extends BIS list in provided subgroup.
+ * This function increases `bis_cnt` in subgroup structure.
+ *
+ * @param[out] sub              Pointer to a `ble_broadcast_subgroup`
+ *                              structure,
+ *                              that will be extended by the new BIS
+ * @param[out] bis              Pointer to a `ble_broadcast_bis`
```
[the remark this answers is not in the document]
KKopyscinski: I'd add line "In case of error, filled out data may be erroneous.". Pointers do not change value, so they are ok - but data pointed by them will contain invalid values. Would that be ok?

nimble/host/include/host/ble_audio_common.h
```diff
@@ -0,0 +1,183 @@
...
+    uint8_t format;
+
+    /** Company ID */
+    uint16_t company_id;
+
+    /** Vendor Specific Codec ID */
+    uint16_t vendor_specific;
+};
+
+struct ble_audio_bis {
```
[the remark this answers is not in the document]
KKopyscinski: I don't think so, sometimes it's just `next` sometimes it has some name appended. I can change that to `bis_next` and `big_sub_next` (in `ble_audio_big_subgroup`). But `next` seems to be perfectly fine, like in `STAILQ_ENTRY(ble_l2cap_coc_srv) next;`

nimble/host/include/host/ble_audio_common.h
```diff
@@ -0,0 +1,183 @@
...
+    /** Subgroup Metadata length*/
+    uint8_t metadata_len;
+
+    /** Link list of BISes */
+    STAILQ_HEAD(, ble_audio_bis) bises;
+};
+
+struct ble_audio_base {
+    /** Broadcast ID */
+    uint32_t broadcast_id;
```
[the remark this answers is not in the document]
KKopyscinski: Yes, I think this can be removed from here

nimble/host/include/host/ble_iso.h
```diff
@@ -0,0 +1,127 @@
...
+    void *cb_arg;
+};
+
+int ble_iso_create_big(const struct ble_iso_create_big_params *create_params,
+                       const struct ble_iso_big_params *big_params);
+
+int ble_iso_terminate_big(uint8_t big_id);
+
+void
+ble_gap_rx_create_big_complete(const struct
```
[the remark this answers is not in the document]
KKopyscinski: this should change prefix to ble_iso, and be moved to src/ble_iso_priv.h (must be created). It's for handling HCI event

nimble/host/include/host/ble_iso.h
```diff
@@ -0,0 +1,127 @@
...
+                               ble_hci_ev_le_subev_create_big_complete *ev);
+void
+ble_gap_rx_terminate_big_complete(const struct
```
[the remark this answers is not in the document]
KKopyscinski: ditto

nimble/host/services/auracast/include/services/auracast/ble_svc_auracast.h
```diff
@@ -0,0 +1,100 @@
...
+ * @param[in] args              Optional arguments to be passed to `destroy_cb`
+ * @param[in] gap_cb            GAP event callback to be associated with
+ *                              Auracast advertisement.
+ *
+ * @return                      0 on success;
+ *                              A non-zero value on failure.
+ */
+int ble_svc_auracast_create(const struct
+                            ble_svc_auracast_create_params *params,
+                            struct ble_broadcast_base_config *config_out,
```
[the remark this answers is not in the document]
KKopyscinski: I agree and my response is also the same :)

nimble/host/include/host/ble_audio_broadcast.h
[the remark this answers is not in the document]
rymanluk: @KKopyscinski  the limitation is coming from here:

https://www.bluetooth.com/specifications/pbp-1-0/

5.1. Broadcast Name Data AD Type

nimble/host/include/host/ble_audio_broadcast.h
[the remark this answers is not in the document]
rymanluk: Yes, it should be advertising instance.

nimble/host/include/host/ble_audio_broadcast.h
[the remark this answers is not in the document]
rymanluk: I believe this callback is called when `ble_audio_broadcast_destroy` is called. Is that correct?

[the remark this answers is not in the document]
KKopyscinski: yes, I'll change description to be more precise

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * This is a helper function can be used to fill out `ble_broadcast_subgroup`
+ * structure. Created subgroup extends subgroup list in provided BASE.
+ * This function increases `num_subgroups` in BASE structure.
+ *
+ * @param[out] base             Pointer to a `ble_audio_base` structure,
+ *                              that will be extended by the new subgroup.
+ *                              In case of error, filled out data may be
+ *                              erroneous.
+ * @param[out] sub              Pointer to a `ble_broadcast_subgroup`
```
rymanluk, line 188: nit: maybe instead of `sub` use a `subgroup` here and in function name?

nimble/host/include/host/ble_audio_common.h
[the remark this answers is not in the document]
rymanluk: agree with @KKopyscinski

nimble/host/services/auracast/include/services/auracast/ble_svc_auracast.h
rymanluk: That is true, BAP allows that. But for Public Broadcaster we want to make sure broadcast source is visible :) therefore PA will be there whenever auracast is created.

nimble/host/services/auracast/include/services/auracast/ble_svc_auracast.h
```diff
@@ -0,0 +1,100 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <stdint.h>
+#include "host/ble_gap.h"
+#include "host/ble_audio_common.h"
+#include "host/ble_audio_broadcast.h"
+
+struct ble_svc_auracast_create_params {
```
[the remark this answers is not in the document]
rymanluk: do you mean additional service data to `Broadcast Audio Announcement Service UUID` ?

nimble/host/services/auracast/include/services/auracast/ble_svc_auracast.h
```diff
@@ -0,0 +1,93 @@
...
+    struct ble_audio_base *base;
+    struct ble_iso_big_params *big_params;
+    const char *name;
+    uint8_t adv_instance;
```
rymanluk, line 29: how application knows `adv_instnace` ?

nimble/host/services/auracast/include/services/auracast/ble_svc_auracast.h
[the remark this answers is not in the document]
KKopyscinski: If we want this field I can add it, like there is in `ble_broadcast_create_params`

nimble/syscfg.yml
```diff
@@ -81,7 +81,7 @@ syscfg.defs:
             This allows to configure maximum size of advertising data and
             scan response data used in LE Advertising Extensions.
             Valid range 31-1650.
-        value: 31
+        value: 251
```
rymanluk, line 84: if we want to change this, it should be separate patch.

nimble/host/services/auracast/include/services/auracast/ble_svc_auracast.h
KKopyscinski: It chooses it by itself, like we do for ext adv and PA. So we may expect that it remembers what it chose.

nimble/host/src/ble_audio_common.c
```diff
@@ -0,0 +1,51 @@
...
+#include <string.h>
+
+int ble_audio_build_codec_config(const struct ble_audio_codec_config_params
+                                 *params,
+                                 uint8_t *codec_config_out,
+                                 uint8_t *codec_config_out_len)
+{
+    uint8_t len = 0;
+
+    memcpy(codec_config_out, &params->sampling_freq, 1);
```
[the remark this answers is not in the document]
KKopyscinski: I don't know how I missed this. Also, I'll try to add this to shell

nimble/host/src/ble_audio_common.c
```diff
@@ -0,0 +1,51 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "host/ble_audio_common.h"
+#include <string.h>
+
+int ble_audio_build_codec_config(const struct ble_audio_codec_config_params
```
KKopyscinski, line 23: change into macro

nimble/syscfg.yml
andrzej-kaczmarek: also if this is required by some feature we should better make 251 conditional depending on that feature, this way default configuration is not changed

nimble/host/services/auracast/syscfg.yml
```diff
@@ -0,0 +1,29 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
+  BLE_SVC_AURACAST_ADV_INSTANCE:
+    description: >
+      Defines which advertising instance is used for Auracast.
+      If set to -1 last advertising instance is used, and if MESH is enabled, one before that.
```
kolodgrz, line 24: What is the reason for defining in configuring which advertising instance should be used ?

porting/examples/linux/include/syscfg/syscfg.h
```diff
@@ -771,6 +783,14 @@
 #define MYNEWT_VAL_BLE_L2CAP_SIG_MAX_PROCS (1)
 #endif
 
+#ifndef MYNEWT_VAL_BLE_MAX_BIG
+#define MYNEWT_VAL_BLE_MAX_BIG (MYNEWT_VAL_BLE_MULTI_ADV_INSTANCES)
+#endif
+
+#ifndef MYNEWT_VAL_BLE_MAX_BIS
+#define MYNEWT_VAL_BLE_MAX_BIS (4)
```
kolodgrz, line 791: it's per BIG ? or total ?

apps/btshell/src/btshell.h
```diff
@@ -182,6 +185,36 @@ int btshell_l2cap_send(uint16_t conn, uint16_t idx, uint16_t bytes);
 int btshell_l2cap_reconfig(uint16_t conn_handle, uint16_t mtu,
                            uint8_t num, uint8_t idxs[]);
 
+#if (MYNEWT_VAL(BLE_ISO_BROADCASTER))
+int btshell_broadcast_base_add(uint8_t adv_instance, uint32_t presentation_delay);
```
kolodgrz, line 189: This is original concept to identify broadcast with adv_instance. Can one advertising instance be used for "multiple" purposes in context of one program ? Broadcast ID seems to be an unique ID

kolodgrz, line 189: What this method is doing ? Isn't this a "StartBasicAudioAnnouncement" function ? not sure about having here presentation delay. Where is the "real" BASE prepared ? 

What's the flow ? How to start from scratch a broadcast ?
1. Who prepares Broadcast Audio Announcement
2. Who prepares Broadcast Source Audio Endpoint
3. Where it'll be attached to EA, PA
4. Where EA, PA would be started
5. Where data would be bind to BISes
?

KKopyscinski: Broadcast ID is indeed unique ID. Internally identifying broadcast with adv_instance (which is also used as BIG ID) saves memory. And AFAIK each adv instance can be only used for single broadcast either way - because it needs to have corelated Ext Adv and PA, and this cannot be started on already active instance. This means, you cannot start another broadcast on same instance.

KKopyscinski: So the flow is like that:
1. `base_add` (calls `btshell_broadcast_base_add`) - allocates memory for `ble_audio_base`, `ble_iso_big_params`. Sets `presentation_delay` and `adv_instance` to base
2. `big_sub_add` (calls `btshell_broadcast_big_sub_add`) - adds subgroup to BASE identified by `adv_instance` (allocates memory, sets fields)
3. `bis_add` (calls `btshell_broadcast_bis_add`) - similar to `big_sub_add`, but for BIS. Adds bis to last added subgroup in BIG identified by `adv_instance`
4. `broadcast_create` (calls `btshell_broadcast_create` - basically calls `ble_audio_broadcast_create`. Here EA and PA params are set, instance configured with both advertisements.
5. `broadcast_start` (calls `btshell_broadcast_start`) - this finally calls `ble_audio_broadcast_start`, so EA, PA are started. `ble_iso_create_big` is also called, and BIG Info advertisements start to broadcast
6. In shell, I haven't added data stream flow to BISes. I'm not sure how this could be done in shell to make sense. I'm preparing standalone demo that does all this, in similar way extended_advertiser demo does. So at this point app has to have some routine (like loop or thread) that calls `ble_iso_tx`, with payload. For BIS `conn_handle` for this function is BIS handle.
7. At this point broadcast can be stopped (`broadcast_stop`), destroyed (`broadcast_destroy`) or updated (`broadcast_update`).

porting/examples/linux/include/syscfg/syscfg.h
KKopyscinski: As for now, total. I have nothing against changing it to per BIG, I feel like both approaches are  valid

nimble/host/services/auracast/syscfg.yml
```diff
@@ -0,0 +1,29 @@
...
+      Must be set to value in range 0 .. BLE_MULTI_ADV_INSTANCES.
```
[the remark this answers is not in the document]
KKopyscinski: changed to -1 .. BLE

nimble/host/services/auracast/syscfg.yml
KKopyscinski: For broadcast, we can choose. For Auracast user has freedom to define single instance (for now) to be used, OR give up this decision to package. This can simplify process of writting apps, especially simple ones

kolodgrz (commented):
Initial Broadcast Source Implementation ?

nimble/host/src/ble_iso.c
```diff
@@ -0,0 +1,369 @@
...
+    void *cb_arg;
+};
+
+static SLIST_HEAD(, ble_iso_big) ble_iso_bigs;
+static struct os_mempool ble_iso_big_pool;
+static os_membuf_t ble_iso_big_mem[
+    OS_MEMPOOL_SIZE(MYNEWT_VAL(BLE_MAX_BIG), sizeof (struct ble_iso_big))];
+
+static struct ble_iso_big *
+ble_iso_big_alloc(uint16_t adv_handle)
```
kolodgrz, line 50: Advertising handle shouldn't be 8bit ?

KKopyscinski:
> Initial Broadcast Source Implementation ?

Changed in commit messages and PR title

nimble/host/src/ble_iso.c
```diff
@@ -0,0 +1,369 @@
...
+            break;
+        }
+    }
+
+    if (!max_pdu) {
+        return BLE_HS_ENOENT;
+    }
+
+    if (data_len < max_pdu) {
+        rc = ble_iso_tx_complete(conn_handle, data, data_len);
```
kolodgrz, line 362: what about return ble_iso_tx_complete  ?

nimble/host/include/host/ble_audio_broadcast.h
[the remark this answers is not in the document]
kolodgrz: I mean args for destroy_cb :)

nimble/host/include/host/ble_audio_broadcast.h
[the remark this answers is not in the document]
kolodgrz: It depends on what is the common way to handle it across nimble. Is this a common solution ?

nimble/host/src/ble_iso.c
KKopyscinski: yes :)

nimble/host/include/host/ble_audio_broadcast.h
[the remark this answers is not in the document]
KKopyscinski: Anything you need, it's a void pointer, app will cast it back to type it needs. Intention is, that for applications using dynamic memory allocation (like mempool) for BASE, BIG subgroups, BISes etc. this could be used as a pointer to to these things, and then app can free these resources. You can see that in `btshell_broadcast_destroy_fn`

nimble/host/include/host/ble_audio_broadcast.h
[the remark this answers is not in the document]
KKopyscinski: The way that's implemented right now, it has no true way to "fail", because all that happens there is memcpy and STAILQ_INSERT_HEAD/TAIL, plus some fields are set. So we always return 0. Description can be made future proof, as it is now, so in the future some error checking might take place. Maybe this comment should be ommited for now, as this function weakest point is memcpy, and we cannot really check this for error. Any thoughts @rymanluk @sjanc ?

nimble/host/include/host/ble_audio_common.h
```diff
@@ -0,0 +1,179 @@
...
+ * @param _audio_channel_alloc         Audio_Channel_Allocation -
+ *                                     four octet value
+ * @param _octets_per_codec_frame      Octets_Per_Codec_Frame -
+ *                                     two octet value
+ * @param _codec_frame_blocks_per_sdu  Codec_Frame_Blocks_Per_SDU -
+ *                                     single octet value
+ *
+ * @return          Pointer to a `ble_uuid16_t` structure.
+ */
+#define BLE_AUDIO_BUILD_CODEC_CONFIG(_sampling_freq,                          \
```
KKopyscinski, line 91: I noticed that this does not behave as expected: I'll rework this macro to fit spec better

nimble/host/include/host/ble_audio_broadcast.h
```diff
@@ -0,0 +1,245 @@
...
+};
+
+/**
+ * @brief Build BIS structure
+ *
+ * This is a helper function can be used to fill out `ble_broadcast_bis`
+ * structure. Created BIS extends BIS list in provided subgroup.
+ * This function increases `bis_cnt` in subgroup structure.
+ *
+ * @param[out] sub              Pointer to a updated `ble_broadcast_subgroup`
```
rymanluk, line 224: can we name it subgroup.

and also this is in/out param right?

nimble/host/include/host/ble_audio_broadcast.h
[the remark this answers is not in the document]
rymanluk: Imho, this is exactly why the `return` is the result of operation. In a case of error returned, we don't care about [out] params and its state is unknown. Current description should be  good.

porting/examples/linux/include/syscfg/syscfg.h
rymanluk: total should be fine.

nimble/host/include/host/ble_audio_broadcast.h
KKopyscinski: right, this should be `ble_audio_big_subgroup`. How to mark in/out params?

nimble/host/services/auracast/syscfg.yml
```diff
@@ -0,0 +1,27 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
+  BLE_SVC_AURACAST_ADV_INSTANCE:
+    description: >
+      Defines which advertising instance is used for Auracast.
+    value: 42
```
andrzej-kaczmarek, line 24: I don't think this should be a syscfg value. app has to call `ble_svc_auracast_create` anyway so it can also include adv instance.

nimble/host/services/auracast/syscfg.yml
```diff
@@ -0,0 +1,27 @@
...
+
+syscfg.restrictions:
+  - BLE_PERIODIC_ADV: 1
```
andrzej-kaczmarek, line 27: this likely doesn't work as expected

nimble/host/services/auracast/src/ble_svc_auracast.c
```diff
@@ -0,0 +1,88 @@
...
+
+    rc = ble_hs_id_infer_auto(0, &own_addr_type);
+    if (rc != 0) {
+        return rc;
+    }
+
+    extended_params.own_addr_type = own_addr_type;
+    extended_params.sid = AURACAST_ADV_INSTANCE;
+    extended_params.secondary_phy = MYNEWT_VAL(BLE_PHY_2M) ?
+                                    BLE_HCI_LE_PHY_2M : BLE_HCI_LE_PHY_1M;
```
andrzej-kaczmarek, line 56: this should be also a parameter

nimble/host/services/auracast/src/ble_svc_auracast.c
```diff
@@ -0,0 +1,88 @@
...
+    uint8_t own_addr_type;
+    int rc;
+
+    rc = ble_hs_id_infer_auto(0, &own_addr_type);
+    if (rc != 0) {
+        return rc;
+    }
+
+    extended_params.own_addr_type = own_addr_type;
+    extended_params.sid = AURACAST_ADV_INSTANCE;
```
andrzej-kaczmarek, line 54: this should be parameter

nimble/host/services/auracast/src/ble_svc_auracast.c
```diff
@@ -0,0 +1,88 @@
...
+    struct ble_gap_periodic_adv_params periodic_params = { 0 };
+    struct ble_gap_ext_adv_params extended_params = {
+        .scannable = 0,
+        .connectable = 0,
+        .primary_phy = BLE_HCI_LE_PHY_1M,
+    };
+    uint8_t own_addr_type;
+    int rc;
+
+    rc = ble_hs_id_infer_auto(0, &own_addr_type);
```
andrzej-kaczmarek, line 48: own addr type should be also specified by app

nimble/host/services/auracast/syscfg.yml
KKopyscinski: as discussed offline, I'd change it to start index of first auracast (BIG). When user creates more, new BIG will occupy <this + offset> instance.

add restriction that additional instances are > 1

KKopyscinski: done

nimble/host/services/auracast/src/ble_svc_auracast.c
KKopyscinski: done, added check if requested PHY is supported

nimble/host/services/auracast/src/ble_svc_auracast.c
KKopyscinski: done

nimble/host/services/auracast/src/ble_svc_auracast.c
KKopyscinski: done

KKopyscinski:
Added sample application for broadcast, fixed minor bug in ble_iso.c. Build check (all apps) is failing, seems to me app tries to be built with ll, but was designed for two core boards (nRF53, net core running ble-hci) I'm not sure how this should be resolved, maybe some config changes?

nimble/host/services/auracast/syscfg.yml
KKopyscinski: I removed this all together, added 7abb8e66a5d672d94d2dde33ff294a8809e9abde to fetch non-configured advertising instance and Auracast will report back what was found. Maybe this commit should be moved to another PR

nimble/host/services/auracast/src/ble_svc_auracast.c
```diff
@@ -0,0 +1,102 @@
...
+            return i;
+        }
+    }
+
+    return 0;
+}
+
+int
+ble_svc_auracast_create(const struct ble_svc_auracast_create_params *params,
+                        uint8_t *adv_instance,
```
KKopyscinski, line 43: auracast_instance (change name)

sjanc (approved):

PR #1620, opened 2023-09-22
  nimble/host:Handled error return value BLE_ERR_UNK_CONN_ID when encryption is enabled without distributing IRK.

ESPAbhinav:
If address resolution is enabled in the controller it will reject any RL modification during scanning, adv, etc. Hence controller sends a Command disallowed error which needs to be sent back to the application. Also when encryption is enabled without distributing IRK and hence the controller has no IRK but the host has LTK then the controller will return error 0x202 which needs not to be returned to delete LTK. (Hence it is skipped to return when the error code is 0 or 0x202.)

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -6042,6 +6042,7 @@ ble_gap_unpair(const ble_addr_t *peer_addr)
 {
 #if NIMBLE_BLE_SM
     struct ble_hs_conn *conn;
+    int rc;
 
     if (!ble_hs_is_enabled()) {
         return BLE_HS_EDISABLED;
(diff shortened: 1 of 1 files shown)


sjanc (commented):
code looks OK but could you make commit message match convention?

commit message and topic should fit in 72 lines (if possible)
topic : "nimble/host:Handled error return value BLE_ERR_UNK_CONN_ID"

and then commit message (new lines at <72 chars)
"When encryption is enabled without distributing IRK and hence controller
 has no IRK but host has LTK. If address resolution is enabled in the
 controller it will reject any RL ......"


This is mainly for keeping 'git log' consitent

ESPAbhinav:
Sure.Thanks for this suggestion. I will follow commit conventions and change the commit message.

nimble/host/src/ble_gap.c
```diff
@@ -6062,6 +6063,9 @@ ble_gap_unpair(const ble_addr_t *peer_addr)
 
     ble_hs_pvcy_remove_entry(peer_addr->type,
```
sjanc, line 6064: looks like return value from  ble_hs_pvcy_remove_entry() is never stored in rc...

abhi152: Thanks for your suggestion. So I have made further changes including above mentioned in this API and will update this PR.

PR #1626, opened 2023-09-26
ci: Add workflow for testing various GCC versions

sjanc:
This will help make sure we build correctly with more GCC version.

### DIFF
--- a/.github/workflows/build_cc_target.yml
+++ b/.github/workflows/build_cc_target.yml
@@ -1,0 +1,62 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


.github/workflows/build_cc_target.yml
```diff
@@ -0,0 +1,62 @@
...
+on: [push, pull_request]
+
+jobs:
+  targets:
+    name: Build GCC test target
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        gcc: ['12.2.Rel1', '11.3.Rel1', '10.3-2021.10', '9-2019-q4', '8-2018-q4']
```
andrzej-kaczmarek, line 31: perhaps we can remove gcc 8 (and 9?) and add 13 instead?

andrzej-kaczmarek (approved):

.github/workflows/build_cc_target.yml
sjanc: I left 9-2019-q4 because it is available in Ubuntu 20.04 which is stil quite common,  we can drop 8 I guess.

I'll add GCC 13 when it becomes available via github action we are using (https://github.com/carlosperate/arm-none-eabi-gcc-action)

PR #1630, opened 2023-10-03
nimble/ll: Add vs hci to set local IRK

andrzej-kaczmarek:
This adds vendor-specific HCI command to set local IRK in controller.
    
Local IRK, if set, is used to generate local RPA in use cases where own address type was set to 0x02 or 0x03 but peer address is not added to resolving list. This for example allows to handle initiating connection to a new peer using RPA as our local address entirely in LL. Without that command it would be required for host to generate an RPA, set it as random address and connect using random address. This however doesn't work well with NimBLE host.
    
If no IRK is set (or set to all-zero), the controller behaves as usual which makes it safe to enable as it won't break anything.

### DIFF
--- a/nimble/controller/src/ble_ll_resolv.c
+++ b/nimble/controller/src/ble_ll_resolv.c
@@ -48,6 +48,16 @@ struct ble_ll_resolv_data g_ble_ll_resolv_data;
 __attribute__((aligned(4)))
 struct ble_ll_resolv_entry g_ble_ll_resolv_list[MYNEWT_VAL(BLE_LL_RESOLV_LIST_SIZE)];
 
+#if MYNEWT_VAL(BLE_LL_HCI_VS_LOCAL_IRK)
+struct local_irk_data {
+    uint8_t is_set;
+    uint8_t irk[16];
+    uint8_t rpa[6];
+};
+/* 0 is for public, 1 is for static address */
+static struct local_irk_data g_local_irk[2];
+#endif
+
 /**
  * Called to determine if a change is allowed to the resolving list at this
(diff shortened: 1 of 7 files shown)


nimble/controller/src/ble_ll_resolv.c
```diff
@@ -141,6 +156,14 @@ ble_ll_resolv_rpa_timer_cb(struct ble_npl_event *ev)
         ++rl;
     }
 
+#if MYNEWT_VAL(BLE_LL_HCI_VS_LOCAL_IRK)
+    OS_ENTER_CRITICAL(sr);
+    if (g_local_irk.is_set) {
+        generate_rpa(g_local_irk.irk, g_local_irk.rpa);
```
sjanc, line 162: RPA could be generated before critical section, and critical section could be only under if (as it may be cleared only from same task)

nimble/include/nimble/hci_common.h
```diff
@@ -1208,8 +1208,10 @@ struct ble_hci_vs_set_data_len_rp {
 struct ble_hci_vs_set_antenna_cp {
     uint8_t antenna;
 } __attribute__((packed));
-
-
+#define BLE_HCI_OCF_VS_SET_LOCAL_IRK                   (MYNEWT_VAL(BLE_HCI_VS_OCF_OFFSET) + (0x000A))
+struct ble_hci_vs_set_local_irk_cp {
+    uint8_t irk[16];
```
sjanc, line 1213: just a loose thought,  maybe this could also have address type?  public vs static_random

sjanc (commented):
I like the idea:)

sjanc (approved):

PR #1635, opened 2023-10-06
host/sm: Add CSIS SIRK encryption

m-gorecki:
This adds functions and algorithms to encrypt and decrypt SIRK from Coordinated Set Identification Service.

This also adds API to resolve RSI. Application can use it to find devices that are part of Coordinated Set.

### DIFF
--- a/nimble/host/src/ble_sm.c
+++ b/nimble/host/src/ble_sm.c
@@ -2910,4 +2910,107 @@ ble_sm_create_chan(uint16_t conn_handle)
     return chan;
 }
 
+#if MYNEWT_VAL(BLE_SM_CSIS_SIRK)
+int
+ble_sm_csis_decrypt_sirk(const uint8_t *ltk, const uint8_t *enc_sirk, uint8_t *out)
+{
+    int rc;
+
+    /* Decrypt SIRK with sdf(K, EncSIRK)  */
+    rc = ble_sm_alg_csis_sdf(ltk, enc_sirk, out);
+
+    return rc;
+}
+
+int
+ble_sm_csis_resolve_rsi(const uint8_t *rsi, const uint8_t *sirk,
+                        const ble_addr_t *ltk_peer_addr)
+{
+    struct ble_store_key_sec key_sec;
+    struct ble_store_value_sec value_sec;
+    uint8_t plaintext_sirk[16] = {0};
+    uint8_t local_hash[3] = {0};
+    uint8_t prand[3] = {0};
+    uint8_t hash[3] = {0};
+    int rc;
+
+    memcpy(hash, rsi, 3);
+    memcpy(prand, rsi + 3, 3);
+
+    if (ltk_peer_addr) {
+        memset(&key_sec, 0, sizeof(key_sec));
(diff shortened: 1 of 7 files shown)


nimble/host/src/ble_sm.c
```diff
@@ -2910,4 +2910,109 @@ ble_sm_create_chan(uint16_t conn_handle)
     return chan;
 }
 
+#if MYNEWT_VAL(BLE_SM_SC)
```
andrzej-kaczmarek, line 2913: I'd add `BLE_SM_CSIS_SIRK` to enable everything related to handling CSIS SIRK and move this to `ble_sm_csis.c`

also mark syscfg as experimental since we may want to move it to some place more appropriate and rename in future with addition of le audio code.

nimble/host/src/ble_sm.c
```diff
@@ -2910,4 +2910,109 @@ ble_sm_create_chan(uint16_t conn_handle)
...
+    } else {
+        memcpy(plaintext_sirk, sirk, 16);
+    }
+
+    rc = ble_sm_alg_csis_sih(plaintext_sirk, prand, local_hash);
+    if (rc != 0) {
+        return rc;
+    }
+
+    if (memcmp(local_hash, hash, 3) == 0) {
```
andrzej-kaczmarek, line 2966: better write this as:
```c
if (memcmp(local_hash, hash, 3)) {
    return BLE_HS_EAUTHEN;
}

return 0;
```

this way return value in case of success is at the very end of function as expected

nimble/host/src/ble_sm_alg.c
```diff
@@ -418,6 +418,139 @@ ble_sm_alg_g2(const uint8_t *u, const uint8_t *v, const uint8_t *x,
...
+    int rc;
+    uint8_t t[16] = {0};
+    uint8_t salt_be[16] = {0};
+    uint8_t n_be[16] = {0};
+
+    /* XXX: Spec does not specify the maximum N and P parameters length.
+     * We assume that 16 bytes is enough and return error if passed len value is greater
+     * than that */
+    if ((n_len > 16) || (p_len > 16)) {
+        return BLE_HS_EUNKNOWN;
```
andrzej-kaczmarek, line 434: einval

nimble/host/src/ble_sm_alg.c
```diff
@@ -418,6 +418,139 @@ ble_sm_alg_g2(const uint8_t *u, const uint8_t *v, const uint8_t *x,
...
+ble_sm_alg_csis_s1(const uint8_t *m, size_t m_len, uint8_t *out)
+{
+    int rc;
+    uint8_t k_zero[16] = {0};
+
+    /* XXX: Spec does not specify the maximum M parameter length.
+     * We assume that 16 bytes is enough and return error if passed len value is greater
+     * than that */
+    if (m_len > 16) {
+        return BLE_HS_EUNKNOWN;
```
andrzej-kaczmarek, line 467: einval

andrzej-kaczmarek (changes requested):
looks good, just few style comments

andrzej-kaczmarek (approved):

PR #1639, opened 2023-10-12
nimble/host: Add option to build with Mbed-TLS

kasjer:
So far TinyCRYPT library was used in SM.
Now it is possible to build it with Mbed-TLS.

### DIFF
--- a/nimble/host/test/syscfg.yml
+++ b/nimble/host/test/syscfg.yml
@@ -32,3 +32,4 @@ syscfg.vals:
     BLE_L2CAP_ENHANCED_COC: 1
     BLE_TRANSPORT_LL: custom
     BLE_EATT_CHAN_NUM: 0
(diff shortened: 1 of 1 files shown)


nimble/host/syscfg.yml
```diff
@@ -101,6 +101,10 @@ syscfg.defs:
         description: 'Security manager secure connections (4.2).'
         value: 0
 
+    BLE_SM_USE_TINYCRYPT:
```
sjanc, line 104: I'd make this choice option

BLE_SM_CRYPTO:
    value: tinycrypt
    choices:
        - tinycrypt
        - mbedtls

Although mbedtls seems to be bigger than tinycrypt it may stil be preferred if application is already using it for other reasons

rymanluk: @kasjer @sjanc what plan we have about this?

PR #1641, opened 2023-10-16
nimble/ll: solve issue with duplicated ISO flags

KKopyscinski:
This patch makes MYNEWT_VAL flags consistent: BLE_LL_CFG_FEAT_LL_ISO defines same functionality as BLE_LL_ISO, and the latter shall be used.

Removed duplicated case in ble_ll_hci_le_cmd_proc for BLE_HCI_OCF_LE_READ_ISO_TX_SYNC OP code.


KKopyscinski:
@andrzej-kaczmarek as discussed offline, I removed code that caused errors. I also checked LL ISO command structs in `hci_common.h` and they are ok.

andrzej-kaczmarek (approved):

PR #1646, opened 2023-10-23
nimble: add sysconfig for enabling additional PHYs

KKopyscinski:
Three new configs are introduced: BLE_ADDITIONAL_PHY that enables additional PHY and two configs that specify what additional PHY shall be enabled (2M or CODED)


nimble/syscfg.yml
```diff
@@ -130,6 +130,25 @@ syscfg.defs:
             This enables LE Connection Subrating feature
         value: 0
 
+    BLE_ADDITIONAL_PHY:
```
sjanc, line 133: I'd name it simply 'BLE_PHY'

nimble/syscfg.yml
```diff
@@ -130,6 +130,25 @@ syscfg.defs:
...
+        description: >
+            This enables support for addtitional PHY (2M or CODED)
+        value: 0
+
+    BLE_2M_PHY:
```
sjanc, line 138: BLE_PHY_2M

nimble/syscfg.yml
```diff
@@ -130,6 +130,25 @@ syscfg.defs:
...
+        description: >
+            This enables support for addtitional 2M PHY
+        value: 0
+        restrictions:
+            - 'BLE_ADDITIONAL_PHY if 1'
+
+    BLE_CODED_PHY:
```
sjanc, line 145: BLE_PHY_CODED

sjanc (commented):
and we should have same for host, ie
BLE_HS_PHY
BLE_HS_PHY_2M etc

nimble/syscfg.yml
andrzej-kaczmarek: why do we need separate syscfg? just enable 2m or coded or both.

KKopyscinski: It's because it may be simplier to cut down code if neither is supported. So when you check something besides 1M is supported you do
```c
#if MYNEWT_VAL(BLE_PHY)
```
instead of 
```c
#if MYNEWT_VAL(BLE_PHY_2M) || MYNEWT_VAL(BLE_PHY_CODED)
```

andrzej-kaczmarek:
```
#define BLE_PHY_ENABLED (MYNEWT_VAL(BLE_PHY_2M) || MYNEWT_VAL(BLE_PHY_CODED))

...

#if BLE_PHY_ENABLED
#endif
```

alternatively smth like this should work

```
syscfg.defs.'BLE_PHY_2M || BLE_PHY_CODED':
    BLE_PHY: 1
```

we should avoid syscfg vals that do nothing except they have to be set because another syscfg is set

KKopyscinski: Makes sense to me, I updated PR with second option. Config generates OK

andrzej-kaczmarek (approved):
if you have some time to spare, you can also add `BLE_LL_PHY` and change ll to use it consistently instead of `BLE_LL_BT5_PHY_SUPPORTED`

KKopyscinski:
> if you have some time to spare, you can also add `BLE_LL_PHY` and change ll to use it consistently instead of `BLE_LL_BT5_PHY_SUPPORTED`

Added second commit

nimble/controller/include/controller/ble_ll_conn.h
```diff
@@ -233,7 +233,7 @@ struct ble_ll_conn_sm
     uint16_t host_req_max_rx_time;
 #endif
 
-#if (BLE_LL_BT5_PHY_SUPPORTED == 1)
+#if (MYNEWT_VAL(BLE_LL_PHY))
```
andrzej-kaczmarek, line 236: remove extra parentheses

KKopyscinski: done

sjanc (approved):

PR #1647, opened 2023-10-23
nimble/transport: Add common HCI RX task

andrzej-kaczmarek:
This adds common task for implementing HCI RX. Any transport can simply
register its own function for RX that will be run in separate task
whenever triggered by dedicated API. This for example allows to easily
implement handling HCI RX in a task instead of in an interrupt since
handling RX in an interrupt doesn't work well with monitor enabled.

If selected transport doesn't register its function, all rx_task code
will be compiled out since there won't be any references to that code.

### DIFF
--- a/nimble/transport/include/nimble/transport.h
+++ b/nimble/transport/include/nimble/transport.h
@@ -35,6 +35,11 @@ struct os_mbuf;
 /* Initialization */
 void ble_transport_init(void);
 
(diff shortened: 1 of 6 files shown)


sjanc (approved):

rymanluk:
is it still needed? @andrzej-kaczmarek

PR #1653, opened 2023-11-20
controller: fix ISO compilation errors when build for Broadcast only, and for nRF52

KKopyscinski:


nimble/controller/src/ble_ll_hci.c
```diff
@@ -1503,7 +1503,8 @@ ble_ll_hci_ctlr_bb_cmd_proc(const uint8_t *cmdbuf, uint8_t len, uint16_t ocf,
     case BLE_HCI_OCF_CB_SET_EVENT_MASK2:
         rc = ble_ll_hci_cb_set_event_mask2(cmdbuf, len);
         break;
-#if MYNEWT_VAL(BLE_LL_CFG_FEAT_LE_PING)
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_ENHANCED_CONN_UPDATE) && \
```
andrzej-kaczmarek, line 1506: this is not related to le ping so whatever it fixes, the problem is somewhere else

KKopyscinski: right, I think I misread ifs. I think it should be only for connection. So
```c
#if (MYNEWT_VAL(BLE_LL_ROLE_PERIPHERAL) || MYNEWT_VAL(BLE_LL_ROLE_CENTRAL)) \
     && MYNEWT_VAL(BLE_LL_CFG_FEAT_LE_PING)
```
instead

KKopyscinski:
@andrzej-kaczmarek Waiting for review

andrzej-kaczmarek (approved):

PR #1661, opened 2023-12-12
nimble/gatts: Modify client supported features READ

Roshan23699:
1. Fix read on client-supported features. 
2. Read cannot return the local client-supported features, it should return the client-supported features for the reading connection. (This change is needed for PTS).

### DIFF
--- a/nimble/host/include/host/ble_gatt.h
+++ b/nimble/host/include/host/ble_gatt.h
@@ -1096,6 +1096,25 @@ int ble_gatts_start(void);
 int ble_gatts_peer_cl_sup_feat_update(uint16_t conn_handle,
                                       struct os_mbuf *om);
 
+/**
+ * Gets Client Supported Features for specified connection.
+ *
+ * @param conn_handle           Connection handle identifying connection for
+ *                              which Client Supported Features should be saved
+ * @param out_supported_feat    Client supported features to be returned.
+ *
+ * @return                      0 on success;
+ *                              BLE_HS_ENOTCONN if no matching connection
+ *                              was found
+ *                              BLE_HS_EINVAL if supplied buffer is empty or
+ *                              if any Client Supported Feature was
+ *                              attempted to be disabled.
+ *                              A BLE host core return code on unexpected
+ *                              error.
+ *
+ */
+int ble_gatts_peer_cl_sup_feat_get(uint16_t conn_handle, uint8_t *out_supported_feat, uint8_t len);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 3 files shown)


Roshan23699:
@sjanc could you PTAL ?

sjanc:
Hi,

which test was affected by this?

Roshan23699:
**GATT/SR/GAS/BV-04-C**
Maintaining Client-Supported Features Characteristic Values for Bonded Devices.

nimble/host/include/host/ble_gatt.h
```diff
@@ -1096,6 +1096,25 @@ int ble_gatts_start(void);
...
+ *                              BLE_HS_ENOTCONN if no matching connection
+ *                              was found
+ *                              BLE_HS_EINVAL if supplied buffer is empty or
+ *                              if any Client Supported Feature was
+ *                              attempted to be disabled.
+ *                              A BLE host core return code on unexpected
+ *                              error.
+ *
+ */
+int ble_gatts_get_peer_cl_sup_feat(uint16_t conn_handle, uint8_t *out_supported_feat);
```
sjanc, line 1116: please name it  ble_gatts_peer_cl_sup_feat_get(),  also there should be len parameter for how many bytes are available in out_supported_feat array

nimble/host/src/ble_gatts.c
```diff
@@ -1580,6 +1580,31 @@ ble_gatts_chr_updated(uint16_t chr_val_handle)
...
+    }
+
+    ble_hs_lock();
+    conn = ble_hs_conn_find(conn_handle);
+    if (conn == NULL) {
+        rc = BLE_HS_ENOTCONN;
+        goto done;
+    }
+
+    memcpy(out_supported_feat, conn->bhc_gatt_svr.peer_cl_sup_feat,
```
sjanc, line 1600: this should copy no more than N bytes of mentioned length parameter

KKopyscinski: > this should copy no more than N bytes of mentioned length parameter

and maybe N should be checked as `N < BLE_GATT_CHR_CLI_SUP_FEAT_SZ`?

sjanc: I'm fine with either,   if buffer is smaller we can fill just up to its size and ignore rest,  currently only 3 bits are defined so it will be 1 for some time anyway

nimble/host/include/host/ble_gatt.h
Roshan23699: Done.

nimble/host/src/ble_gatts.c
Roshan23699: Done.

sjanc (approved):

PR #1666, opened 2023-12-21
Fix HCI Set Event Mask Page 2 support in host and controller

donatieng:
This patch adds a new function, `ble_hs_startup_read_sup_cmd_tx()`, which reads the list of supported HCI commands from the controller, which is used on HCI startup. This is used to determine whether a HCI Set Event Mask Page 2 command should be sent to the controller.

The second commit adds HCI Set Event Mask Page 2 to the list of supported
commands. This command is already implemented in the controller but
was not listed as supported.

This is my first Nimble PR, so any guidance is appreciated.


donatieng:
Issues (host tests and style) should now be fixed

sjanc (approved):
In future we may need to get more generic API to test for supported commands , but this is good enough for now

PR #1668, opened 2023-12-29
Feature: Support to ask user for authorization

Singh-Sumeet:
This MR aims to add:
1. Support to ask user for authorization in case of GATT read or write operations.
2. Relevant event handling in bleprph app.


Singh-Sumeet:
@sjanc Please TAL

nimble/host/include/host/ble_gap.h
```diff
@@ -258,6 +258,9 @@ struct hci_conn_update;
 /** GAP event: BIG (Broadcast Isochronous Group) information report */
 #define BLE_GAP_EVENT_BIGINFO_REPORT        30
 
+/** GAP event: Authorization request for GATT operations */
+#define BLE_GAP_EVENT_AUTHORIZE             30
```
andrzej-kaczmarek, line 262: duplicated id

nimble/host/include/host/ble_gap.h
```diff
@@ -292,6 +295,19 @@ struct hci_conn_update;
...
+/**
+ * @defgroup GAP Authorize event possible responses.
+ * @{
+ */
+
+/** GAP Authorize event response: accept */
+#define BLE_GAP_AUTHORIZE_ACCEPT            1
+
+/** GAP Authorize event response: reject */
+#define BLE_GAP_AUTHORIZE_REJECT            2
```
andrzej-kaczmarek, line 307: authorization can be done using return value from event callback

andrzej-kaczmarek (changes requested):

nimble/host/include/host/ble_gap.h
Singh-Sumeet: Changed to 31

nimble/host/src/ble_gap.c
```diff
@@ -6521,6 +6521,33 @@ ble_gap_unhandled_hci_event(bool is_le_meta, bool is_vs, const void *buf,
...
+#if MYNEWT_VAL(BLE_ROLE_PERIPHERAL)
+    struct ble_gap_event event;
+
+    memset(&event, 0, sizeof event);
+    event.type = BLE_GAP_EVENT_AUTHORIZE;
+    event.authorize.conn_handle = conn_handle;
+    event.authorize.attr_handle = attr_handle;
+    event.authorize.is_read = is_read;
+
+    ble_gap_event_listener_call(&event);
```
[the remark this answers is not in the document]
Singh-Sumeet: Removed this call

nimble/host/src/ble_gap.c
```diff
@@ -6458,7 +6458,7 @@ ble_gap_subscribe_event(uint16_t conn_handle, uint16_t attr_handle,
                         uint8_t prev_notify, uint8_t cur_notify,
                         uint8_t prev_indicate, uint8_t cur_indicate)
 {
-#if NIMBLE_BLE_CONNECT
+#if MYNEWT_VAL(BLE_ROLE_PERIPHERAL)
```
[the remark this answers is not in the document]
Singh-Sumeet: Reverted to original.

nimble/host/include/host/ble_gap.h
Singh-Sumeet: I see two problems with that.

1. The return values are normally used for error codes, I guess maybe returning BLE_HS_EREJECT will work for rejecting.
2. The default return code in gap_event in the apps is 0. So if the user does not handle the event, the app will return 0, which will be interpreted as "ACCEPT".

Singh-Sumeet:
@sjanc PTAL

nimble/host/include/host/ble_gap.h
andrzej-kaczmarek: 1. just return non-zero and this means reject
2. that's expected behavior, otherwise apps that do not have this event handled would reject all requests which is not what we want

andrzej-kaczmarek:
@Singh-Sumeet we discussed with @sjanc and we agreed that return value from callback should be used instead of event struct field. we'd also like to have an option in future to make it authorization asynchronous so the return values should be reject, accept, pending - please add appropriate macros for that. reject should be set to 0 so authorization is rejected by default.

Singh-Sumeet:
Hi @andrzej-kaczmarek , I have made the required changes. I don't understand what you mean by making it "Authorization Asynchronous", but I have created the macro nonetheless.

PR #1671, opened 2024-01-02
nimble/host: Fix possible overflow in ble_gatts_peer_cl_sup_feat_update

sjanc:
This function would write pass conn->bhc_gatt_svr.peer_cl_sup_feat buffer. Since supported features are likely to be small (currently only 3 bits are defined) lets keep this simple and just make local copy before validation.


sjanc:
@Roshan23699 FYI

KKopyscinski (approved):
LGTM

PR #1672, opened 2024-01-05
Signed Write support

Singh-Sumeet:
This PR aims to:
1. Add support for sending signed write commands
2. Add support of receiving signed write commands


Singh-Sumeet:
@sjanc PTAL

sjanc:
Hi,

Few general things (I'll have more detailed comments after doing full code review)
1) add support for this in bttester (and maybe btshell)
2) test it with PTS   (eg via autopts (https://github.com/auto-pts/auto-pts))
3) Commit message (title and body) needs improvements
4) Maybe this should gracefully fallback to Write Without Response if link is encrypted? (just a thought, I'm fine with returning error as is now)

In general I'm bit reluctant about adding this feature as it is barely used (TBH I've never seen this being used in real applications and have not seen any profile that would require it).

But if you need it and above steps are completed I'm not going to block it

PR #1673, opened 2024-01-16
Fix remove from resolving list issue

szymon-czapracki:
IRK of the peer might not be present on the resolving list, but we still should be able to remove that entry.
Previously this operation would produce an error, and removal was unsuccessful.


andrzej-kaczmarek:
issue description is missing in commit message

sjanc (changes requested):
describe problem and fix in commit message

sjanc (approved):

PR #1676, opened 2024-01-22
host/ble_hs: Add public API for getting the host eventq

vrahane:
- This is needed if we want to perform other operations on the host eventq
- In our specific case we would like to perform all BLE comms (App level encoding/decoding) related work on this eventq and did not have a way to get it


vrahane:
I don't need this anymore, closing it

PR #1677, opened 2024-01-22
nimble/audio: Add Audio Broadcast Sink initial implementation

MariuszSkamra:
Depends on: https://github.com/apache/mynewt-nimble/pull/1691


nimble/host/audio/src/ble_audio_bsnk.c
```diff
@@ -0,0 +1,724 @@
...
+#include "host/ble_gap.h"
+#include "host/ble_hs.h"
+#include "host/ble_iso.h"
+#include "host/ble_uuid.h"
+#include "host/audio/ble_audio.h"
+
+#include "ble_audio_priv.h"
+
+#if MYNEWT_VAL(BLE_ISO_BROADCAST_SINK)
+struct ble_audio_bsnk {
```
KKopyscinski, line 32: I think this struct should be public not opaque, so moved fully to `ble_audio_bsnk.h`, because higher layers (like BASS) may want to access information contained there. Or at least moved to `ble_audio_priv.h`, but including priv headers in services is better to be avoided

KKopyscinski: Also, I think this struct should also contain `Num_Subgroups`, `BIS_Sync_State` bifield and `Metadata` with length, for every subgroup as described in BASS 3.2.1.8-11, which would be filled on sync established. Maybe something like this:

```c
struct ble_audio_bsnk_soubgroup_entry {
      uint32_t bis_sync_state;
      uint8_t metadata_len;
      uint8_t *metadata;
}

struct ble_audio_bsnk {
    /** Advertiser Address for the Broadcast Source */
    ble_addr_t adv_addr;

    /** Advertising Set ID */
    uint8_t adv_sid;
    .
    .
    .
    uint8_t num_subgroups;

    STAILQ_HEAD(, ble_audio_bsnk_soubgroup_entry) subgroups;

    /** Singly-linked list entry. */
    SLIST_ENTRY(ble_audio_bsnk) next;
};
```
and I think Broadcast Sink should handle memory for `ble_audio_bsnk_soubgroup_entry` and metadata

KKopyscinski: Discussed offline, I think that these can stat as is, as for BASS reading this from ble_audio_bsnk is ok, but writing would be troublesome. BASS will, sadly, keep copy of the BSNK state that can be updated from BASS control point characteristic

nimble/host/src/ble_hs_hci_evt.c
```diff
@@ -65,10 +65,14 @@ static ble_hs_hci_evt_le_fn ble_hs_hci_evt_le_periodic_adv_rpt;
 static ble_hs_hci_evt_le_fn ble_hs_hci_evt_le_periodic_adv_sync_lost;
 static ble_hs_hci_evt_le_fn ble_hs_hci_evt_le_scan_req_rcvd;
 static ble_hs_hci_evt_le_fn ble_hs_hci_evt_le_periodic_adv_sync_transfer;
-#if MYNEWT_VAL(BLE_ISO_BROADCASTER)
+#if MYNEWT_VAL(BLE_ISO_BROADCAST_SOURCE)
 static ble_hs_hci_evt_le_fn ble_hs_hci_evt_le_create_big_complete;
 static ble_hs_hci_evt_le_fn ble_hs_hci_evt_le_terminate_big_complete;
 #endif
+#if MYNEWT_VAL(BLE_ISO_BROADCAST_SINK)
+static ble_hs_hci_evt_le_fn ble_hs_hci_evt_le_big_sync_extablished;
```
KKopyscinski, line 73:
```suggestion
static ble_hs_hci_evt_le_fn ble_hs_hci_evt_le_big_sync_established;
```

nimble/host/audio/include/host/audio/ble_audio.h
```diff
@@ -0,0 +1,413 @@
...
+/** @brief Audio Broadcast Sink Sync state changed event */
+struct ble_audio_event_bsnk_sync_state_changed {
+    /** Audio Broadcast Sink instance */
+    struct ble_audio_bsnk *snk;
+
+    /** Audio Broadcast Sync state */
+    enum ble_audio_bsnk_sync_state state;
+
+    /** BIS Synchronization */
+    uint32_t bis_sync;
```
KKopyscinski, line 148: I think this should be an array - probably pointer with count of entries (subgroups)

apps/btshell/src/cmd_leaudio.c
```diff
@@ -382,4 +384,479 @@ cmd_leaudio_broadcast_stop(int argc, char **argv)
...
+        return rc;
+    }
+
+    instance_id = parse_arg_uint8("instance_id", &rc);
+    if (rc != 0) {
+        console_printf("invalid instance_id parameter\n");
+        return rc;
+    }
+
+    bis_sync_mask = parse_arg_uint16("bis_sync_mask", &rc);
```
MariuszSkamra, line 784: parse_arg_uint32

nimble/host/audio/include/host/audio/ble_audio.h
MariuszSkamra: done

nimble/host/audio/include/audio/ble_audio.h
```diff
@@ -536,17 +536,8 @@ struct ble_audio_broadcast_name {
 /** BLE Audio event: BASS - Remote Scan Started */
 #define BLE_AUDIO_EVENT_BASS_REMOTE_SCAN_STARTED             4
 
-/** BLE Audio event: BASS - Add Source */
```
KKopyscinski: What's the reason to remove these events? I guess the upper layer knows that some actions were performed because the action callback will be called, but this callback is not mandatory. So if the operation was accepted autonomously, there will be no feedback for upper layer/app. For example, we could never enter here:
https://github.com/apache/mynewt-nimble/pull/1677/commits/e5a346bdc39590f58896fdbf25262088cde34480#diff-b8c009ce2010b2473424d883cd93e8fe365b1c1af565ad8ba875280fd069133cR324

nimble/host/audio/include/audio/ble_audio_broadcast_sink.h
```diff
@@ -0,0 +1,267 @@
...
+            uint8_t source_id;
+
+            /** ISO Interval. */
+            uint16_t iso_interval;
+
+            /** Presentation delay. */
+            uint32_t presentation_delay;
+
+            /** Pointer to Maximum Subevents value to initialize. */
+            uint8_t *out_mse;
```
rymanluk, line 112: We should add NSE and BN here so application has any means to choose out_MSE.

Also we should set default to 0x00 which means controller can choose up to NSE

nimble/host/audio/include/audio/ble_audio_broadcast_sink.h
```diff
@@ -0,0 +1,267 @@
...
+
+            /** Pointer to Sync Timeout value to initialize. */
+            uint16_t *out_sync_timeout;
```
rymanluk, line 115: Lets add comment about minimum and also set default here:

```
BIG_Sync_Timeout set by the Host is less than 6 × ISO_Interval, the
Controller shall set the timeout to 6 × ISO_Interval
```

MariuszSkamra: done

nimble/host/audio/include/audio/ble_audio_broadcast_sink.h
MariuszSkamra: done

nimble/host/audio/include/audio/ble_audio.h
MariuszSkamra: not applicable

nimble/host/src/ble_hs_hci_evt.c
MariuszSkamra: done

MariuszSkamra:
Fixed missing `ble_audio_scan_delegator_receive_state_set` call in `ble_audio_scan_delegator_receive_state_set`

nimble/host/audio/src/ble_audio_broadcast_sink.c
```diff
@@ -0,0 +1,1647 @@
...
+    qsort(bis_params, bis_cnt, sizeof(*bis_params), bis_params_bis_index_cmp);
+
+    big_sync_create_params.bis_cnt = bis_cnt;
+
+    /* Stop scan first */
+    disc_stop();
+
+    rc = ble_iso_big_sync_create(&big_sync_create_params, &sink->big_handle);
+    if (rc != 0) {
+        BLE_HS_LOG_ERROR("big sync failed (%d)\n", rc);
```
szymon-czapracki, line 1047: Shouldn't there also be `big_sync_state_set(sink, BIG_SYNC_STGATE_FAILED)` ?

rymanluk (approved):
LGTM

PR #1680, opened 2024-01-25
host/gap: fix typo in GAP event macro

wpiet:
Corrects spelling in BLE_GAP_EVENT_PAIRING_COMPLETE


sjanc (changes requested):
unfortunately this is part of public API so we need to keep this backward compatible
so please add
#define BLE_GAP_EVENT_PARING_COMPLETE BLE_GAP_EVENT_PAIRING_COMPLETE
with a comment that PARING is deprecated

sjanc (commented):
looks ok but please also update apps (afair it is used only in bttester)

sjanc (approved):

PR #1681, opened 2024-01-26
nimble/host: ble_gap_adv_get_free_instance shall return non-negative value

KKopyscinski:
All public GAP API functions shall return non-negative error codes. Now, function fills out pointer with found advertising instance and returns correct error code.


KKopyscinski:
PTAL @wpiet

wpiet (approved):
LGTM

sjanc (approved):

PR #1685, opened 2024-02-07
nimble/host: Add API to allow given address from whitelist

rahult-github:
This patch adds a  API which allows host to remove a given address from whitelist.

Current existing APIs clear complete whitelist. This API provides flexibility to application to manage whitelist address by deleting a particular address from whitelist.


andrzej-kaczmarek:
Existing API allows to set complete whitelist which means app manages whitelist contents on its own (add/remove/clear/whatever) and then calls that API to set whitelist. There's no need to have separate add/remove/clear API in such design and I'm reluctant to add new APIs which are redundant.

In any case, the name of API is incorrect (should be `ble_gap_wl_remove`) and there should be also `add` and `clear` APIs, if we decide to add it anyway.

rahult-github:
@andrzej-kaczmarek  , ok . Thanks for explaining the architecture design. Closing the PR

PR #1687, opened 2024-02-07
nimble/host: Zero initialize buffer after allocation

rahult-github:
At many places , after getting buffer using os_memblock_get, buffer is not zero initiatlized. Added change to memset the same to 0.


rahult-github:
@andrzej-kaczmarek , @sjanc  , please take a look

sjanc:
Hi,

Does this fix any use of uninitialized memory or is just a precaution?

As for style, please use sizeof(*foo)  syntax

rahult-github:
> Hi,
> 
> Does this fix any use of uninitialized memory or is just a precaution?
> 
> As for style, please use sizeof(*foo) syntax

Currently , just precaution, since i didn't find any issue as such during my testing. But in many places in nimble, such memset is already existing, hence added at places where it is missing. 

For style, sure, will change and re-push

rahult-github:
Hello @sjanc  , please take a look

sjanc (commented):
I'm not sure about this patch,     most are memseting places that already initialize all fields afterwards, and some of them are frequently called, add extra overhead o function call for no little benefit

maybe we should memset (or explicitly initialized members) only where it is really needed?

PR #1693, opened 2024-02-13
fix(nimble): Change 'move_count' to uint8_t type

darshan7patel:
- Change 'move_count' to uint8_t type to prevent out-of-bounds errors during memmove with compiler optimization enabled.


andrzej-kaczmarek:
please add proper prefix to commit title (see other commits as guideline) and include error message in commit message

sjanc:
I'm having trouble understanding this fix,    this basically limit move_count to 255

what is exactly fixed here?

sjanc:
okay, so I looked at this and it appears that GCC is not able to properly track relations between globals (eg ble_store_config_num_peer_secs) and memmove in ble_store_config_delete_obj

sjanc:
I'd suggest to use assert() or if() to hint GCC

BTW BLE_STORE_MAX_CCCDS() might have similar issues  (if also configured to 1)

darshan7patel:
Basically, the size has been increased since, when compiler optimization was enabled, it was previously creating out-of-bounds errors. So, now it will prevent these errors.

sjanc:
Hi,

please check if https://github.com/apache/mynewt-nimble/pull/1737 solves this issue for you

PR #1699, opened 2024-02-19
host/ble_audio_broadcast: add clarification for big_params pointer

KKopyscinski:
Added clarification for `big_params` pointer in
`ble_broadcast_create_params`.


MariuszSkamra:
LGTM

PR #1705, opened 2024-02-21
nimble/controller: Enable LL encryption if BLE_LL_ISO_BROADCASTER

MariuszSkamra:
To support encrypting/decrypting broadcasts, LL Encryption shall be enabled by BLE_LL_ISO_BROADCASTER.


MariuszSkamra:
@andrzej-kaczmarek Please take a look

rymanluk (approved):
If it does not build without this flag it makes sens to set it.
However,  I can imagine broadcaster using no encryption and would like to not waste memory for that.

Suggest to create an issue to allow Broadcaster without encryption.

For now this LGTM

MariuszSkamra:
> Suggest to create an issue to allow Broadcaster without encryption.

 Done, https://github.com/apache/mynewt-nimble/issues/1713

PR #1708, opened 2024-02-21
nimble/audio: Add LE Audio event listener and BASE parser

MariuszSkamra:
This adds dedicated event listener for LE Audio events and BASE parser implementation along with unit tests.

### DIFF
--- a/nimble/host/audio/include/host/audio/ble_audio.h
+++ b/nimble/host/audio/include/host/audio/ble_audio.h
@@ -21,8 +21,146 @@
 #define H_BLE_AUDIO_
 
 #include <stdint.h>
+#include <sys/queue.h>
 
 #include "host/ble_audio_common.h"
+
+/** @brief Public Broadcast Announcement features bits */
+enum ble_audio_pub_bcst_announcement_feat {
+    /** Broadcast Stream Encryption */
+    BLE_AUDIO_PUB_BCST_ANNOUNCEMENT_FEAT_ENCRYPTION = 1 << 0,
+
+    /** Standard Quality Public Broadcast Audio */
+    BLE_AUDIO_PUB_BCST_ANNOUNCEMENT_FEAT_SQ = 1 << 1,
+
+    /** High Quality Public Broadcast Audio */
+    BLE_AUDIO_PUB_BCST_ANNOUNCEMENT_FEAT_HQ = 1 << 2,
+};
+
+/** @brief Public Broadcast Announcement structure */
+struct ble_audio_pub_bcst_announcement {
+    /** Public Broadcast Announcement features bitfield */
+    enum ble_audio_pub_bcst_announcement_feat features;
+
+    /** Metadata length */
+    uint8_t metadata_len;
+
+    /** Metadata */
+    const uint8_t *metadata;
+};
+
+struct ble_audio_bcst_name {
+    /** Broadcast Name length */
+    uint8_t name_len;
+
+    /** Broadcast Name */
+    const char *name;
+};
+
+/**
+ * @defgroup ble_audio_events Bluetooth Low Energy Audio Events
+ * @{
+ */
+
+/** BLE Audio event: Broadcast Announcement */
+#define BLE_AUDIO_EVENT_BCST_ANNOUNCEMENT               0
+
+/** @} */
+
+/** @brief Broadcast Announcement */
+struct ble_audio_event_bcst_announcement {
+    /** Extended advertising report */
+    const struct ble_gap_ext_disc_desc *ext_disc;
+
+    /** Broadcast ID */
+    uint32_t broadcast_id;
+
+    /** Additional service data included in Broadcast Audio Announcement */
+    const uint8_t *svc_data;
+
+    /** Additional service data length  */
+    uint16_t svc_data_len;
+
+    /** Optional Public Broadcast Announcement data */
+    struct ble_audio_pub_bcst_announcement *pub_announcement_data;
+
+    /** Optional Broadcast Name */
+    struct ble_audio_bcst_name *name;
+};
+
+/**
+ * Represents a BLE Audio related event. When such an event occurs, the host
+ * notifies the application by passing an instance of this structure to an
+ * application-specified callback.
+ */
+struct ble_audio_event {
+    /**
+     * Indicates the type of BLE Audio event that occurred. This is one of the
+     * BLE_AUDIO_EVENT codes.
+     */
+    uint8_t type;
+
+    /**
+     * A discriminated union containing additional details concerning the event.
+     * The 'type' field indicates which member of the union is valid.
+     */
+    union {
+        /**
+         * @ref BLE_AUDIO_EVENT_BCST_ANNOUNCEMENT
+         *
+         * Represents a received Broadcast Announcement.
+         */
+        struct ble_audio_event_bcst_announcement bcst_announcement;
+    };
+};
+
+/** Callback function type for handling BLE Audio events. */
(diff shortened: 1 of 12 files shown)


MariuszSkamra:
The CI is failing with false negative result

nimble/host/audio/src/ble_audio.c
```diff
@@ -0,0 +1,524 @@
...
+            break;
+        }
+
+        uuid16.value = get_le16(&field->value[offset]);
+        offset += 2;
+
+        switch (uuid16.value) {
+        case BLE_BROADCAST_AUDIO_ANNOUNCEMENT_SVC_UUID:
+            if ((value_len - offset) < 3) {
+                /* Stop parsing */
```
KKopyscinski, line 63: As far as I can see, if we ever enter here, it means that this field is malformed. What if previous one wasn't, and `data->success` is already set to `true`? I'm not sure we want to report malformed advertising reports, maybe we should set `data->success = false` before we return, here and in later checks?

nimble/host/audio/src/ble_audio.c
```diff
@@ -0,0 +1,524 @@
...
+    if (listener == NULL) {
+        BLE_HS_LOG_ERROR("NULL listener!\n");
+        return BLE_HS_EINVAL;
+    }
+
+    /*
+     * We check if element exists on the list only for sanity to let caller
+     * know whether it registered its listener before.
+     */
+
```
KKopyscinski, line 218:
```suggestion
```
probably not needed empty line

nimble/host/audio/src/ble_audio.c
```diff
@@ -0,0 +1,524 @@
...
+    group->presentation_delay = get_le24(data);
+    offset += 3;
+
+    group->num_subgroups = data[offset];
+    offset++;
+
+    if (group->num_subgroups < 1) {
+        BLE_HS_LOG_DEBUG("Rule 1 violation: There shall be at least one subgroup.\n");
+    }
```
KKopyscinski, line 314: We know the rules, but it imho looks a bit weird :) And don't we want to return early here, like in other errors? I think BASE with no subgroups as just as bad as empty one - both don't follow spec. Maybe something like this?
```suggestion
    if (group->num_subgroups < 1) {
        BLE_HS_LOG_ERROR("Invalid BASE: no subgroups!\n");
        return BLE_HS_EINVAL;
    }
```

nimble/host/audio/syscfg.yml
```diff
@@ -0,0 +1,21 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+syscfg.defs:
+
+syscfg.logs:
```
KKopyscinski, line 21: I think we don't need syscfg.logs here. And under defs, I'd move BLE Audio configs from `host/syscfg.yml`: `BLE_MAX_BIG` and `BLE_MAX_BIS`. And maybe ISO related configs too? I mean `BLE_ISO`, `BLE_ISO_BROADCAST_SOURCE` and so on. What do you think @andrzej-kaczmarek ? This way we'll limit bloating of host syscfg

nimble/host/include/host/ble_audio_common.h
```diff
@@ -47,6 +47,7 @@
                                                       __VA_ARGS__)
 
 #define BLE_BROADCAST_AUDIO_ANNOUNCEMENT_SVC_UUID                 0x1852
+#define BLE_BROADCAST_PUBLIC_BROADCAST_ANNOUNCEMENT_SVC_UUID      0x1856
```
KKopyscinski, line 50: Shorter = better?
```suggestion
#define BLE_BROADCAST_PUB_BCST_ANNOUNCEMENT_SVC_UUID      0x1856
```
and probably doxygen comment with full name.

KKopyscinski (changes requested):

nimble/host/audio/include/host/audio/ble_audio.h
```diff
@@ -0,0 +1,266 @@
...
+ * under the License.
+ */
+
+#ifndef H_BLE_AUDIO_
+#define H_BLE_AUDIO_
+
+#include <stdint.h>
+#include <sys/queue.h>
+
+#include "host/ble_audio_common.h"
```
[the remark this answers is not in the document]
MariuszSkamra: I agree, but I would like this PR to be merged first. Then we can move the existing code to `host/audio/` as you suggested in subsequent PR.

nimble/host/audio/src/ble_audio.c
MariuszSkamra: Yeah, I was considering that, probably you're right.
I was thinking about this as an Advertising data that display to the user - when you parse e.g. Complete Name, Service Data and Appearance. When you fail to parse Service data I think you still want to report the Complete Name and Appearance.
Let's wait for 3rd eye pair to look at this.

nimble/host/audio/syscfg.yml
MariuszSkamra: Once the code will be moved to `host/syscfg.yml` in subsequent PR, I'll place the configs as you suggested.

nimble/host/audio/include/host/audio/ble_audio.h
```diff
@@ -0,0 +1,266 @@
...
+
+/** @brief Public Broadcast Announcement features bits */
+enum ble_audio_bcst_public_feature {
```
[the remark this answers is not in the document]
MariuszSkamra: I'll got with `ble_audio_pub_bcst_announcement`

MariuszSkamra:
Rebased

nimble/host/audio/include/host/audio/ble_audio.h
```diff
@@ -0,0 +1,128 @@
...
+
+#ifndef H_BLE_AUDIO_
+#define H_BLE_AUDIO_
+
+#include <stdint.h>
+
+#include "host/ble_audio_common.h"
+
+/**
+ * BASE iterator
```
rymanluk, line 166: could you add here a comment explaining the common usage of this API ?

nimble/host/audio/src/ble_audio.c
rymanluk: I suggest BLE_HS_LOG_ERROR here and return as @KKopyscinski  suggested.

rymanluk: I like @MariuszSkamra  error message better, so just change DEBUG to ERROR here.

nimble/host/audio/src/ble_audio.c
rymanluk: agree with @KKopyscinski

nimble/host/audio/src/ble_audio.c
```diff
@@ -23,6 +23,238 @@
...
+        return BLE_HS_EINVAL;
+    }
+
+    SLIST_FOREACH(evl, &ble_audio_event_listener_list, next) {
+        if (evl == listener) {
+            break;
+        }
+    }
+
+    if (!evl) {
```
rymanluk, line 187: maybe to avoid nesting,

if (evl) {
   return BLE_HS_EALREADY;
}

nimble/host/audio/src/ble_audio.c
MariuszSkamra: I agree as well, and I already changed that

KKopyscinski (approved):

rymanluk (approved):
LGTM

PR #1712, opened 2024-02-27
ci: Add Doxygen Check workflow

wpiet:
This commit adds a new workflow for Doxygen Check to ensure documentation integrity. The workflow verifies if the Doxygen output contains any warnings.

### DIFF
--- a/.github/check_doxygen.py
+++ b/.github/check_doxygen.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python3
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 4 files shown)


.github/check_doxygen.py
```diff
@@ -0,0 +1,112 @@
...
+    out = subprocess.run(cmd, text=True, shell=True, stdout=subprocess.PIPE,
+                         stderr=subprocess.DEVNULL).stdout
+    return out.splitlines()
+
+
+def is_ignored(filename: str) -> bool:
+    if Path(filename).suffix != ".h" or "priv" in filename:
+        return True
+    if filename == "nimble/include/nimble/ble.h":
+        return False
```
MariuszSkamra, line 47: Why this particular file is about to be ignored?

Consider having a list of ignored files that can be easily extended in the future.
```
ignored_files = [
   "nimble/include/nimble/ble.h",
]

(...)

if (filename in ignored_files):
  return False
```

.github/check_doxygen.py
```diff
@@ -0,0 +1,112 @@
...
+
+def main() -> bool:
+    if len(sys.argv) > 1:
+        commit = sys.argv[1]
+    else:
+        commit = "HEAD"
+    if len(sys.argv) > 2:
+        upstream = sys.argv[2]
+    else:
+        upstream = "origin/master"
```
MariuszSkamra, line 62: Nitpick: (but can be useful) consider using argparse for command line parameters, to describe the possible options

.github/check_doxygen.py
wpiet: Thanks, will add the list.

Actually the mentioned file is about to be included (is_ignored returns False), because it is outside the directory to be included which is specified by the regular expression. Nonetheless, it contains code widely used throughout the NimBLE, so it is desirable to have it documented.

sjanc (approved):

PR #1715, opened 2024-03-01
host/ble_hs: call ble_iso_rx_data if BLE_ISO_BROADCAST_SINK

KKopyscinski:
BLE_ISO defines ISO as whole, but we RX iso data only when device is broadcast sink.


KKopyscinski:
@MariuszSkamra I saw app not compiling with BLE_ISO enabled but disabled BLE_ISO_BROADCAST_SINK

MariuszSkamra (approved):

PR #1716, opened 2024-03-01
ci: Fix Check ports syscfg update workflow

wpiet:
Changes the way 'cd' command is used, so it takes into account that each step is run in a separate shell instance.

### DIFF
--- a/.github/workflows/ports_syscfg_check.yml
+++ b/.github/workflows/ports_syscfg_check.yml
@@ -52,21 +52,21 @@ jobs:
              newt upgrade --shallow=1
              rm -rf repos/apache-mynewt-nimble
              git clone .. repos/apache-mynewt-nimble
-             cd ..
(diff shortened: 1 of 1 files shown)


.github/workflows/ports_syscfg_check.yml
```diff
@@ -58,15 +58,18 @@ jobs:
         run: |
              cd build
              ./repos/apache-mynewt-nimble/porting/update_generated_files.sh
-             cd repos/apache-mynewt-nimble
+             cd ..
```
andrzej-kaczmarek, line 61: there's no point in changing working directory at the end of script

wpiet: Agreed, but that was what @sjanc suggested for this script to be more future-proof. Should I remove these 'cd's at the end of each step?

PR #1719, opened 2024-03-04
ext: add libsamplerate library

KKopyscinski:
Adds libsamplerate library that allows to resample audio streams to different samplerate.


rymanluk (approved):
LGTM

PR #1722, opened 2024-03-06
apps/bttester: fix find_included_cb

KKopyscinski:
included_handle was always the same, instead, it should be equal to start handle + 1 + index of currently appended service


KKopyscinski:
@piotrnarajowski please try this patch

PR #1743, opened 2024-04-04
host/ble_audio.c: fix ble_audio_adv_parse_broadcast_announcement

KKopyscinski:
data->success was set to false at every entry to the funcion. That made the valid advertising payload fail after parsing
BLE_BROADCAST_PUB_ANNOUNCEMENT_SVC_UUID - final break in block was reached but the success set in parsing
BLE_BROADCAST_AUDIO_ANNOUNCEMENT_SVC_UUID was cleared.


KKopyscinski:
@MariuszSkamra PTAL

MariuszSkamra (approved):

PR #1749, opened 2024-04-08
nimble/bttester: Make adv parameters global & small fixes

szymon-czapracki:
MYNEWT_VAL is corrected in multiple places for an up-to-date name. 
Missing semicolon and invalid function name are fixed in btp_bap.c
Change codec configuration to be set on BIS level.
Add necessary syscfg options for BAP tests.


szymon-czapracki:
@KKopyscinski @sjanc FYI

KKopyscinski (approved):

apps/bttester/syscfg.yml
```diff
@@ -97,6 +97,8 @@ syscfg.vals:
     BLE_ISO_MAX_BISES: 1
     BLE_ISO_MAX_BIGS: 1
     BLE_EXT_ADV: 1
+    BLE_PHY_2M: 1
+    BLE_EXT_ADV_MAX_SIZE: 261
```
sjanc, line 101: why is this needed? it is quite big value

szymon-czapracki: Updated the value to `40` - previous value was an overkill. This is necessary for broadcast, and specially extended advertising configuration in broadcast bttester application.

sjanc (approved):

PR #1750, opened 2024-04-09
host/audio/ble_iso.c: fix deallocation of BIS

KKopyscinski:
Freeing memory back to BIS pool in loop while changing removing ISO connections from list lead to memory leak. Now, memory is freed in separate loop.


KKopyscinski:
@MariuszSkamra PTAL

MariuszSkamra (approved):

KKopyscinski:
fixed `rem_bis` being uninitialised

PR #1756, opened 2024-04-10
Fix incorrect PACS characteristic notification

szymon-czapracki:
Fix for incorrect  PACS characteristic sending notification after changing it's state.


szymon-czapracki:
@KKopyscinski @sjanc FYI

KKopyscinski (approved):

PR #1757, opened 2024-04-10
host/audio/pacs: fix possible NULL dereference

KKopyscinski:
avail_ctx may be null.


KKopyscinski:
FYI @szymon-czapracki

sjanc (approved):

PR #1759, opened 2024-04-11
host/audio/pacs: fix notifications for available contexts

KKopyscinski:
Available contexts values are per connection. When value is changed, read access for notification gives BLE_HS_CONN_HANDLE_NONE as conn handle. This was making impossible to fetch the right value. Now, flag was introduced, for marking the changed value.


KKopyscinski:
@szymon-czapracki PTAL

szymon-czapracki:
> @szymon-czapracki PTAL

Tested againts PTS - works fine.

sjanc (approved):

PR #1761, opened 2024-04-11
nimble: controller: Add DRBG for Channel Sounding

mkasenberg:


mkasenberg:
Just rebase

nimble/controller/src/ble_ll_cs_drbg.c
```diff
@@ -0,0 +1,681 @@
...
+        shuffled_array[j] = channel_array[i];
+    }
+
+    return 0;
+}
+
+static int
+cs_autocorrelation_score(uint32_t sequence)
+{
+    int i, k, score = 0;
```
sjanc: should those be unsigned and abs() dropped?  (at least clang seems to be complaining)

PR #1762, opened 2024-04-12
nimble/porting: Fix OS_MEMPOOL_SIZE() macro

donatieng:
The OS_MEMPOOL_SIZE() macro returns an invalid size when guards are used (an additional 4 bytes must be reserved for each block, which is not done at the moment). Correct this in line with the macro defined in mynewt core.


donatieng:
Reference: https://github.com/apache/mynewt-core/blob/cc006e6025682c4e0563496fc1014b7d9d07b78c/kernel/os/include/os/os_mempool.h#L179

sjanc (approved):

PR #1763, opened 2024-04-15
host/audio/pacs: Minor improvements

KKopyscinski:


KKopyscinski:
@szymon-czapracki if this gets merged, PACS tests will require supported contexts to be set before setting available ones, and affect the tests. I think you could test this and add setting supported contexts even before this is merged

sjanc (approved):

PR #1766, opened 2024-04-17
host/ble_l2cap_coc: fix possible race condition in L2CAP COC

KKopyscinski:
Two functions: `ble_l2cap_coc_continue_tx` and `ble_l2cap_coc_recv_ready` can be reached from other task than host, for example separate thread running in application, by calling `ble_l2cap_recv_ready` or `ble_l2cap_send`. Because pointer to chan passed into  these functions may be outdated (connection is being terminated by host) this will lead to either:
 - triggering BLE_HS_DBG_ASSERT(conn != NULL) in ble_hs_conn_find_assert
 - dereferencing NULL pointer to conn, after ble_hs_conn_find_assert returns NULL (case with disabled BLE_HS_DEBUG) These two cases shall not cause assert, and should be prepared for chan not existing anymore. Now, BLE_HS_ENOTCONN is returned if connection no longer exists.

### DIFF
--- a/nimble/host/src/ble_l2cap_coc.c
+++ b/nimble/host/src/ble_l2cap_coc.c
@@ -506,7 +506,12 @@ ble_l2cap_coc_continue_tx(struct ble_l2cap_chan *chan)
             goto failed;
         }
 
-        conn = ble_hs_conn_find_assert(chan->conn_handle);
+        conn = ble_hs_conn_find(chan->conn_handle);
+        if (!conn) {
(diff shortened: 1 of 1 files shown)


KKopyscinski:
Rest of functions that use `ble_hs_conn_find_assert` are reachable from host and should indeed cause assert to be triggered, or have conn already checked, as described in `8ec1b8d BLE Host - Check conn state once before proc.`

rymanluk (approved):
LGTM

PR #1767, opened 2024-04-18
ext/libsamplerate: add patch with single precision compilation option

KKopyscinski:
This allows to build library with floats instead of doubles.


KKopyscinski:
@m-gorecki PTAL, this works but I want to be sure it's OK :)

m-gorecki:
LGTM

andrzej-kaczmarek (approved):

sjanc (changes requested):
please hold this until I figure out on how we should handle patch files licencing

rymanluk:
@sjanc any ideas on how to handle that ?

PR #1770, opened 2024-04-29
nimble: iso: Move Rx callback param to Setup ISO Data Path command

MariuszSkamra:
This moves the ISO Rx callback to be provided by application as Setup ISO Data Path function parameter, as the callback will be used only for HCI data path, thus there is no point to require the callback when the broadcast sync is created.


KKopyscinski (approved):

KKopyscinski (approved):
LGTM

PR #1776, opened 2024-05-08
apps/auracast_usb: broadcast improvements

KKopyscinski:
this commit:
- adds fix for crash on to much/corrupted data on USD
- adds support for all LC3 samplerates (besides 44.1kHz - unsupported)
- fixes broadcast with HCI feedback disabled
- by default resampler is disabled - only works for low samplerates, or low quality


rymanluk (changes requested):
Looks like it should be 5 separate commits

KKopyscinski:
> Looks like it should be 5 separate commits

will split

rymanluk (approved):
LGTM

PR #1784, opened 2024-05-17
[DNM][WIP]BASS fixes in source handling.

szymon-czapracki:
Fix infinite loops
Fix event naming in remove source.
Add variable to hold characteristic handle in remove source.
Fix incorrect address type in add source.


szymon-czapracki:
Do not merge - work in progress.

PR #1785, opened 2024-05-20
nimble/transport: Fix undeclared function 'min' warning

guy-m:
Fix build warning generated with llvm based compilers on a call to undeclared function 'min'.


sjanc (approved):
@andrzej-kaczmarek I guess we should think about some common ble_utils.h or similar in long run...

guy-m:
@sjanc Yes. I agree. That way we can avoid this definitions (and others) in multiple c files.

andrzej-kaczmarek (approved):

PR #1792, opened 2024-06-04
nimble/host: NULL assign data pointer after free

rahult-github:
After buffer is freed, explicitly NULL assign the pointer to data.

### DIFF
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -3671,6 +3671,7 @@ ble_gap_ext_adv_set_data(uint8_t instance, struct os_mbuf *data)
 
 done:
     os_mbuf_free_chain(data);
+    data = NULL;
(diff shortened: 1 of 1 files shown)


rahult-github:
@sjanc  Please take a look and review

sjanc (commented):
Hi,

I'm not sure what this patch fix, data is local pointer

rahult-github:
Hi @sjanc , 

So data here is the pointer which is passed as parameter ble_gap_ext_adv_set_data . Apps like bleprph, create a data pointer, allocate it and then invoke ble_gap_ext_adv_set_data .  Since the function does have the code to free the "os_mbuf *data,  this change is to also NULL initialize it.

andrzej-kaczmarek:
you set *local* variable literally just before returning from function. that is just pointless.

PR #1815, opened 2024-07-30
nimble/host: Modify Advertising Tx power levels

rahult-github:
For LE Read Advertising Physical Channel Tx Power command , different version of BLE specification have updated the return values for Tx Power level. 

Currently (spec 5.2 onwards) the supported values are from (-127) to (20). 

Latest values as per Spec ver 5.4 , Vol 4, Part E, section 7.8.6 has range specified. 

Update the min / max macros to reflect the same


rahult-github:
> this depends on core spec version so should be #ifdef
> 
> also if you refer to spec, please include reference to exact core spec version and section

Thanks for this, have updated the PR to reflect different versions and updated the description too

sjanc (approved):

PR #1824, opened 2024-08-01
apps/auracast_usb: Fix calculation for octets_per_frame and max_sdu

andrzej-kaczmarek:


KKopyscinski (approved):
besides exceeded line lengths caught by compliance check LGTM :)

PR #1828, opened 2024-08-05
nimble/ll: Add workaround (hack) for empty BIS pretransmissions

andrzej-kaczmarek:
This adds a very ugly hack to avoid sending empty BIS PDUs for pretransmissions in case we don't have data to send. If receiver receives an empty PDU in pretransmission, it may decide not to try and receive a PDU in proper event later and audio will be corrupted.


rymanluk:
we do not need it anymore, rigth

PR #1831, opened 2024-08-06
nimble/ll: Add option to prefill isoal mux

andrzej-kaczmarek:
This adds option to queue initial SDUs in mux until there are enough data to fill entire ISO event.

This makes it easier to handle pre-transmissions since we don't care if host is aware that it should send more data for the 1st event to fill pre-transmission subevents - we will simply not "release" data from mux until there are enough SDUs to fill complete event (i.e. initial ISO events may contain only empty packets).


sjanc (approved):

rymanluk (approved):
LGTM and works perfect

PR #1833, opened 2024-08-08
nimble/audio: Modify BASS control point write access.

szymon-czapracki:
Modify BASS control point write access in a way that it would use proper mbuf for holding incoming data.

Fix handler check from simple bool comparison to ensuring that it's not NULL.

### DIFF
--- a/nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
+++ b/nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
@@ -617,13 +617,16 @@ ble_svc_audio_bass_find_handler(uint8_t opcode)
 static int
 ble_svc_audio_bass_ctrl_point_write_access(struct ble_gatt_access_ctxt *ctxt, uint16_t conn_handle)
 {
+    uint8_t opcode;
+    struct os_mbuf *om;
+    uint16_t mbuf_len = OS_MBUF_PKTLEN(ctxt->om);
     struct ble_svc_audio_bass_ctrl_point_handler *handler;
 
-    uint8_t opcode = ctxt->om->om_data[0];
+    opcode = ctxt->om->om_data[0];
 
     handler = ble_svc_audio_bass_find_handler(opcode);
 
-    if (!handler) {
+    if (handler == NULL) {
         return BLE_SVC_AUDIO_BASS_ERR_OPCODE_NOT_SUPPORTED;
     }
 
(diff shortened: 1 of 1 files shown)


nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
```diff
@@ -618,12 +618,16 @@ static int
 ble_svc_audio_bass_ctrl_point_write_access(struct ble_gatt_access_ctxt *ctxt, uint16_t conn_handle)
 {
     struct ble_svc_audio_bass_ctrl_point_handler *handler;
+	uint16_t mbuf_len = OS_MBUF_PKTLEN(ctxt->om);
+	uint8_t opcode;
+	uint8_t data[MYNEWT_VAL(BLE_SVC_AUDIO_BASS_METADATA_MAX_SZ)];
 
-    uint8_t opcode = ctxt->om->om_data[0];
+	os_mbuf_copydata(ctxt->om, 0, mbuf_len, data);
```
rymanluk, line 625: are we sure we want to put those data on the stack?

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
```diff
@@ -618,12 +618,16 @@ static int
 ble_svc_audio_bass_ctrl_point_write_access(struct ble_gatt_access_ctxt *ctxt, uint16_t conn_handle)
 {
     struct ble_svc_audio_bass_ctrl_point_handler *handler;
+	uint16_t mbuf_len = OS_MBUF_PKTLEN(ctxt->om);
+	uint8_t opcode;
+	uint8_t data[MYNEWT_VAL(BLE_SVC_AUDIO_BASS_METADATA_MAX_SZ)];
```
MariuszSkamra, line 623: please follow the coding style - code is not aligned

MariuszSkamra (changes requested):
Could you please improve the commit message, I'm not sure I understand what this change supposed to fix.

But I see what it breaks unfortunately. The data you pass to the handler include the opcode (as you pass &data[0], not &data[1]), while the handles assume the 0 parameter won't be opcode

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
szymon-czapracki: Previous implementation produced some problems regarding passing data into specific handlers, as not all of the data was present - after discussing this with @sjanc we've decided that a good solution would be to use os_mbuf for passing the data that write access function handles.

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
szymon-czapracki: Done

MariuszSkamra (approved):

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
```diff
@@ -633,7 +636,12 @@ ble_svc_audio_bass_ctrl_point_write_access(struct ble_gatt_access_ctxt *ctxt, ui
         return BLE_ATT_ERR_WRITE_REQ_REJECTED;
     }
 
-    return handler->handler_cb(&ctxt->om->om_data[1], ctxt->om->om_len - 1, conn_handle);
+    om = os_mbuf_pullup(ctxt->om, mbuf_len);
```
szymon-czapracki, line 639: Disregard om, use ctx->om

sjanc (approved):
in long term we should consider passing mbuf directly to handler so that pullup is not needed

rymanluk (approved):
lgtm

PR #1836, opened 2024-08-08
nimble/audio/bass: BASS modify source update

szymon-czapracki:
Update modify source to work with
new mbuf-style data input.

Also a handling for proper receive state fields is added.

Fix a never-ending goto statement returning to goto itself.

### DIFF
--- a/nimble/host/audio/services/bass/include/services/bass/ble_audio_svc_bass.h
+++ b/nimble/host/audio/services/bass/include/services/bass/ble_audio_svc_bass.h
@@ -326,6 +326,9 @@ struct ble_svc_audio_bass_operation {
 
             /** BIS Synchronisation of subgroups */
             uint32_t bis_sync[BLE_SVC_AUDIO_BASS_SUB_NUM_MAX];
+            /** Subgroup entries */
+            struct ble_svc_audio_bass_subgroup
+                subgroups[BLE_SVC_AUDIO_BASS_SUB_NUM_MAX];
         } modify_source;
 
         /**
--- a/nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
+++ b/nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
@@ -457,6 +457,7 @@ ble_svc_audio_bass_modify_source(uint8_t *data, uint16_t data_len, uint16_t conn
             .status = 0
         }
     };
+    uint8_t *metadata_ptr;
     uint8_t offset = 0;
     int rc = 0;
     int i;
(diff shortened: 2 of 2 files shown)


nimble/host/audio/services/bass/include/services/bass/ble_audio_svc_bass.h
```diff
@@ -325,7 +325,11 @@ struct ble_svc_audio_bass_operation {
             uint16_t num_subgroups;
 
             /** BIS Synchronisation of subgroups */
+            M
```
rymanluk, line 328: looks like trash :)

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
```diff
@@ -477,8 +479,7 @@ ble_svc_audio_bass_modify_source(uint8_t *data, uint16_t data_len, uint16_t conn
 
     operation.modify_source.pa_interval = get_le16(&data[offset]);
     offset += 2;
-    operation.modify_source.num_subgroups = get_le16(&data[offset]);
-    offset += 2;
+    operation.modify_source.num_subgroups = get_le16(&data[offset++]);
```
rymanluk, line 482: is `offset++`  intended here?

szymon-czapracki: It's moved below - this was a mistake. We have to move offset after reading number of subroups (which size is 1 byte - but I'm sure u know that :) ) 
image

nimble/host/audio/services/bass/include/services/bass/ble_audio_svc_bass.h
szymon-czapracki: Removed

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
```diff
@@ -478,7 +479,7 @@ ble_svc_audio_bass_modify_source(uint8_t *data, uint16_t data_len, uint16_t conn
     operation.modify_source.pa_interval = get_le16(&data[offset]);
     offset += 2;
     operation.modify_source.num_subgroups = get_le16(&data[offset]);
-    offset += 2;
```
rymanluk, line 481: this seems to be bad

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
```diff
@@ -489,7 +490,18 @@ ble_svc_audio_bass_modify_source(uint8_t *data, uint16_t data_len, uint16_t conn
 
     for (i = 0; i < operation.modify_source.num_subgroups; i++) {
         operation.modify_source.bis_sync[i] = get_le32(&data[offset]);
+        rcv_state->state.subgroups[i].bis_sync_state = operation.modify_source.bis_sync[i];
```
rymanluk, line 493: this state will be updated when actual SYNC will happen. At this stage we cannot do that as we overwrite a state.

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
```diff
@@ -509,12 +521,30 @@ ble_svc_audio_bass_modify_source(uint8_t *data, uint16_t data_len, uint16_t conn
...
+    }
 
 done:
     if (!rc) {
         rc = ble_svc_audio_bass_receive_state_notify(rcv_state);
         ev.bass_operation_status.status = rc;
-        goto done;
     }
 
     ble_audio_event_listener_call(&ev);
```
rymanluk, line 550: this even here, will request application to sync. Once this is done, state will be set correctly and expected notification will be sent out

nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
szymon-czapracki: Removed

Issue #1090, opened 2021-11-14
GATTS long reads behavior


rojer:
In general, I like that NimBLE tries to hide the complexity of long reads and writes from the user, automatically servicing them within the stack, it's a welcome change from Bluedroid that spills all this onto the user of the API and expects the user to deal with all the complications.

However, currently observed behavior for long reads is that NimBLE will repeatedly request the value from the attribute callback while responding to central's requests and sending MTU chunks at a time (at least i suspect that's what's happening). However, if the value of the characteristic happens to change between reads, incorrect value will be returned to the client. As far as i can tell, there is current;y no way for the API user to tell if  a particular BLE_GATT_ACCESS_OP_READ_CHR is part of the same long read or not - for example, read offset is not provided to the user, so user application cannot work this around either.

IMO, NimBLE should buffer the long response and serve long reads from internally buffered copy of the response and not rely on the value staying the same.

ivanholmes:
I ran in to this as well. My callback contained a function that gathered data from an external source which would run multiple times and crash the microcontroller. I must say I found it quite confusing that this was happening, although I have now had a practical education on one of the inner workings of BLE, which is a good thing I suppose!

I think @rojer's idea of buffering the data is a good one. It would also be useful to be able to tell inside the callback function if it is being run for the first time in a long read, the last time, or somewhere in the middle. 
This would allow the implementation of characteristics that iterate through a data structure, which would be useful in my application.

macksal:
Running into this sporadically. It is a confusing behaviour that the offset reads are handled, but the developer still needs to consider this case.

Buffering internally in Nimble would be good, but I suppose that could be an equally nasty surprise in a low resource environment.

@ivanholmes @rojer did you end up with any workarounds? I'm thinking of caching the value when it's first read, and returning the cached value if it's read again within some time period (e.g. 5 seconds).

rojer:
@macksal yep, in my library i buffer the full response and wait until MinBLE reads it out in MTU-1 byte sized chunks ([here](https://github.com/mongoose-os-libs/bt-common/blob/a8aae713f4fae16169f09c27e07f707a5f52a570/src/esp32xx/esp32xx_bt_gatts.c#L569)). this seems to be working so far but could break if nimble decides to change their behavior.

sjanc:
Hi,

[...]

PR #1846, opened 2024-08-10
Offset field in struct ble_gatt_access_ctxt as an indication of long attribute read

vandy:
To obtain an attribute value when the attribute size is greater than configured MTU long attribute read procedure could be used.
A client sends consecutive `ATT_READ_BLOB_REQ` requests (passing along an offset) to consume the attribute value. The offset is used internally in the nimble stack but is not passed to a user provided access callback in a GATT characteristic description. This callback is called the same number of times as the number of blob requests, but it's not possible to say whether it's a first request (or maybe just `ATT_READ_REQ`) or consecutive request.

The provided offset could indicate this situation to the app developer. However, Core Spec 5.4 Vol 3 Part F 3.4.4.5 notes:
> Note: The value of a Long Attribute may change between the server receiving one ATT_READ_BLOB_REQ PDU and the next ATT_READ_BLOB_REQ PDU. A higher layer specification should be aware of this and define appropriate behavior.

It's up to a developer to decide what to do, when consecutive (long attribute) reading takes place. This was discussed in #1090.

### DIFF
--- a/nimble/host/src/ble_gatts.c
+++ b/nimble/host/src/ble_gatts.c
@@ -416,6 +416,7 @@ ble_gatts_chr_val_access(uint16_t conn_handle, uint16_t attr_handle,
 
     gatt_ctxt.op = ble_gatts_chr_op(att_op);
     gatt_ctxt.chr = chr_def;
+    gatt_ctxt.offset = offset;
 
     ble_gatts_chr_inc_val_stat(gatt_ctxt.op);
     rc = ble_gatts_val_access(conn_handle, attr_handle, offset, &gatt_ctxt, om,
--- a/nimble/host/include/host/ble_gatt.h
+++ b/nimble/host/include/host/ble_gatt.h
@@ -960,6 +960,12 @@ struct ble_gatt_access_ctxt {
          */
         const struct ble_gatt_dsc_def *dsc;
     };
+
+    /**
+     * An offset in case of BLE_ATT_OP_READ_BLOB_REQ.
+     * If the value is greater than zero it's an indication of a long attribute read.
+     */
+    uint16_t offset;
 };
 
 /**


vandy:
I've also made some changes to other parts of the code (only code style without change in functionality). 

And provided some points to consider with `REVIEW` comments. I'd love to get a response on this topics and if they are approved I'd make changes to the code as described in this comments.

porting/nimble/src/os_mbuf.c
```diff
@@ -633,19 +626,20 @@ os_mbuf_adj(struct os_mbuf *mp, int req_len)
          * after the remaining size.
          */
         len = -len;
+        if (m->om_len >= len) {
```
vandy, line 629: If the mbuf size is greater than length than we don't need to iterate over the whole chain to count the overall length.

vandy:
Right now the nimble stack takes all attribute data from an application and truncates it to be the size of MTU (actually MTU - 1). E.g. attribute length is 500 bytes, MTU is 23 bytes. So the application copies 500 bytes to the provided mbuf, and then nimble stack copies only 22 bytes according to an offset.

I think it would be nice to allow the application callback to provide only part of the data (as it already knows the offset and could obtain mtu length information). What do you think?

andrzej-kaczmarek:
> I've also made some changes to other parts of the code (only code style without change in functionality).

code style fixes should be at least in separate commit, but preferably in separate PR

vandy:
@andrzej-kaczmarek, I've split this PR into 3: #1847 and #1848. But left `REVIEW` comments here until the discussion and decision is settled. I believe it would be better to make another PR for discussed changes.

andrzej-kaczmarek:
you can make separate PR for other changes so we can discuss them.
just keep adding the offset field to the struct in this PR and it's good to go. please don't reformat other pieces of code though...

vandy:
Done, left only changes related to the offset.

vandy:
Amended the commit message. Also, moved issues to discuss into #1849.

nimble/host/src/ble_gatts.c
```diff
@@ -407,21 +407,19 @@ ble_gatts_chr_val_access(uint16_t conn_handle, uint16_t attr_handle,
...
     gatt_ctxt.op = ble_gatts_chr_op(att_op);
     gatt_ctxt.chr = chr_def;
+    gatt_ctxt.offset = offset;
 
     ble_gatts_chr_inc_val_stat(gatt_ctxt.op);
-    rc = ble_gatts_val_access(conn_handle, attr_handle, offset, &gatt_ctxt, om,
-                              chr_def->access_cb, chr_def->arg);
 
-    return rc;
+    return ble_gatts_val_access(conn_handle, attr_handle, offset, &gatt_ctxt, om,
```
andrzej-kaczmarek, line 421: do not reformat code as a part of this PR, just add what is necessary to handle offset

andrzej-kaczmarek (changes requested):

nimble/host/src/ble_gatts.c
vandy: I've reverted reformat code changes, only offset is left.

vandy:
@andrzej-kaczmarek, can we merge now?

andrzej-kaczmarek (approved):

PR #1847, opened 2024-08-11
Code style changes

vandy:
This PR is linked to #1846. Code style changes made during code investigation on the offset.


andrzej-kaczmarek (changes requested):
those are not code style changes, you just reformat code that is perfectly fine, e.g. using `rc` as intermediate value instead of just returning value from function call directly is our preferred way in code.

vandy:
"... just reformat code that is perfectly fine" - is what I thought code style changes are. I've reverted some changes. So, please, consider what to do with this PR. Would you accept these changes or not. If not, the PR could be close I suppose.

PR #1848, opened 2024-08-11
os_mbuf trailing adjustment changes.

vandy:
This PR is linked with #1846.

If the `os_mbuf` size is greater than length than we don't need to iterate over the whole chain to count the overall length.

### DIFF
--- a/porting/nimble/src/os_mbuf.c
+++ b/porting/nimble/src/os_mbuf.c
@@ -540,28 +540,21 @@ struct os_mbuf *
 os_mbuf_off(const struct os_mbuf *om, int off, uint16_t *out_off)
 {
     struct os_mbuf *next;
-    struct os_mbuf *cur;
-
     /* Cast away const. */
-    cur = (struct os_mbuf *)om;
-
-    while (1) {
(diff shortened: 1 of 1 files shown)


porting/nimble/src/os_mbuf.c
```diff
@@ -633,19 +626,20 @@ os_mbuf_adj(struct os_mbuf *mp, int req_len)
          * after the remaining size.
          */
         len = -len;
+        if (m->om_len >= len) {
+            m->om_len -= len;
+            if (OS_MBUF_IS_PKTHDR(mp))
+                OS_MBUF_PKTHDR(mp)->omp_len -= len;
+            return;
+        }
```
andrzej-kaczmarek, line 634: this will trim the length of 1st mbuf in chain instead of the last one

andrzej-kaczmarek (changes requested):

porting/nimble/src/os_mbuf.c
vandy: That's right. My bad. Closing the PR.

PR #1851, opened 2024-08-14
nimble/host: Add support for ext adv param v2 HCI command

rahult-github:
BLE spec version 5.4 introduced support for host to pass primary and secondary phy options in set extended advertising parameters. For this a new HCI command "HCI_LE_Set_Extended_Advertising_Parameters v[2] " with OCF 0x007F got introduced. 

This patch adds support in host for this HCI command.

### DIFF
--- a/nimble/host/include/host/ble_gap.h
+++ b/nimble/host/include/host/ble_gap.h
@@ -1582,6 +1582,12 @@ struct ble_gap_ext_adv_params {
 
     /** Advertising Set ID */
     uint8_t sid;
+
+    /** Primary phy options */
+    uint8_t pri_phy_opt;
+
+    /** Secondary phy options */
+    uint8_t sec_phy_opt;
 };
 
 /**
--- a/nimble/host/src/ble_gap.c
+++ b/nimble/host/src/ble_gap.c
@@ -3124,45 +3124,38 @@ ble_gap_adv_active(void)
 
 #if MYNEWT_VAL(BLE_EXT_ADV)
 static int
-ble_gap_ext_adv_params_tx(uint8_t instance,
-                          const struct ble_gap_ext_adv_params *params,
-                          int8_t *selected_tx_power)
-
+ble_gap_set_ext_adv_params(struct ble_hci_le_set_ext_adv_params_cp *cmd,
+                           uint8_t instance, const struct ble_gap_ext_adv_params *params,
+                           int8_t *selected_tx_power)
 {
-    struct ble_hci_le_set_ext_adv_params_cp cmd;
-    struct ble_hci_le_set_ext_adv_params_rp rsp;
-    int rc;
-
-    memset(&cmd, 0, sizeof(cmd));
-
-    cmd.adv_handle = instance;
+    cmd->adv_handle = instance;
 
     if (params->connectable) {
-        cmd.props |= BLE_HCI_LE_SET_EXT_ADV_PROP_CONNECTABLE;
+        cmd->props |= BLE_HCI_LE_SET_EXT_ADV_PROP_CONNECTABLE;
     }
     if (params->scannable) {
-        cmd.props |= BLE_HCI_LE_SET_EXT_ADV_PROP_SCANNABLE;
+        cmd->props |= BLE_HCI_LE_SET_EXT_ADV_PROP_SCANNABLE;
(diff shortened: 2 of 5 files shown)


rahult-github:
@sjanc  please take a look and review

rahult-github:
Hi @sjanc  , please review.

rahult-github:
Hi @sjanc , @andrzej-kaczmarek  , please let me know in case of any further comments

nimble/host/include/host/ble_gap.h
```diff
@@ -1582,6 +1582,14 @@ struct ble_gap_ext_adv_params {
 
     /** Advertising Set ID */
     uint8_t sid;
+
+#if MYNEWT_VAL(BLE_EXT_ADV_V2)
```
rymanluk, line 1586: since below fields are OK to be `0` which means, `host has not preference` I would skip `BLE_EXT_ADV_V2` .

Instead, host could choose to use V2 or V1 based on the Read Supported Commands result.

@sjanc @andrzej-kaczmarek  what do you think?

rahult-github: Hi @rymanluk  
 
> Instead, host could choose to use V2 or V1 based on the Read Supported Commands result. 

Just to understand this comment, are you implying that use V2 directly if its support is present ? ( i.e. don't use V1  in such cases ? )  Will host not able to use V1, if controller has support for V2 ?

sjanc:
until now we didn't have need for V2 (or more) HCI commands in host but I think we should start tracking controller supported commands.

I believe we should be able to simply extend host API and if no preference is selected we can fallback to V1 if V2 is not supported,  in case of preference set by app, we fail if V2 is not supported

nimble/host/include/host/ble_gap.h
rahult-github: Hi @rymanluk , i have updated the PR to not use the BLE_EXT_ADV_V2 macro and rather decide based on Read Supported Commands.

rahult-github:
> until now we didn't have need for V2 (or more) HCI commands in host but I think we should start tracking controller supported commands.
> 
> I believe we should be able to simply extend host API and if no preference is selected we can fallback to V1 if V2 is not supported, in case of preference set by app, we fail if V2 is not supported

So current implementation to read supported commands is only for event mask2 . I have extended it to use all commands bits ( so that future implementation can use it as needed).

nimble/host/include/host/ble_gap.h
```diff
@@ -1582,6 +1582,19 @@ struct ble_gap_ext_adv_params {
 
     /** Advertising Set ID */
     uint8_t sid;
+
+    /** Version of HCI command to be used
+     *  0 : No version set, choose V2 or fallback to v1
+     *  1 : V1
+     *  2 : V2
+     */
+    unsigned int version:2;
```
andrzej-kaczmarek, line 1591: this is not needed - you can return an error if phy options are non-zero and v2 is not supported, otherwise send either v1 or v2.

nimble/host/src/ble_gap.c
```diff
@@ -3124,17 +3124,113 @@ ble_gap_adv_active(void)
 
 #if MYNEWT_VAL(BLE_EXT_ADV)
 static int
-ble_gap_ext_adv_params_tx(uint8_t instance,
```
andrzej-kaczmarek, line 3127: v1 and v2 structs have the same field order so we can have single helper to fill common fields

andrzej-kaczmarek (changes requested):

nimble/host/src/ble_gap.c
rahult-github: So , since the structures become different, i made some changes in defination to club the common parts. Hope this is ok.

nimble/host/include/host/ble_gap.h
rahult-github: Addressed. Please check once

sjanc:
looks good to me!  but please fix those few style issues reported by CI

rahult-github:
> looks good to me! but please fix those few style issues reported by CI

Thank you. Have pushed again.

sjanc (approved):
CI failures are due to regression in newt tool

PR #1854, opened 2024-08-30
nimble/drivers: Add driver for SKY66405 FEM

andrzej-kaczmarek:
This adds driver for SKY66405 FEM.

The bypass function is only supported using dedicated pin (CPS). Enabling bypass by setting both CTX and CRX to high state is not supported due to limitation in FEM API.

### DIFF
--- a/nimble/drivers/fem/sky66405/src/sky66405.c
+++ b/nimble/drivers/fem/sky66405/src/sky66405.c
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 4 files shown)


nimble/drivers/fem/sky66405/src/sky66405.c
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <assert.h>
```
sjanc, line 20: is this needed?

sjanc (commented):
one nitpick, otherwise LGTM

sjanc (approved):

PR #1855, opened 2024-09-02
porting/npl/linux: Handle EINTR from sem_timedwait

rojer:
It's been observed that `sem_timedwait()` can return `EINTR` even if `sigaction()` specified `SA_RESTART`.

Not sure what the deal with that is but handling it is simple enough.

### DIFF
--- a/porting/npl/linux/src/os_sem.c
+++ b/porting/npl/linux/src/os_sem.c
@@ -71,9 +71,14 @@ ble_npl_sem_pend(struct ble_npl_sem *sem, uint32_t timeout)
         wait.tv_sec  += timeout / 1000;
         wait.tv_nsec += (timeout % 1000) * 1000000;
 
-        err = sem_timedwait(&sem->lock, &wait);
-        if (err && errno == ETIMEDOUT) {
-            return BLE_NPL_TIMEOUT;
+        while ((err = sem_timedwait(&sem->lock, &wait)) != 0) {
+            switch (errno) {
+            case EINTR:
+                continue;
+            case ETIMEDOUT:
+                return BLE_NPL_TIMEOUT;
+            }
+            break;
         }
     }
 


porting/npl/linux/src/os_sem.c
```diff
@@ -71,9 +71,12 @@ ble_npl_sem_pend(struct ble_npl_sem *sem, uint32_t timeout)
         wait.tv_sec  += timeout / 1000;
         wait.tv_nsec += (timeout % 1000) * 1000000;
 
-        err = sem_timedwait(&sem->lock, &wait);
-        if (err && errno == ETIMEDOUT) {
-            return BLE_NPL_TIMEOUT;
+        while ((err = sem_timedwait(&sem->lock, &wait)) != 0) {
+            switch (errno) {
+                case EINTR: continue;
```
sjanc, line 76: this should be

```
case EINTR:
    continue;
case ETIMEDOUT:
    return BLE_NPL_TIMEOUT;
```

I also wonder if there should be default case with assert maybe?

rojer: fixed style.
in the default case we fall through to return `BLE_NPL_ERROR` and stack restart is attempted.
it happens to fail and assert but that's another story[1]. In theory a stack restart is the right thing to do in this case and might fix things.

[1] Assert happens [here](https://github.com/apache/mynewt-nimble/blob/f8d60614b7df21db77f0ff75e30ee104c71fa649/nimble/host/src/ble_hs_hci.c#L316) because ble_hs_sched_reset [here](https://github.com/apache/mynewt-nimble/blob/f8d60614b7df21db77f0ff75e30ee104c71fa649/nimble/host/src/ble_hs_hci.c#L326) bites on its own tail - tries to tx a reset command.

sjanc (approved):

rojer:
fixed whitespace, sorry

sjanc:
no worries,  and thanks for contributing :)

PR #1861, opened 2024-09-14
nimble/host/src/ble_gap.c:  undefined SET_BIT before defining it.

dmainz:
Was getting duplicate definitions of SET_BIT when compiling for an STM32F4 device.


sjanc (approved):
I'm fine with this for time being, in long term we should have common helpers for this

dmainz:
Can you point me to an example of a helper?

PR #1867, opened 2024-09-18
nimble/host: Preempt GAP operations before removing entry from RL

andrzej-kaczmarek:


sjanc (approved):

nimble/host/src/ble_hs_pvcy.c
```diff
@@ -60,7 +60,7 @@ ble_hs_pvcy_set_resolve_enabled(int enable)
 }
 
 int
```
sjanc, line 62: could be static though

rymanluk (approved):

PR #1871, opened 2024-09-23
nimble/audio: Use proper array size in assert within scan delegator.

szymon-czapracki:
This commit changes subgroups filed in assert
used in operation handler in scan delegator.
Now proper defined array size is used.


szymon-czapracki:
@MariuszSkamra @rymanluk FYI

sjanc (approved):

sjanc (approved):

PR #1876, opened 2024-09-23
nimble/ll: Fix periodic advertising event timing

andrzej-kaczmarek:
This modifies padv event start time calculations to use single event
start time as anchor and calculate subsequent event start times as an
offset from that event.

This virtually removes drift due to rounding errors and avoids relative
drift of padv vs. BIG events.


rymanluk (approved):
LGTM, also tested

PR #1884, opened 2024-10-01
nimble/ll: Add set ext adv params v2 command

m-gorecki:
This adds second version of Set Extended
Advertising Parameters command (7.8.53)
Introduced in Core Specification v5.4.

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -3613,6 +3613,57 @@ done:
     return rc;
 }
 
+#if MYNEWT_VAL(BLE_VERSION) >= 54
+static uint8_t
+ble_ll_adv_ext_phy_mode_get(uint8_t phy, uint8_t phy_opt)
+{
+    if (phy != BLE_HCI_LE_PHY_CODED) {
+        return phy;
+    }
+
+    switch (phy_opt) {
+    case BLE_HCI_ADVERTISING_PHY_OPT_S2_PREF:
+    case BLE_HCI_ADVERTISING_PHY_OPT_S2_REQ:
+        return BLE_PHY_MODE_CODED_500KBPS;
+    case BLE_HCI_ADVERTISING_PHY_OPT_NO_PREF:
+    case BLE_HCI_ADVERTISING_PHY_OPT_S8_PREF:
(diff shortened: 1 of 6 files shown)


sjanc (approved):

nimble/include/nimble/hci_common.h
```diff
@@ -1637,6 +1637,13 @@ struct ble_hci_vs_set_local_irk_cp {
 #define BLE_HCI_PRIVACY_NETWORK                     (0)
 #define BLE_HCI_PRIVACY_DEVICE                      (1)
 
+/* --- LE set advertising coded PHY options (OCF 0x007F) */
+#define BLE_HCI_ADVERTISING_PH_OPT_NO_PREF      0x0
```
andrzej-kaczmarek, line 1641: PHY

nimble/controller/src/ble_ll_hci_supp_cmd.c
```diff
@@ -286,6 +286,9 @@ static const uint8_t octet_46 = OCTET(
     BIT(0) /* HCI LE Set Default Subrate */
     BIT(1) /* HCI LE Subrate Request */
 #endif
+#if MYNEWT_VAL(BLE_VERSION) >= 54
```
andrzej-kaczmarek, line 289: missing #ifdef for ext adv support

nimble/controller/src/ble_ll_adv.c
```diff
@@ -3613,6 +3613,56 @@ ble_ll_adv_ext_set_param(const uint8_t *cmdbuf, uint8_t len,
     return rc;
 }
 
+#if MYNEWT_VAL(BLE_VERSION) >= 54
+static uint8_t
+ble_ll_adv_ext_set_coded_opt(uint8_t phy_opt)
```
andrzej-kaczmarek, line 3618: this is not a setter function so 'set' in name is confusing

e.g. `ble_ll_adv_ext_coded_phy_mode_get`

also you can make func that takes both phy and phy_opt as parameter and call it `ble_ll_adv_ext_phy_mode_get`, this way you don't need extra checks for coded by later in the code - just always call this one

andrzej-kaczmarek (approved):

PR #1891, opened 2024-10-17
nimble/ll: Fix offset calculation in BIGInfo

andrzej-kaczmarek:
Offset calculation in BIGInfo assumes BIG event start time is always after AUX_SYNC_IND since BIG should be advanced to next event by the time AUX_SYNC_IND is sent. However, if AUX_SYNC_IND is scheduled close to BIG event this may not happend and thus calculated offset is invalid.

This makes sure that we always use BIG event after AUX_SYNC_IND to calculate offset.


rymanluk (approved):
LGTM

PR #1892, opened 2024-10-17
nimble/ll: Simplify SyncInfo calculation

andrzej-kaczmarek:
This simplifies calculation of SyncInfo used in AUX_ADV_IND and LL_PERIODIC_SYNC_IND to use the same code in both cases.

In case of LL Control PDU we do not "rewind" connsm to find proper connection event but instead "advance" padv event the same way as it was done for AUX_ADV_IND PDU.

### DIFF
--- a/nimble/controller/src/ble_ll_adv.c
+++ b/nimble/controller/src/ble_ll_adv.c
@@ -625,83 +625,73 @@ ble_ll_adv_pdu_make(uint8_t *dptr, void *pducb_arg, uint8_t *hdr_byte)
 #if MYNEWT_VAL(BLE_LL_CFG_FEAT_LL_PERIODIC_ADV)
 static void
 ble_ll_adv_put_syncinfo(struct ble_ll_adv_sm *advsm,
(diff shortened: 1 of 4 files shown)


nimble/controller/src/ble_ll_adv.c
```diff
@@ -625,83 +625,73 @@ ble_ll_adv_pdu_make(uint8_t *dptr, void *pducb_arg, uint8_t *hdr_byte)
...
-        offset -= anchor_rem_us;
-        offset += advsm->padv_event_start_rem_us;
+        ble_ll_conn_anchor_get(connsm, &anchor_event_cntr,
+                               &anchor_ticks, &anchor_rem_us);
 
         /* connEventCount */
-        put_le16(conn_event_cnt, conn_cnt);
+        put_le16(conn_event_cntr, anchor_event_cntr);
+#else
+        BLE_LL_ASSERT(0);
```
rymanluk, line 650: nit: Could be a comment here, that `connsm` shall be used only for SYNC_TRANSFER

rymanluk (approved):
LGMT

PR #1896, opened 2024-10-28
libsamplerate: Add single precision compilation option

MariuszSkamra:
This adds cocci rule that changes all double type occurences to fp_t type that is defined in config.h file.
Depending on LIBSAMPLERATE_USE_SINGLE_PRECISION_FP option, one can enable single precision floating point arithmetic.


MariuszSkamra:
Related newt PR: https://github.com/apache/mynewt-newt/pull/577

PR #1897, opened 2024-10-29
nimble/ll: Set handle in cmd_resp before return

m-gorecki:
Minor fix. We should set conn_handle in the
command response buffer before returning.


andrzej-kaczmarek (approved):

sjanc (changes requested):
shouldn't this also be done in line 2075?

sjanc (approved):

PR #1906, opened 2024-10-31
Fix condition in address recognition

szymon-czapracki:
This commit fixes logic for address
recognition in add_source operation

### DIFF
--- a/nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
+++ b/nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
@@ -329,7 +329,8 @@ ble_svc_audio_bass_add_source(uint8_t *data, uint16_t data_len, uint16_t conn_ha
     operation.conn_handle = conn_handle;
 
     operation.add_source.adv_addr.type = data[offset++];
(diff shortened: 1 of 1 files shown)


nimble/host/audio/services/bass/src/ble_audio_svc_bass.c
```diff
@@ -307,8 +307,9 @@ ble_svc_audio_bass_add_source(uint8_t *data, uint16_t data_len, uint16_t conn_ha
     operation.conn_handle = conn_handle;
 
     operation.add_source.adv_addr.type = data[offset++];
-    if (operation.add_source.adv_addr.type != BLE_ADDR_PUBLIC ||
-        operation.add_source.adv_addr.type != BLE_ADDR_RANDOM) {
+
+    if (!(operation.add_source.adv_addr.type != BLE_ADDR_PUBLIC ||
```
sjanc, line 311: so what is the intention here?  which address types are allowed?  0x00 and 0x01?
Asking cause this check looks rather unorthodox...

sjanc (approved):

PR #1918, opened 2024-11-18
nimble/ll: Add vs hci cmd with additional scan cfg

m-gorecki:
This allows to configure scanner to ignore either
legacy or extended packets. If scan is enabled
command returns command disallowed error.

### DIFF
--- a/nimble/controller/src/ble_ll_hci_vs.c
+++ b/nimble/controller/src/ble_ll_hci_vs.c
@@ -356,6 +356,37 @@ ble_ll_hci_vs_set_local_irk(uint16_t ocf, const uint8_t *cmdbuf, uint8_t cmdlen,
 }
 #endif
 
+#if MYNEWT_VAL(BLE_LL_HCI_VS_SET_SCAN_CFG)
+static int
+ble_ll_hci_vs_set_scan_cfg(uint16_t ocf, const uint8_t *cmdbuf, uint8_t cmdlen,
+                           uint8_t *rspbuf, uint8_t *rsplen)
+{
+    const struct ble_hci_vs_set_scan_cfg_cp *cmd = (const void *)cmdbuf;
+    uint32_t flags;
+    int rc;
+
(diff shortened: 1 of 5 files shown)


nimble/controller/src/ble_ll_hci_vs.c
```diff
@@ -356,6 +356,33 @@ ble_ll_hci_vs_set_local_irk(uint16_t ocf, const uint8_t *cmdbuf, uint8_t cmdlen,
...
+                           uint8_t *rspbuf, uint8_t *rsplen)
+{
+    const struct ble_hci_vs_set_scan_cfg_cp *cmd = (const void *)cmdbuf;
+    int rc;
+
+    if (cmdlen != sizeof(*cmd)) {
+        return BLE_ERR_INV_HCI_CMD_PARMS;
+    }
+
+    if (cmd->flags > BLE_HCI_OCF_VS_SET_SCAN_CFG_NO_EXT) {
```
sjanc, line 371: this should be converted from le32 to host order before check (even if this will be noop on most systems)

it should (at least for now) disallow command if both types are to be disabled

sjanc (approved):

nimble/include/nimble/hci_common.h
```diff
@@ -1396,6 +1396,14 @@ struct ble_hci_vs_set_local_irk_cp {
     uint8_t irk[16];
 } __attribute__((packed));
 
+#define BLE_HCI_OCF_VS_SET_SCAN_CFG_NO_LEGACY           0x00000001
+#define BLE_HCI_OCF_VS_SET_SCAN_CFG_NO_EXT              0x00000002
```
andrzej-kaczmarek, line 1400: ble_hci_ocf prefix suggests this is an OCF which is misleading

andrzej-kaczmarek (changes requested):

sjanc (approved):

andrzej-kaczmarek (approved):

PR #1919, opened 2024-11-18
nimble: btshell: Fix multiple issues

MariuszSkamra:
This fixes multiple issues found during tests in btshell and BASS


apps/btshell/src/cmd.c
```diff
@@ -4235,7 +4235,7 @@ static const struct shell_cmd_help sync_stats_help = {
 #endif
 #endif
 
-#if MYNEWT_VAL(BLE_ISO_BROADCAST_SOURCE)
+#if MYNEWT_VAL(BLE_AUDIO) && MYNEWT_VAL(BLE_ISO_BROADCAST_SOURCE)
```
MariuszSkamra, line 4238: We should have a dedicated syscfg option like `BLE_AUDIO_BROADCAST_SOURCE` actually IMO

sjanc (approved):

PR #1929, opened 2024-11-26
apps: Merge auracast and auracast_usb applications

MariuszSkamra:
There is no point to maintain those two applications as they functionality is the same. This adds new syscfg  `AUDIO_USB` option to toggle support for USB Audio. If the option is enabled, the audio samples from USB are processed as in `apps/auracast_usb`, or test data as in `apps/auracast` otherwise. 
The resulting application is placed in `apps/auracast`.


MariuszSkamra:
CC: @sjanc @KKopyscinski

sjanc (approved):

PR #1930, opened 2024-11-26
apps/blehci: Enable HCI vs event on assert for all asserts

andrzej-kaczmarek:
If HCI vs event on assert is enabled we can log all asserts, not only those wrapped in BLE_LL_ASSERT.


andrzej-kaczmarek:
this requires https://github.com/apache/mynewt-core/pull/3349

PR #1932, opened 2024-11-27
nimble/ll: make includes config dependent

bvdberg:
The call to ble_ll_isoal_reset() is behind define, but include was not. Same for iso_big


sjanc:
hmm not sure why only labeler CI kick in, but others not... could you force push to restart CI?

bvdberg:
pushed (added dot in commit msg)

sjanc:
hmm not sure why only labeler fires up...

andrzej-kaczmarek (approved):

PR #1949, opened 2024-12-28
nimble/phy: Add initial support for nRF54L15 PHY

mkasenberg:


andrzej-kaczmarek:
seems like lost of the code was copied from the nrf5x driver so this should be part of that driver, i.e. the same way as we have nrf52 and nrf53

nimble/drivers/nrf5x/src/nrf53/phy_hw.h
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "../nrf52/phy_hw.h"
```
sjanc: I don't like this approach, although I don't have better idea for the moment (other than common phy_hw with #ifdefs...)   :)

sjanc (commented):
package should be name `nrf54l` since L15,L10 and L5 share same radio

mkasenberg:
I force-pushed the version without the static inline functions because I had @m-gorecki patch applied to it. Should I keep this version or switch to static inline? I remember there was a discussion about the readability of this version, but I don't remember what was the conclusion.

I use this version for Channel Sounding testing.

mkasenberg:
UPF tested

PR #1960, opened 2025-01-21
host: sm: Add support in sm min and max key size configuration

guy-m:
Add option to define the minimum and maximum key size in syscfg.
This is required for test case GAP/SEC/SEM/BI-09.


sjanc:
Hi,

Those are for spec values and not to be tuned by users.
If you want to specify security mode you should use BLE_SM_SC_ONLY and BLE_SM_LVL sysconfigs

PR #1971, opened 2025-02-06
nimble: bttester: add variable to properly initialize characteristics

piotrnarajowski:
New variable that stores value handle of characterisic is introduced. It's needed to properly register predefined characteristics. Previously, two characteristics were using same variable to store value handle, so registering both of them resulted in overwriting handle of the former one


KKopyscinski (approved):
LGTM

PR #1975, opened 2025-02-12
apps/bttester: Fixes

andrzej-kaczmarek:


apps/bttester/src/btp_l2cap.c
```diff
@@ -748,6 +749,8 @@ tester_init_l2cap(void)
     tester_register_command_handlers(BTP_SERVICE_ID_L2CAP, handlers,
                                      ARRAY_SIZE(handlers));
 
+#endif
```
KKopyscinski, line 752: I wonder if proper way to do this would be register commands, but do not compile any command besides `BTP_L2CAP_READ_SUPPORTED_COMMANDS`, and in `supported_commands()` set only

```c
tester_set_bit(rp->data, BTP_L2CAP_READ_SUPPORTED_COMMANDS);
```
and rest put under `#if MYNEWT_VAL(BLE_L2CAP_COC_MAX_NUM)`

sjanc (commented):
L2CAP service is mainly for COC tests, while fixed channels are tested via GAP

I updated autopts project accordingly [1], so I'd just skip l2cap part of that PR

[1] https://github.com/auto-pts/auto-pts/pull/1392

sjanc (approved):

PR #1984, opened 2025-02-21
nimble/ll: Fix supported HCI commands

andrzej-kaczmarek:
We don't need to support this command if there are no host-controllable
features present in LL.

### DIFF
--- a/nimble/controller/src/ble_ll_conn_hci.c
+++ b/nimble/controller/src/ble_ll_conn_hci.c
@@ -32,6 +32,7 @@
 #include "controller/ble_ll_ctrl.h"
 #include "controller/ble_ll_scan.h"
 #include "controller/ble_ll_adv.h"
+#include "ble_ll_priv.h"
 #include "ble_ll_conn_priv.h"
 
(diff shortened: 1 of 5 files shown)


nimble/controller/src/ble_ll_conn_hci.c
```diff
@@ -1932,6 +1933,39 @@ ble_ll_conn_hci_wr_auth_pyld_tmo(const uint8_t *cmdbuf, uint8_t len,
...
+    }
+
+    handle = le16toh(cmd->conn_handle);
+    connsm = ble_ll_conn_find_by_handle(handle);
+    if (!connsm) {
+        return BLE_ERR_UNK_CONN_ID;
+    }
+
+    rsp->conn_handle = cmd->conn_handle;
+    rsp->tx_level = g_ble_ll_tx_power;
```
sjanc, line 1961: According to spec valid range is -30 to 20 dBm...

Isn't this BR/EDR command anyway?

sjanc (approved):
it seems that it is mandatory for LE too...

PR #1992, opened 2025-02-26
ci: Fix doxygen checks

sjanc:
It looks like newer version of Doxygen has issues if no output is generated. Specify to generate HTML output in tmp folder.


sjanc:
@wpiet FYI

kasjer (approved):

PR #1996, opened 2025-02-27
[WIP] nimble/hci: Document vendor specific commands

szymon-czapracki:
Commands were not documented.
Lack of the descriptions could led
to some ambiguous interpretations.


szymon-czapracki:
@sjanc FYI https://github.com/apache/mynewt-nimble/issues/1643

sjanc (commented):
as discussed offline,    this should be documented in text document similarly to how HCI commands and events are specified in core spec

PR #2003, opened 2025-03-06
porting: add missing call to ble_ll_init

gmarull:
`nimble_port_init()` needs to call ble_ll_init() when controller is enabled.


gmarull:
note: https://github.com/apache/mynewt-nimble/pull/1956 resolves the same problem, but needs some tweaks.

PR #2005, opened 2025-03-07
nimble/ll: Fix ISO BIG events scheduling

andrzej-kaczmarek:
ISO events should have 'fixed' flag set for scheduling, except for the 1st one, so they can preempt any other event.


andrzej-kaczmarek:
@MariuszSkamra fyi

sjanc (approved):

PR #2012, opened 2025-03-20
host/store: remove unneeded base64/base64.h include

gmarull:
The include is not needed in ble_store_config.c. It also makes it impossible to compile config store without persistent storage out of mynewt.


gmarull:
@sjanc

sjanc (approved):

PR #2015, opened 2025-03-27
nimble/ll: Introduce ble_ll_addr

andrzej-kaczmarek:


sjanc (commented):
I'm fine in general with this change.

Just wonder if we shouldn't do provisioning API the other way around ie. API that would be  called from provider towards LL.

eg addr_ublox would have init that would call ble_ll_addr_provide_public(const *addr) and set expected value for LL,  and not LL calling externally implemented function

sjanc:
ok, after rethinking this using newt APIs is also fine

PR #2019, opened 2025-04-04
[WIP] EATT improvements

szymon-czapracki:


nimble/host/syscfg.yml
```diff
@@ -331,6 +331,19 @@ syscfg.defs:
...
+    BLE_EATT_AUTO_CONNECT:
+        description: >
+            Enable auto connect for EATT
+        value: 1
+        restrictions:
+            - BLE_L2CAP_ENHANCED_COC
+    BLE_EATT_CHAN_PER_CONN:
+        description: >
+            Maximum number of supported EATT channels per connection
+        value: 5
```
sjanc, line 343: same here,

I'm not sure if there is support arithmetic in here, if not I'd suggest to keep it simple for now and default to 0 when eatt is disabled and 1 when enabled,    default can be tuned later, while target can set what value is needed

nimble/host/src/ble_eatt.c
```diff
@@ -75,7 +76,14 @@ static struct ble_gap_event_listener ble_eatt_listener;
 static struct ble_npl_event g_read_sup_cl_feat_ev;
 
 static void ble_eatt_setup_cb(struct ble_npl_event *ev);
-static void ble_eatt_start(uint16_t conn_handle);
+
+struct os_callout eatt_conn_timer;
```
sjanc, line 80: this needs to be per con, and should be compiled in only when peripheral is supported

nimble/host/src/ble_eatt.c
```diff
@@ -75,7 +76,14 @@ static struct ble_gap_event_listener ble_eatt_listener;
...
+
+struct conn_cb_arg {
+    uint16_t conn_handle;
+};
+
+static struct conn_cb_arg conn_cb_arg;
```
sjanc, line 86: same here,  per conn

nimble/host/src/ble_eatt.c
```diff
@@ -219,39 +244,77 @@ ble_eatt_l2cap_event_fn(struct ble_l2cap_event *event, void *arg)
...
     struct ble_eatt *eatt = arg;
     struct ble_gap_conn_desc desc;
+    uint8_t free_channels;
     uint8_t opcode;
     int rc;
 
     switch (event->type) {
     case BLE_L2CAP_EVENT_COC_CONNECTED:
-        BLE_EATT_LOG_DEBUG("eatt: Connected \n");
+        eatt = ble_eatt_find_by_conn_handle(event->connect.conn_handle);
```
sjanc, line 253: so eatt seems to be also passed as arg to this function   (line 245) ?

nimble/host/src/ble_eatt.c
```diff
@@ -355,7 +441,22 @@ ble_gatt_eatt_write_cl_cb(uint16_t conn_handle,
...
+    struct ble_gap_conn_desc desc;
+    int rc;
+
+    rc = ble_gap_conn_find(conn_handle, &desc);
+    assert(rc == 0);
+
+    if (desc.role == BLE_GAP_ROLE_SLAVE) {
+        /* Add initial delay as peripheral to avoid collision */
+        conn_cb_arg.conn_handle = conn_handle;
+        os_callout_reset(&eatt_conn_timer, 500 * OS_TICKS_PER_SEC / 1000);
```
sjanc, line 454: I'd suggest to use same value as spec recommends for  GATT collisions mitigation resolving

"Peripheral shall  wait a minimum of 100 ms before retrying; on LE connections, the Peripheral shall wait
at least 2 × (connPeripheralLatency + 1) × connInterval if that is longer.  "

nimble/host/src/ble_eatt.c
szymon-czapracki: Yes and no - eatt is passed when this funciton is called in setup_cb. But when server is running, and events are happening, eatt structure might be incomplete, as some of the fields are assigned as a result of event-handling within this function. Thus I thought to look if EATT exsists for specific connection - do you see a better solution here?

nimble/host/src/ble_eatt.c
szymon-czapracki: Done

nimble/host/syscfg.yml
```diff
@@ -331,6 +331,19 @@ syscfg.defs:
         description: >
             MTU used for EATT channels.
         value: 128
+    BLE_EATT_AUTO_CONNECT:
+        description: >
+            Enable auto connect for EATT
+        value: 1
+        restrictions:
```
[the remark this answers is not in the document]
szymon-czapracki: Done

nimble/host/syscfg.yml
szymon-czapracki: Done

sjanc (changes requested):
as discussed offline,  ble_eatt will have to track connected channels (assuming we want single ble_eatt per conn)

PR #2023, opened 2025-04-10
controller: add missing os/util.h include

bvdberg:
ble_ll_hci_vs.c module uses ARRAY_SIZE, defined in os/util.h.


sjanc (approved):
I wish they would have finally resolve this https://github.com/orgs/community/discussions/15452 so that we could more easily track various ports build status without spraying CI results red (we aren't US stock market! :P)

PR #2032, opened 2025-05-07
nimble/host: Add support for Anonymous address type

rahult-github:
Devices can send anonymous address type during advertising.  Added support for this address type which is needed while Adding device to filter accept list

Core spec version 6.0, Vol 4 , Part E, section 7.8.16 lists Anonymous Address type as valid value for LE Add Device to Filter Accept List Command.  Current code doesn't consider this address type. 

image


sjanc (approved):
looks good to me!

sjanc:
CodeQL is currently broken (I'm looking at this) so no need to worry about it

Issue #2007, opened 2025-03-10
GAP/BROB/BCST/BV-05-C test is failing


sjanc:
GAP/BROB/BCST/BV-05-C is failing in AutoPTS

PR #2035, opened 2025-05-09
apps/bttester: Filter policy setup & advertising data fix

szymon-czapracki:
Introduce setting filter policy when accept list is used.
Fix for improper loop initialization during advertising setup.

Those fixes are related to GAP/BROB/BCST/BV-05-C test failing.

### DIFF
--- a/apps/bttester/src/btp_gap.c
+++ b/apps/bttester/src/btp_gap.c
@@ -64,6 +64,8 @@ static ble_addr_t peer_id_addr;
 static ble_addr_t peer_ota_addr;
 static bool encrypted = false;
 
+static bool use_filter_policy = false;
+
 static struct os_callout                    update_params_co;
 static struct btp_gap_conn_param_update_cmd update_params;
 
@@ -515,6 +517,10 @@ start_advertising(const void *cmd, uint16_t cmd_len,
 
 #if MYNEWT_VAL(BLE_EXT_ADV)
     adv_params.own_addr_type = own_addr_type;
+    if (use_filter_policy) {
+        adv_params.filter_policy = BLE_HCI_ADV_FILT_BOTH;
+    }
+
     err = ble_gap_ext_adv_configure(0, &adv_params, NULL, gap_event_cb, NULL);
     if (err) {
         SYS_LOG_ERR("Failed to configure extended advertiser; rc=%d", err);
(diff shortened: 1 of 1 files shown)


szymon-czapracki:
https://github.com/apache/mynewt-nimble/issues/2007

apps/bttester/src/btp_gap.c
```diff
@@ -496,7 +496,7 @@ start_advertising(const void *cmd, uint16_t cmd_len,
         i += ad[adv_len].data_len;
     }
 
-    for (sd_len = 0U; i < cp->scan_rsp_len; sd_len++) {
+    for (i = 0, sd_len = 0U; i < cp->scan_rsp_len; sd_len++) {
```
sjanc, line 502: this will cause scan response to set from adv data...  I think (not tested) that correct fix would be to change loop end condition here to check for both ad and sd len

apps/bttester/src/btp_gap.c
```diff
@@ -515,6 +515,10 @@ start_advertising(const void *cmd, uint16_t cmd_len,
 
 #if MYNEWT_VAL(BLE_EXT_ADV)
     adv_params.own_addr_type = own_addr_type;
+
+    if (sd_len != 0) {
```
sjanc, line 522: we may also have to unset it so I'd suggest

adv_params.scannable = sd_len != 0;

apps/bttester/src/btp_gap.c
```diff
@@ -1992,6 +2000,10 @@ set_filter_accept_list(const void *cmd, uint16_t cmd_len,
 
     SYS_LOG_DBG("");
 
+    if (cp->list_len != 0) {
+        use_filter_policy = true;
```
sjanc, line 2004: use_filter_policy = cp->list_len != 0;

apps/bttester/src/btp_gap.c
```diff
@@ -64,6 +64,9 @@ static ble_addr_t peer_id_addr;
 static ble_addr_t peer_ota_addr;
 static bool encrypted = false;
 
+static bool use_filter_policy = false;
+#define FILTER_POLICY_SCAN_CONN_ONLY_FROM_LIST 0x03
```
sjanc, line 68: this should be in hci_common.h

sjanc (changes requested):

apps/bttester/src/btp_gap.c
szymon-czapracki: This will need to be further discussed, as there are situations where sd_len is actually 0, but we don't want to set scannable to 0. TBD

apps/bttester/src/btp_gap.c
szymon-czapracki: Done

sjanc (approved):

PR #2036, opened 2025-05-16
nimble/phy/nrf5x: fix race conditions in TIMER0 setup

jwise:
In the nRF5x LL PHY, we use TIMER0 to trigger various Tx/Rx sequencing tasks.  The PHY attempted to recover from some cases in which it was 'late' setting up TIMER0; in theory, these ought not have happened, but if interrupt priorities do not provide absolute priority for the LL (for instance, in RTOS systems that have other priorities, or scheduling systems that have incompatible interrupt priority in order to call into RTOS queue management routines), then the PHY can indeed end up failing to schedule a transmit / receive window in time.

In some cases, the recovery process for these had race conditions, which could result in the LL state machine locking up (the end-of-tx interrupt would never fire).  One race condition was extremely short (a few instructions wide; if the timer was set, but the timer triggered before the appropriate PPI could be configured); the other race condition was a microsecond wide (comparison to see if the timer had already triggered needed the current time to be greater in order to fail, but if it triggered within that microsecond, the current time would be equal).  There are also other potential issues around timer event reset and check that we did not replicate, but did notice by inspection.

These cases appeared in a few different forms, so we factor out a "safe reset" routine to ensure that a timer will not trigger (to be used before setting up a PPI), and then we also factor out a routine to check whether a timer has been missed.

This change is verified to fix some observed stability issues in PebbleOS under heavy load.

cc: @gmarull @crc32 @sjp4 @ericmigi

### DIFF
--- a/nimble/drivers/nrf5x/src/ble_phy.c
+++ b/nimble/drivers/nrf5x/src/ble_phy.c
@@ -584,6 +584,31 @@ ble_phy_rxpdu_copy(uint8_t *dptr, struct os_mbuf *rxpdu)
 }
 
 /**
+ * There is no way on nRF5 to disable a COMPARE event entirely.  So we fake
+ * it, by setting it as far in the future as possible (by triggering a
+ * capture to it), and then clearing the event if it was set.
+ */
+static void
+timer0_safely_reset(int cc)
+{
+    nrf_timer_task_trigger(NRF_TIMER0, nrf_timer_capture_task_get(cc));
+    NRF_TIMER0->EVENTS_COMPARE[cc] = 0;
+}
+
+/**
+ * It is possible that we set up a timer, but we were too late -- it didn't
+ * fire.  Detect when this happens, reliably, by capturing the current time
+ * to a scratch counter and comparing with wraparound.
+ */
+static bool
+timer0_did_miss(int cc, int cc_scratch)
+{
+    nrf_timer_task_trigger(NRF_TIMER0, nrf_timer_capture_task_get(cc_scratch));
+    uint32_t ticks_rem = NRF_TIMER0->CC[cc] - NRF_TIMER0->CC[cc_scratch];
+    return (ticks_rem >= 0x80000000) && !NRF_TIMER0->EVENTS_COMPARE[cc];
+}
+
+/**
  * Called when we want to wait if the radio is in either the rx or tx
  * disable states. We want to wait until that state is over before doing
  * anything to the radio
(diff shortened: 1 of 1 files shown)


nimble/drivers/nrf5x/src/ble_phy.c
```diff
@@ -583,6 +583,31 @@ ble_phy_rxpdu_copy(uint8_t *dptr, struct os_mbuf *rxpdu)
            sizeof(struct ble_mbuf_hdr));
 }
 
+/**
+ * There is no way on nRF5 to disable a COMPARE event entirely.  So we fake
+ * it, by setting it as far in the future as possible (by triggering a
+ * capture to it), and then clearing the event if it was set.
+ */
+static void
+timer0_safely_reset(int timer)
```
andrzej-kaczmarek, line 592: int cc

nimble/drivers/nrf5x/src/ble_phy.c
```diff
@@ -583,6 +583,31 @@ ble_phy_rxpdu_copy(uint8_t *dptr, struct os_mbuf *rxpdu)
...
+    NRF_TIMER0->EVENTS_COMPARE[timer] = 0;
+}
+
+/**
+ * It is possible that we set up a timer, but we were too late -- it didn't
+ * fire.  Detect when this happens, reliably, by capturing the current time
+ * to a scratch counter and comparing with wraparound.
+ */
+static bool
+timer0_did_miss(int timer, int scratch)
```
andrzej-kaczmarek, line 604: int cc, int scratch_cc

nimble/drivers/nrf5x/src/ble_phy.c
```diff
@@ -583,6 +583,31 @@ ble_phy_rxpdu_copy(uint8_t *dptr, struct os_mbuf *rxpdu)
...
+{
+    nrf_timer_task_trigger(NRF_TIMER0, nrf_timer_capture_task_get(scratch));
+    uint32_t ticks_rem = NRF_TIMER0->CC[timer] - NRF_TIMER0->CC[scratch];
+    return (ticks_rem >= 0x80000000) && !NRF_TIMER0->EVENTS_COMPARE[timer];
```
andrzej-kaczmarek, line 608: you can simply compare `NRF_TIMER0->CC[timer] < NRF_TIMER0->CC[scratch]`

andrzej-kaczmarek (changes requested):
looks good, few style comments

jwise:
Thanks! I'll make these changes tomorrow -- it's a holiday in the US today.

nimble/drivers/nrf5x/src/ble_phy.c
jwise: Nope, this is a fix for the bug that I originally suspected I was hitting, but in the end I was not :-)

`NRF_TIMER0->CC` is a `uint32`.  If `scratch_cc` wrapped but `cc` did not -- i.e., `NRF_TIMER0->CC[cc]` is `0xFFFF_FFFE` but `NRF_TIMER0->CC[scratch_cc]` is `0x0000_0002`, then it's not true that `NRF_TIMER0->CC[cc] < NRF_TIMER0->CC[scratch_cc]`, but it *is* true that `(NRF_TIMER0->CC[cc] - NRF_TIMER0->CC[scratch_cc]) >= 0x80000000`.

nimble/drivers/nrf5x/src/ble_phy.c
jwise: Bumped in 63e39460.

sjanc:
I believe this is fixed in master now

PR #2037, opened 2025-05-16
nimble/bttester: Add option to enable/disable leaudio support

szymon-czapracki:
Add syscfg fields to allow user to enable/disable leaudio.
Some testcases were failing due to having leaudio services and performing writes which resulted in errors.


apps/bttester/syscfg.yml
```diff
@@ -22,6 +22,9 @@ syscfg.defs:
             Disable any NimBLE configuration in bttester app.
             This is useful for building bttester apps with custom configuration.
         value: 0
+    BTTESTER_LEAUDIO:
+        description: 'Enable NimBLE le-audio support in bttester app'
```
sjanc, line 26: le-audio  > LE Audio

sjanc (commented):
nitpick: we should follow BTSIG naming guidelines:)  otherwise looks good

apps/bttester/syscfg.yml
szymon-czapracki: Done

sjanc (approved):

PR #2040, opened 2025-05-22
BTP Periodic Sync Transfer  & Sync established event refactor

szymon-czapracki:
Get BTP event up to speed with autopts requirements.
Fill missing field in sync established event.


szymon-czapracki:
@sjanc FYI

sjanc (approved):

PR #2041, opened 2025-05-22
apps/bttester: Refactor reporting supported commands

szymon-czapracki:
Implement new way for reporting supported commands. Now only commands with registered handler will
be present in the queried response.


szymon-czapracki:
@sjanc FYI

sjanc (approved):

PR #2043, opened 2025-06-06
nimble/host: Enable address resolution if IRK is the same

guy-m:
Address resolution should be enabled in case of setting same IRK.
It should not be assumed it is still enabled from last IRK set as host can disable it with command or send HCI reset to controller (for example with host stop/start) which disabled it.
Otherwise a central peer with RPA will fail to connect.


sjanc:
Hi,

Please check if https://github.com/apache/mynewt-nimble/pull/2068 fix this issue for you. After HCi reset, resolving list is cleared anyway and we should have setup all from start, while that didn't happen was due to missing cleanup on host stop (fixed in 2068)

PR #2052, opened 2025-06-23
host: reset random address when starting

gmarull:
I have observed that, when stopping the host and scheduling a new start, advertising fails (with error 530). This is because the static variable `ble_hs_id_rnd` is never reset, but, at least in case of nRF52, the LL random address is reset in ble_ll_reset(), which gets called when host starts up via ble_hs_startup_reset_tx().

I'm not sure the solution is strictly correct for all cases/controllers. `ble_hs_id_pub` is never reset either, so maybe a more generic solution is needed.


gmarull:
Fixed by https://github.com/apache/mynewt-nimble/pull/2054

PR #2055, opened 2025-06-25
EATT manual connection & multiple channels initial support

szymon-czapracki:
New syscfg options for multiple EATT channels.
Introduce multiple channels to `ble_eatt` structure.
Add new API function for connecting EATT on demand.
Add new shell command for manually connecting EATT channels.


nimble/host/src/ble_eatt.c
```diff
@@ -472,17 +572,25 @@ int
...
     BLE_EATT_LOG_DEBUG("eatt: %s, size %d ", __func__, OS_MBUF_PKTLEN(txom));
     eatt = ble_eatt_find(conn_handle, cid);
-    if (!eatt || !eatt->chan) {
+    if (!eatt) {
+        BLE_EATT_LOG_ERROR("Eatt not available");
+        rc = BLE_HS_ENOENT;
+        goto error;
+    }
+
+    channel = ble_eatt_channel_find_by_cid(eatt, cid);
```
MariuszSkamra, line 586: This looks odd to me. If `ble_eatt_find()` takes as parameter `cid`, I would expect it will return `eatt` instance matching `cid` given. Now, when you place an array of channels into eatt so it does not simply wrap L2CAP channel, the `ble_eatt_find()` should not take `cid` as parameter IMO.

nimble/host/src/ble_eatt.c
```diff
@@ -231,7 +310,15 @@ ble_eatt_l2cap_event_fn(struct ble_l2cap_event *event, void *arg)
             ble_eatt_free(eatt);
             return 0;
         }
-        eatt->chan = event->connect.chan;
+        chan_idx = ble_eatt_free_channel_idx(eatt);
+        if (chan_idx > 0) {
+            eatt->chan[chan_idx] = event->connect.chan;
```
MariuszSkamra, line 315: Probably, the condition is incorrect, as array index 0 is valid

MariuszSkamra (changes requested):

PR #2056, opened 2025-06-25
Exclude btshell cmd.c from clang formatting

szymon-czapracki:
Formatting this file will result in extensive and non-necessary changes that make review and history tracking difficult.
To preserve clarity and minimize noise, this file is now excluded from clang-format.


sjanc (changes requested):
this is not a solution

PR #2059, opened 2025-06-26
Disable string breaking in clang-format

szymon-czapracki:
Previous configuration produced weirdly formatted
results while dealing with long strings.


sjanc (approved):

sjanc:
@szymon-czapracki we should have cron github action that would do this ie. periodically check if there were changes in core repo and create PR to sync this,  similar to how we sync ports syscfg

PR #2060, opened 2025-06-26
Keep sizeof parenthesis consistent in btp

szymon-czapracki:
Ensure all sizeof use parenthesis for consitency.

For reference:
https://lkml.org/lkml/2012/7/11/103


szymon-czapracki:
@andrzej-kaczmarek FYI

PR #2063, opened 2025-06-27
Skip style-check workflow if label is present

szymon-czapracki:
This update adds a label-based conditional to the compliance_check.yml GitHub Actions workflow. If a pull request includes the label `skip-style-check`, the style check step will be skipped.

This allows maintainers to bypass formatting validation in exceptional cases, such as bulk updates or style-irrelevant changes.

### DIFF
--- a/.github/workflows/compliance_check.yml
+++ b/.github/workflows/compliance_check.yml
@@ -19,7 +19,9 @@
 
 name: Compliance check
 
-on: [pull_request]
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


sjanc (commented):
to make it automatically restart when label is applied this should be triggered on labeled and unlabeled too

szymon-czapracki:
> to make it automatically restart when label is applied this should be triggered on labeled and unlabeled too

done

sjanc (approved):

PR #2080, opened 2025-07-09
nimble/btp: Implement missing BAP layer commands & events

szymon-czapracki:
Bring BAP commands and events up to date with their corresponding counterparts in Auto-PTS.


szymon-czapracki:
@sjanc Fixed compliance issues

PR #2082, opened 2025-07-16
nimble/host: Reject L2CAP_FLOW_CONTROL_CREDIT_IND with zero credits

guy-m:
L2CAP_FLOW_CONTROL_CREDIT_IND with 0 credits value is invalid and should be responded with L2CAP_COMMAND_REJECT_RSP.


sjanc:
Hmm L2CAP/LE/REJ/BI-02-C is for testing rejection of unknown PDU and opcode 0x16 would be sent only if CoC is not supported by IUT

Also it does pass in our internal tests (PTS 8.9.1)

guy-m:
You are correct. After further debug it appeared there was an ETS update for L2CAP that fixed that behavior.
After replacing the file PTS doesn't send the L2CAP_FLOW_CONTROL_CREDIT_IND at all and test pass without any fix.
This PR can be rejected.

PR #2090, opened 2025-07-28
host: att: convert BLE_HS_ATT_SVR_QUEUED_WRITE_TMO to system ticks

guy-m:
The value of BLE_HS_ATT_SVR_QUEUED_WRITE_TMO is the expiry time for incoming ATT queued writes in ms. When used it needs to be converted to system ticks.


sjanc:
Since tests also had to be udpated, I've cherry-picked this into https://github.com/apache/mynewt-nimble/pull/2092

PR #2118, opened 2025-09-15
fix(nimble): Address buffer copy for mbuf in different pools.

rahult-github:
Given a chain of at least 2 mbufs, of which the mbufs come from more than one pool of different-sized buffers and the first mbuf is a smaller size than at least one of the rest, the memcpy() in os_mbuf_dup() will write beyond the limits of the allocated mbuf.

This is because os_mbuf_dup() assumes all mbufs in a chain come from the same pool as the first mbuf in the chain. Fix the same


rahult-github:
@sjanc  , Please take a look

PR #2123, opened 2025-10-07
nimble/host: Fix ATT Read By Type Response handling

andrzej-kaczmarek:
ATT_READ_BY_TYPE_RSP handler will extract proc from the proc list and then call the callback for each handle-value pair in the PDU. However, this means each time proc is extracted it will be then added to the tail of the proc list. If there are more than 1 matching procesures on the list, subsequent extracts when handling *the same PDU* may extract invalid proc from the list and break the procedure.

This fixes the problem by adding proc back to list at the head instead of the tail of proc list. This way subsequent iterations of handle-value pairs handling will extract proper proc as it will be the first matching on the list.


nimble/host/src/ble_gattc.c
```diff
@@ -4892,8 +4896,15 @@ ble_gattc_rx_read_type_adata(uint16_t conn_handle, uint16_t cid,
                                          &rx_entry);
     if (proc != NULL) {
         rc = rx_entry->cb(proc, adata);
-        ble_gattc_process_status(proc, rc);
+        /* We need to put procedure on head of the queue since caller expects
+         * to process the same proc again.
+         */
+        ble_gattc_process_status(proc, rc, true);
```
sjanc, line 4902: shoudn't this be put back only on success?

PR #2125, opened 2025-10-18
fix(host): add parameter validation to Channel Sounding APIs

akshatsinha0:
NimBLE v1.8.0:xx
Tried to implements the missing functionality in the Channel Sounding (BLE 5.4) host APIs.


akshatsinha0:
@MariuszSkamra

PR #2130, opened 2025-10-31
nimble/host: Fix format string in ble_gap_log_duration

ehaas:
The format string was using %d for an int32_t (long int)


sjanc (approved):
I'm okay with this fix, however in my very personal opinion casting duration_ms to int would be just fine too

sjanc (approved):

PR #2134, opened 2025-11-13
asf.yaml: Force linear history

sjanc:
This disables merge commits and allows only rebase or squash commits. This is to keep repo history linear.


vrahane (approved):
LGTM

PR #2207, opened 2026-04-17
host: Fix recursive lock in ble_hs_stop_done

rojer:
ble_hs_stop_done() held ble_hs_lock while calling ble_gap_reset_state() which internally acquires the same lock via ble_hs_atomic_first_conn_handle() and ble_gap_conn_broken(). On platforms with non-recursive mutexes this causes a deadlock or assertion failure.

Move ble_hs_stop_hci_reset(), ble_gap_reset_state(), ble_hs_id_reset(), and ble_hs_pvcy_reset() outside the locked section. These functions either manage their own locking internally or don't need the lock. ble_hs_enabled_state is set to OFF under the lock before releasing, preventing new operations.

### DIFF
--- a/nimble/host/src/ble_hs_stop.c
+++ b/nimble/host/src/ble_hs_stop.c
@@ -69,17 +69,20 @@ ble_hs_stop_done(int status)
 
     ble_hs_enabled_state = BLE_HS_ENABLED_STATE_OFF;
 
+    ble_hs_unlock();
+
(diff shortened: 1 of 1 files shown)


sjanc:
BTW I do think that in general we should get rid of this assert and simply make ble_hs_lock() recursive  (TBH I have no idea why it is like that in first place...)

PR #2219, opened 2026-05-04
nimble/ll: Fix BIGInfo report parsing

MariuszSkamra:
This fixes the missing BIGInfo size check as well as the PHY value passed to the HCI BIGInfo report, which was not corresponding to the one in the LL event.
Additionally, this fix addresses the framing field.

In accordance with the Core Specification,
unsupported PHY reports are now discarded.

### DIFF
--- a/nimble/controller/src/ble_ll_sync.c
+++ b/nimble/controller/src/ble_ll_sync.c
@@ -673,11 +673,36 @@ ble_ll_sync_send_truncated_per_adv_rpt(struct ble_ll_sync_sm *sm, uint8_t *evbuf
 }
 
 #if MYNEWT_VAL(BLE_LL_PERIODIC_ADV_SYNC_BIGINFO_REPORTS)
+static uint8_t
+ble_ll_sync_biginfo_phy_get(const uint8_t *biginfo)
+{
+    uint8_t phy;
+
+    phy = (biginfo[27] >> 5) & 0x07;
+
+    switch (phy) {
+    case 0:
+        return BLE_PHY_1M;
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_LE_2M_PHY)
+    case 1:
+        return BLE_PHY_2M;
+#endif
+#if MYNEWT_VAL(BLE_LL_CFG_FEAT_LE_CODED_PHY)
+    case 2:
+        return BLE_PHY_CODED;
+#endif
+    default:
+        return 0x00;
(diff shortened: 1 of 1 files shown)


MariuszSkamra:
Rebase only

sjanc (commented):
I think HCI ev should not be send if received data is not valid (ie we should not send HCI event with RFU values).  example would be PHY=0 or Max_SDU=0

nimble/controller/src/ble_ll_sync.c
```diff
@@ -712,25 +743,42 @@ ble_ll_sync_parse_biginfo_to_ev(struct ble_hci_ev_le_subev_biginfo_adv_report *e
...
+    /* The BIGInfo length is 33 octets for an unencrypted BIG and 57 octets for an encrypted BIG. */
+    if (biginfo_len != BLE_LL_SYNC_BIGINFO_LEN &&
+        biginfo_len != BLE_LL_SYNC_BIGINFO_LEN_ENC) {
+        return;
+    }
+
+    memset(&ev, 0, sizeof(ev));
+    ble_ll_sync_parse_biginfo_to_ev(&ev, biginfo, biginfo_len);
+
+    if (!ev.phy) {
```
MariuszSkamra: @sjanc Regarding your comment, the PHY is checked here, if invalid, the event won't be sent. But I agree, there might be missing checks like the Max_SDU mentioned.

sjanc: right,  I'd suggest to change this check to something like `validate_hci_big_ev()` or similar that would validate all fields that are derived from PDU

MariuszSkamra: The PHY requirement to ignore the unsupported value is explicitly defined in spec.
In case of other event structure members there is no such requirement. The Core defines the All other values are RFU, thus maybe it's worth to pass them to the host anyway?

MariuszSkamra:
rebase only

sjanc (approved):

nimble/controller/src/ble_ll_sync.c
sjanc, line 775: return true is missing,  rebase artefact?

sjanc (changes requested):

nimble/controller/src/ble_ll_sync.c
MariuszSkamra: updated

PR #2220, opened 2026-05-04
nimble: Fix the HCI ISO packet lost status value

MariuszSkamra:
As per the Core, the Packet_Status_Flag of an HCI ISO Data packet must be one of the following: 0b00, 0b01, or 0b10.
This fixes the BLE_HCI_ISO_PKT_STATUS_LOST value which shall be 0x02 (0b10).


MariuszSkamra:
Rebase only

sjanc (approved):

PR #2231, opened 2026-05-12
nimble/host: Refactor GATT Read Multiple Variable Char Value response

sjanc:
This reworks ATT_READ_MULTIPLE_VARIABLE_RSP parsing. If response doesn't match request (number of values requested) or LV parsing shows incorrect format we ignore that and report error to app.

### DIFF
--- a/nimble/host/src/ble_gattc.c
+++ b/nimble/host/src/ble_gattc.c
@@ -3361,49 +3361,62 @@ ble_gattc_read_mult_cb_var(struct ble_gattc_proc *proc, int status,
     for (i = 0; i < proc->read_mult.num_handles; i++) {
         attr[i].handle = proc->read_mult.handles[i];
         attr[i].offset = 0;
-        if (om == NULL || OS_MBUF_PKTLEN(*om) == 0) {
-            continue;
-        }
-
-        *om = os_mbuf_pullup(*om, 2);
-        assert(*om);
-
-        attr_len = get_le16((*om)->om_data);
-
-        os_mbuf_adj(*om, 2);
-
-        if (attr_len > BLE_ATT_ATTR_MAX_LEN) {
-            /*TODO Figure out what to do here */
-            break;
-        }
-
-        attr[i].om = os_msys_get_pkthdr(attr_len, 0);
-        if (!attr[i].om) {
-            /*TODO Figure out what to do here */
-            break;
-        }
-
-        rc = os_mbuf_appendfrom(attr[i].om, *om, 0, attr_len);
-        if (rc) {
-            /*TODO Figure out what to do here */
-            break;
-        }
-
-        os_mbuf_adj(*om, attr_len);
     }
 
-    /*FIXME Testing assert */
-    assert(i == proc->read_mult.num_handles);
+    if (status == 0) {
+        for (i = 0; i < proc->read_mult.num_handles; i++) {
+            if (OS_MBUF_PKTLEN(*om) < 2) {
+                break;
+            }
+
+            *om = os_mbuf_pullup(*om, 2);
+            if (*om == NULL) {
+                break;
+            }
+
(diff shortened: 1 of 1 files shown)


nimble/host/src/ble_gattc.c
```diff
@@ -3361,49 +3361,59 @@ ble_gattc_read_mult_cb_var(struct ble_gattc_proc *proc, int status,
...
-                    proc->read_mult.cb_arg);
+                            ble_gattc_error(status, att_handle), &attr[0],
+                            proc->read_mult.num_handles,
+                            proc->read_mult.cb_arg);
 
     for (i = 0; i < proc->read_mult.num_handles; i++) {
-        if (attr[i].om != NULL) {
-            os_mbuf_free_chain(attr[i].om);
-        }
+        os_mbuf_free_chain(attr[i].om);
```
MariuszSkamra, line 3419: If the mbuf allocation fails in 3384, the attr[i].om will be NULL, or the buffer is dropped in 3402. I think the check should not be removed.

nimble/host/src/ble_gattc.c
```diff
@@ -3361,49 +3361,59 @@ ble_gattc_read_mult_cb_var(struct ble_gattc_proc *proc, int status,
...
-    assert(i == proc->read_mult.num_handles);
+            status = BLE_HS_EBADDATA;
+        }
+    }
 
     proc->read_mult.cb_mult(proc->conn_handle,
-                    ble_gattc_error(status, att_handle), &attr[0],
-                    i,
-                    proc->read_mult.cb_arg);
+                            ble_gattc_error(status, att_handle), &attr[0],
```
MariuszSkamra, line 3415: The status set in 3406 (BLE_HS_EBADDATA) is passed to ble_gattc_error() function which is bad IMO.

nimble/host/src/ble_gattc.c
```diff
@@ -3361,49 +3361,59 @@ ble_gattc_read_mult_cb_var(struct ble_gattc_proc *proc, int status,
...
 
-        attr[i].om = os_msys_get_pkthdr(attr_len, 0);
-        if (!attr[i].om) {
-            /*TODO Figure out what to do here */
-            break;
-        }
+            attr[i].om = os_msys_get_pkthdr(attr_len, 0);
+            if (!attr[i].om) {
+                break;
+            }
```
MariuszSkamra, line 3389: This will result in BLE_HS_EBADDATA error, which can be misleading I think. Consider BLE_HS_ENOMEM.

MariuszSkamra (changes requested):

nimble/host/src/ble_gattc.c
MariuszSkamra: Or maybe not. The ble_gattc_error() function name seems to be misleading.

nimble/host/src/ble_gattc.c
sjanc: os_mbuf_free_chain() handles NULL (just like free() does) on its own

MariuszSkamra (approved):

nimble/host/src/ble_gattc.c
```diff
@@ -3361,49 +3361,62 @@ ble_gattc_read_mult_cb_var(struct ble_gattc_proc *proc, int status,
...
+
+            os_mbuf_adj(*om, attr_len);
         }
 
-        os_mbuf_adj(*om, attr_len);
-    }
+        /* failed to correctly parse response,
+         * cleanup any partial data and set status if not set already
+         */
+        if (i != proc->read_mult.num_handles || OS_MBUF_PKTLEN(*om) != 0) {
```
rymanluk: nit:   i <  proc->read_mult.num_handles

rymanluk (approved):

PR #2237, opened 2026-05-20
btp: Coding style fix for BTP handles

szymon-czapracki:
Clang-format was flagging these structures as invalid. This led to many failing CI's during PR process.

As there is no specific option in cf to enforce existing style, change this to format that cf will accept.


szymon-czapracki:
@piotrnarajowski @sjanc FYI

piotrnarajowski (approved):

PR #2243, opened 2026-05-27
fix portability: replace min() macro with MIN()

manuelinbrain:
Some standalone builds (non-Mynewt environments) do not provide min() macro by default, which can lead to compilation failures in controller sources depending on the toolchain.

This change replace min() macro with MIN() to improve portability.


sjanc (changes requested):
Hi,

ble_ll_utils.h already provides MIN() macro which should be used instead

manuelinbrain:
Hi,

It's already done. Thanks

sjanc (approved):

PR #2247, opened 2026-06-09
nimble/transport/common/hci_h4: add HCI H4 packet callback

jwise:
Certain controllers have extensions to the HCI H4 protocol (i.e., TI CC256x has eHCILL, which are additional bytes that show up in the stream to indicate power mode transitions).  The transport needs to be able to handle these; provide a callback that a custom hcill_uart_transport could request to be able to consume bytes at packet boundaries in the HCI H4 state machine.


jwise:
Hi,

Any chance this could be merged?

sjanc:
could you try rebasing so that we can re-run with updated CI?

PR #2250, opened 2026-06-26
nimble/host: Fix rejected SM SC debug keys

mkasenberg:


nimble/host/src/ble_sm_sc.c
```diff
@@ -612,13 +612,15 @@ ble_sm_sc_public_key_rx(uint16_t conn_handle, struct os_mbuf **om,
     }
 
     cmd = (struct ble_sm_public_key *)(*om)->om_data;
+#if !MYNEWT_VAL(BLE_SM_SC_DEBUG_KEYS)
```
sjanc, line 624: even with debug keys enabled this should allow to have same public keys only if keys is debug

sjanc (changes requested):

PR #2251, opened 2026-07-01
nimble/audio: Bound broadcast sink subgroup count

Old-Ding:
## Summary

- reject broadcast sink sync options whose subgroup count exceeds capacity
- add a broadcast sink unit test for oversized subgroup counts
- enable broadcast sink coverage in the audio unit test syscfg
- enable the periodic advertising sync transfer and BIGINFO report syscfg options needed by the broadcast sink GAP event handler

## Root cause

`ble_audio_broadcast_sink_config()` copied `sync_opt->num_subgroups`
into `sink->num_subgroups` and then used it as the loop bound for the
fixed `sink->subgroups[]` array. The visible guard was only a debug
assertion in the scan delegator path, so production builds could still
receive an oversized sync option.

Enabling `BLE_AUDIO_BROADCAST_SINK` in the audio unit test also compiles
`ble_audio_broadcast_sink.c`, whose GAP handler references periodic sync
transfer and BIGINFO report event fields. The test syscfg now enables the
feature flags that expose those fields and declarations.

## Validation

- `git diff --check`
- `newt test @apache-mynewt-nimble/nimble/host/audio/test -j 4` parses the package and updated syscfg on this Windows host, then stops at the native compiler step because `gcc` is unavailable locally.
- PR-target checks are passing. Heavy `pull_request` workflows for head `e93fbb98e6eec1e69680067a7dcba30f28fb9e31` are `action_required` with no jobs, so the PR is waiting for maintainer workflow approval before those CI jobs can rerun.

Fixes #2188.


sjanc:
Hi,

I have mixed feelings on this PR, while fix itself seems to be OK from quick look, unittest is not building, and if it would build it would fail (due to https://github.com/apache/mynewt-nimble/pull/2254).

So, was this tested at all?

PR #2270, opened 2026-07-16
nimble/host: add Client Supported Features bit masks

piotrnarajowski:
Define possible bit masks for GATT Client Supported Features characteristic value to get rid of magic numbers.

### DIFF
--- a/nimble/host/src/ble_gattc.c
+++ b/nimble/host/src/ble_gattc.c
@@ -60,6 +60,7 @@
 #include "host/ble_uuid.h"
 #include "host/ble_gap.h"
 #include "ble_hs_priv.h"
+#include "../services/gatt/include/services/gatt/ble_svc_gatt.h"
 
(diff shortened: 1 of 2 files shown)


piotrnarajowski:
@KKopyscinski FYI

nimble/host/src/ble_gattc.c
```diff
@@ -60,6 +60,7 @@
 #include "host/ble_uuid.h"
 #include "host/ble_gap.h"
 #include "ble_hs_priv.h"
+#include "../services/gatt/include/services/gatt/ble_svc_gatt.h"
```
sjanc, line 63: I don't like it,  it add hidden service dependency to gatt code

I'm not yet sure on how to handle this though, maybe make those in public host header, or maybe add explicit dependency on that service (and GAP?)

need to think about this a bit more

PR #2281, opened 2026-08-07
apps/bttester: Remove a leftover/broken implementation

mkasenberg:
It seems that the ccc_handle was intended to store the handle of the most recently updated CCC descriptor, however the field was overwritten with the descriptor value instead, so the unsubscription was failing. Apart from that in the RAP profile test cases there are multiple subsequent subscriptions and unsubscriptions, so the storing and the checks would be an obstacle.


sjanc:
looks ok but please fix compliance check

sjanc (approved):

PR #2288, opened 2026-08-24
nimble/host: Fix handle validation in Read Mult Response

piotrnarajowski:
If one of the requested characteristic values filled the MTU, this caused the while loop to exit early. Remaining handle validation was skipped and as a result a response was issued even when the remaining handle was invalid.

Validate all requested handles first. Move MTU check outside while loop and truncate the response if needed after all handles were validated.

Fixes qualification test case GATT/SR/GAR/BI-19-C


piotrnarajowski:
Update: freeing txom when exception occured was missing

passes 20/20 times including iterations where PTS sends Read Mult Req with (LONG VAL HDL, INVALID HDL)

image

sjanc (approved):

